PR #16, opened 2026-03-05
feat: diagnostic logging for index_folder — skip counts and no-symbol tracking

gebeer:
## Problem

When `index_folder` produces fewer indexed files than expected, there's no way to understand why. Files can be silently dropped at two distinct stages:

1. **Discovery** — filtered by skip patterns, `.gitignore`, size limits, secret detection, etc.
2. **Parsing** — files that pass discovery but contain no extractable AST symbols (e.g. procedural config files, template includes with no function/class definitions)

### Diagnostic response fields on `index_folder`

Every `index_folder` call now returns:

- **`discovery_skip_counts`** — per-reason counts of files excluded during discovery (`skip_pattern`, `gitignore`, `symlink`, `wrong_extension`, `too_large`, `secret`, `binary`, `unreadable`, `path_traversal`, `symlink_escape`)
- **`no_symbols_count`** — number of files that passed discovery but yielded zero AST symbols
- **`no_symbols_files`** — up to 50 example paths, for inspection

### Server-level debug logging

`main()` now accepts `--log-level` and `--log-file` flags (also via `JCODEMUNCH_LOG_LEVEL` / `JCODEMUNCH_LOG_FILE` env vars). At `DEBUG` level every file decision is logged — `ACCEPT`, `SKIP <reason>`, `NO SYMBOLS`, `PARSE ERROR`. Defaults to `WARNING` so normal operation is silent.

### Documentation

- README: new env var table entries + debug config example
- MCP tool description: updated to document the diagnostic response fields (kept lean)

### Test fix

`discover_local_files()` now returns a 3-tuple `(files, warnings, skip_counts)`. Updated all 5 call sites in `tests/test_security.py`.

## Motivation

Validated against a ProcessWire CMS project with 3,225 PHP files. The diagnostic output immediately explained an apparent discrepancy (11,165 `skip_pattern` hits from `vendor/`+`.git/`, 5,791 gitignore hits, 206 files with no symbols due to procedural PHP patterns) — confirming the indexer was behaving correctly.

### DIFF
--- a/src/jcodemunch_mcp/tools/index_folder.py
+++ b/src/jcodemunch_mcp/tools/index_folder.py
@@ -356,6 +356,9 @@ def index_folder(
                 "changed": len(changed), "new": len(new), "deleted": len(deleted),
                 "symbol_count": len(updated.symbols) if updated else 0,
                 "indexed_at": updated.indexed_at if updated else "",
+                "discovery_skip_counts": skip_counts,
+                "no_symbols_count": len(incremental_no_symbols),
+                "no_symbols_files": incremental_no_symbols[:50],
             }
             if warnings:
                 result["warnings"] = warnings
--- a/tests/test_security.py
+++ b/tests/test_security.py
@@ -260,7 +260,7 @@ class TestDiscoverLocalFilesSecure:
         (tmp_path / "kept.py").write_text("x = 1\n")
         (tmp_path / "ignored.py").write_text("y = 2\n")
 
-        files, _, __ = discover_local_files(tmp_path)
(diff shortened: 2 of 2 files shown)


gebeer:
Thanks for the thorough review!

Both issues addressed in 67bbb42:

- **Bug fix**: `discovery_skip_counts`, `no_symbols_count`, and `no_symbols_files` are now included in the incremental return path — they were computed but never surfaced in the result dict.
- **Nit**: replaced `_, __` unpacking with `*_` in the three `discover_local_files` call sites in `test_security.py`.

All tests pass. Ready for re-review.

PR #17, opened 2026-03-05
feat: support ANTHROPIC_BASE_URL for third-party endpoints

Cobdog:
Allows pointing the Anthropic summarizer at third-party endpoints
(e.g. z.ai) that implement the Anthropic API.

When `ANTHROPIC_BASE_URL` is set alongside `ANTHROPIC_API_KEY`, the
value is passed as `base_url` to the `Anthropic()` client. When unset,
behavior is unchanged.

### Changes

- `batch_summarize.py` — read env var, build kwargs conditionally
- `README.md` — added to both env var tables
- `tests/test_summarizer.py` — two tests (with and without base URL)
  using `sys.modules` mock injection to handle the optional `anthropic`
  import

### Test plan

- [x] 188 tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,12 +194,12 @@ After saving the config, **restart Claude Desktop / Claude Code** for the server
 
 Environment variables are optional:
 
-| Variable            | Purpose                                      |
-| ------------------- | -------------------------------------------- |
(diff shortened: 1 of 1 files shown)


jgravelle (approved):
Clean, minimal change. Good to go.

PR #28, opened 2026-03-05
Make the indexing file cap configurable

allenguarnes:
## What changed

jCodeMunch has been capped at 500 files during indexing. That default is still sensible, but it becomes a real limitation on larger repos.

This PR keeps `500` as the default and adds an escape hatch:

- new env var: `JCODEMUNCH_MAX_INDEX_FILES`
- applies to both `index_folder` and `index_repo`
- invalid or non-positive values fall back to the default
- explicit `max_files` arguments still win if they are passed directly

I also cleaned up the truncation messaging so we only say "indexed first N" when files were actually dropped. If a repo happens to have exactly `N` files, we no longer report that as truncation.

## Why

The current hard-coded cap is safe, but too rigid for real-world codebases. This change makes the limit adjustable without changing the default behavior for existing users.

In other words: nothing changes unless someone opts in.

## Notes

- default remains `500`
- docs now mention `JCODEMUNCH_MAX_INDEX_FILES`
- tests cover:
  - default behavior
  - env override
  - invalid env fallback
  - explicit argument precedence
  - exact-limit vs actual truncation

## Testing

```bash
uv run --extra test pytest
```
Full test suite passed locally: `222 passed`

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,6 +374,7 @@ For **LM Studio**, ensure the Local Server is running (usually on port 1234):
 | `OPENAI_MODEL`              | Model name for local LLMs (default: `qwen3-coder`) | No |
 | `OPENAI_TIMEOUT`            | Timeout in seconds for local requests (default: `60.0`) | No |
 | `CODE_INDEX_PATH`           | Custom cache path         | No       |
+| `JCODEMUNCH_MAX_INDEX_FILES`| Maximum files to index per repo/folder (default: `500`) | No |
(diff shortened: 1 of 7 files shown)


jgravelle (approved):
Clean implementation. Single source of truth in security.py, correct precedence (explicit > env > default 500), the truncated bool properly replaces the old >= 500 false positive, and tests cover all the cases including exact-limit-is-not-truncation. Merging.

PR #61, opened 2026-03-07
Add token-stats CLI (--token-stats, --token-stats-all, --output-format)

snafu4:
# Motivation
- Expose a small CLI for inspecting accumulated token-savings stored in the local index to help users understand savings and costs.
- Provide a stable programmatic API to produce enriched savings reports for use by the CLI and other tooling.
- Document usage and examples in README.md and ensure terminal-friendly output with optional Rich styling.

# Description
- Added --token-stats, --token-stats-all, and --output-format CLI flags to main in src/jcodemunch_mcp/server.py with a custom help formatter and field explainer via _HelpFormatter and _token_stats_fields_explainer().
- Implemented get_savings_report in src/jcodemunch_mcp/storage/token_tracker.py to return an enriched JSON report containing totals, cost estimates, equivalent context windows, telemetry flags, and the savings file path.
- Added human-readable formatting and printing helpers (_format_token_stats_text, _token_stats_summary, and _print_token_stats) and optional rich rendering when available, and exported get_savings_report from src/jcodemunch_mcp/storage/__init__.py.
- Updated pyproject.toml to add rich as a dependency and extended README.md with CLI usage examples and sample JSON/text output for the token-stats commands.
- Added CLI tests in tests/test_cli.py to assert JSON and text output for the new flags and that help output contains the explainer text.
# Testing
- Ran the test suite with pytest and the updated CLI tests in tests/test_cli.py, and all tests passed.
- Verified the new JSON output shape by running the new tests that call main(["--token-stats", "--output-format", "json"]) and main(["--token-stats-all", "--output-format", "json"]), which succeeded.
- Verified the human-readable text output via the new test_main_token_stats_text_outputs_human_readable, which succeeded.
- Confirmed --help prints defaults and the token-stats fields explainer via the updated help test, which succeeded.

**Note** Repository maintainer should confirm that the stats reported match with what is being sent to server.


jgravelle (commented):
## Review

### Good stuff
- The `_savings_path` fix to honor `CODE_INDEX_PATH` env var is the right fix — it's the bug that was causing CLI and server-side recording to diverge. Your note about confirming stats alignment is addressed by this fix and validated by `test_savings_report_uses_code_index_path_env`. Stats are aligned.
- `_read_savings_data` / `_write_savings_data` helpers are solid refactors — encoding fallbacks (`utf-8-sig`, `utf-8`, `cp1252`) are a real-world win on Windows.
- `_HelpFormatter` combo is clean.
- Tests are well-structured and cover the real behavior.

---

### Issues

**[Blocker] `rich` should be optional, not a hard dependency**
`pyproject.toml` adds `"rich>=13.7.0"` to `dependencies`, but `_print_token_stats` already has the correct fallback logic for when rich isn't present. Making it a hard dep forces a 2MB+ package on every install just for optional CLI formatting. Please move it to `optional-dependencies` (e.g., an `[extras]` group) and rely on the existing fallback.

**[Minor] README duplication**
The `--token-stats` (no flags) example appears twice — once at the top and again under "Prefer a non-JSON, terminal-friendly summary?". One of them should be cut.

**[Minor] GPT-5 pricing is speculative**
Hardcoding `gpt5_latest: 0.00001` ($10/1M tokens) in `PRICING` ties the output to a competitor's unreleased pricing. Consider removing it or replacing with a second Claude model (e.g., `claude_haiku`) or another reference point with known, stable pricing.

**[Minor] Fragile import in `test_token_tracker.py`**
Using `importlib.util.spec_from_file_location` to import `token_tracker` by file path is brittle and bypasses the package system. Replace with a normal import:
```python
from jcodemunch_mcp.storage.token_tracker import record_savings, get_savings_report, get_total_saved
```

---

### Summary

The core logic is sound and the path-alignment fix is correct. One blocker: `rich` must not be a hard dependency — the fallback logic is already written correctly for optional rich, just wire it up properly. The other items are minor polish. Good contribution — happy to merge after the blocker is addressed.

PR #71, opened 2026-03-08
feat: concurrent batch summarization with local LLM tuning docs

zrk02:
### Motivation

Tested various models and settings to make use of a local LLM for AI summaries.
Ran into  problems of low quality outputs, or low gpu usage,  and or long runtimes.

Providing here some new tunables, as well as tips for running a local LLM via LM studio, with a few crucial settings that made huge differences across approx 20 controlled test runs. 
                                   
 ###  What does this change do?
  - Adds concurrent batch processing, configurable with OPENAI_CONCURRENCY (default:  1) 
  - Adds OPENAI_BATCH_SIZE env var for runtime tuning.
  - Adds OPENAI_MAX_TOKENS env var for runtime tuning.

### Impact on default values
- None, these adds tunables that are not required to be set. 
                                 
### Documentation
  - Add "Local LLM Tuning Tips" section to USER_GUIDE.md covering Qwen3-8B setup, LM Studio /no_think config, and performance tuning  env vars
  - Documented new env vars in README.md


jgravelle (approved):
Nice contribution — this is exactly the kind of real-world tuning feedback the project needs.

**Code review**

The concurrent batch dispatch is sound:
- Each  future gets a separate slice of  objects, so there's no shared mutable state between concurrent batches.
-  already catches its own exceptions and falls back to signatures, so  won't surface API errors to callers.
- Default  means  — functionally identical to the old sequential loop, zero behavior change for existing users.

**Minor note:** At , you do pay thread-pool overhead on every  call. For the common case (local LLM, sequential) this is negligible since batches are network IO-bound, but could be micro-optimized by short-circuiting to the old loop when `max_workers == 1`. Up to you — not a blocker.

**Docs** are excellent. The `/no_think` tip for Qwen3 with LM Studio is the kind of hard-won operational knowledge that saves other users hours.

Merging — thanks for the thorough testing across 20 runs!

jgravelle (approved):
Nice contribution. The concurrent batch dispatch is sound: each future gets a separate Symbol slice (no shared mutable state), and _summarize_one_batch already catches its own exceptions so future.result() won't surface API errors. Default OPENAI_CONCURRENCY=1 means functionally identical behavior to the old sequential loop. Minor note: at max_workers=1 you pay thread-pool overhead on every call; could short-circuit to the old loop in that case, but not a blocker since batches are network IO-bound. The /no_think tip for Qwen3 + LM Studio is the kind of hard-won knowledge that saves other users hours. Merging.

Issue #77, opened 2026-03-09
Claude Opus cost


MariusAdrian88:
Hi J,

Why is the cost of the Claude Opus 4.6 so inflated, at $15 for 1M tokens for input context. 

The actual costs now: Claude Opus 4.6  --> $5 / MTok which is 1/3 from the estimation that we get.
Source: https://platform.claude.com/docs/en/about-claude/pricing

Perhaps we could auto-detect the used model and get the cost from a configurable table in a config file.

Best regards,
Marius

PR #79, opened 2026-03-09
feat: SQL language support with dbt Jinja preprocessing

paperlinguist:
## Summary
- Add SQL as a supported language (`.sql` files) using the `tree-sitter-sql` (derekstride) grammar
- Extract symbols from `CREATE TABLE`, `CREATE VIEW`, `CREATE FUNCTION`, `CREATE INDEX`, `CREATE SCHEMA`, and CTE
definitions
- Preprocess dbt Jinja-templated SQL (`{{ }}`, `{% %}`, `{# #}`) before parsing so tree-sitter can handle dbt model files
- Extract dbt directives (`{% macro %}`, `{% test %}`, `{% snapshot %}`, `{% materialization %}`) as first-class symbols
with signatures, params, and docstrings
- Fix `test_respects_env_file_limit` to set `JCODEMUNCH_MAX_FOLDER_FILES` alongside the legacy env var so the test passes
when the folder-specific var is set in the environment
## Real-world validation
Tested against a production dbt project with 4,200+ SQL files:
| Metric | Result |
|--------|--------|
| SQL files indexed | 4,202 |
| Symbols extracted | 7,272 |
| Index time (no AI summaries) | 56.7 seconds |
| Files with no extractable symbols | 2,546 (simple SELECT models — expected) |
## Files changed
**New files:**
- `src/jcodemunch_mcp/parser/sql_preprocessor.py` — Jinja stripping and dbt directive extraction
- `tests/test_sql_language.py` — 27 tests covering DDL, CTEs, Jinja preprocessing, and all dbt directive types

**Modified files:**
- `src/jcodemunch_mcp/parser/extractor.py` — `_parse_sql_symbols()` function
- `src/jcodemunch_mcp/parser/languages.py` — `SQL_SPEC`, `.sql` extension mapping, registry entry
- `CHANGELOG.md` — `[Unreleased]` section documenting SQL support
- `LANGUAGE_SUPPORT.md` — SQL row in language table + `sql` in valid language names
- `README.md` — SQL row in Supported Languages table
- `tests/test_security.py` — fix `test_respects_env_file_limit` to also set `MAX_FOLDER_FILES_ENV_VAR`

## SQL symbol extraction details
| SQL construct | Symbol kind | Signature example |
|---------------|-------------|-------------------|
| `CREATE TABLE` | type | `CREATE TABLE orders (id, customer_id, status)` |
| `CREATE VIEW` | type | `CREATE VIEW active_orders` |
| `CREATE FUNCTION` | function | `CREATE FUNCTION get_total(p_id INT) RETURNS DECIMAL` |
| `CREATE INDEX` | type | `CREATE INDEX idx_status ON orders` |
| `CREATE SCHEMA` | type | `CREATE SCHEMA analytics` |
| CTE (`WITH x AS`) | function | `WITH ranked AS (...)` |
| `{% macro %}` | function | `{% macro dedupe(source_cte, partition_by) %}` |
| `{% test %}` | function | `{% test row_count(model, source_name) %}` |
| `{% snapshot %}` | type | `{% snapshot snap_contacts %}` |
| `{% materialization %}` | function | `{% materialization cortex_agent %}` |

## Test plan
- [x] All 27 new SQL tests pass (`tests/test_sql_language.py`)
- [x] Full suite: 431 passed, 4 skipped, 0 failed
- [x] Validated on a real 4,200-file dbt project
- [x] Validated on Python 3.10, 3.11, 3.12


paperlinguist:
> This is a genuinely impressive contribution — real-world validation on 4,200 dbt files with 431 tests passing is exactly the right way to land a feature this size. The architecture is clean and the implementation details are solid throughout.
> 
> A few things before this can merge:
> 
> ### Rebase required
> Per project policy, large PRs need to be rebased onto current `main` before merging. Can you pull the latest `jgravelle/jcodemunch-mcp:main` into your fork and resolve any conflicts? The issue #77 pricing fix just landed and there are other recent commits this should incorporate.
> 
> ### Context provider performance: no opt-out path
> `discover_providers()` runs on every `index_folder` call and may parse thousands of YAML files across a large project tree. On the 4,200-file dbt repo you tested, that's potentially a lot of schema.yml I/O on every re-index. There's currently no way to disable this.
> 
> Suggestion: a `JCODEMUNCH_CONTEXT_PROVIDERS=0` env var (or `--no-context-providers` parameter to `index_folder`) would give users an escape hatch for cases where they don't want the overhead or are hitting edge cases. The "zero config" default is great; just needs an override.
> 
> ### `detect()` sets instance state as a side effect
> ```python
> def detect(self, folder_path: Path) -> bool:
>     self._dbt_yml_path = _detect_dbt_project(folder_path)  # ← side effect
>     return self._dbt_yml_path is not None
> ```
> 
> If `load()` is ever called without `detect()` having been called first (e.g. in a test or a future code path), it raises `AttributeError`. Safer to initialize `self._dbt_yml_path = None` in `__init__` and guard in `load()`:
> 
> ```python
> def __init__(self):
>     self._dbt_yml_path: Optional[Path] = None
>     self._doc_blocks: dict[str, str] = {}
>     self._models: dict[str, DbtModelMetadata] = {}
> ```
> 
> Minor, but `detect()`/`load()` contract in the ABC makes this worth hardening.
> 
> ### Minor nits (non-blocking)
> * `bash.exe.stackdump` snuck into `.gitignore` — fine to leave, but is a Windows build artifact, not a project file.
> * `SQL_SPEC` has all-empty dicts with a comment explaining why. Worth adding a `# pragma: no cover` note or a docstring so future contributors don't try to "fix" it.
> * File context lookup by stem means a file named `schema.sql` would match a dbt model named `schema` — unlikely in practice but worth a comment.
> 
> ### Everything else is solid
> * `strip_jinja` + `extract_dbt_directives` before tree-sitter parsing is exactly right
> * `dirnames[:] =` mutation in os.walk (cross-pollinates nicely with PR [feat: speed up folder indexing #80](https://github.com/jgravelle/jcodemunch-mcp/pull/80)'s approach)
> * `@register_provider` decorator pattern is clean and extensible
> * The `FileContext` / `ContextProvider` ABC split is well-designed
> * 27 SQL tests + 58 context provider tests = right level of coverage
> * CONTEXT_PROVIDERS.md is genuinely useful, especially the "Adding a New Provider" guide
> 
> Rebase + the opt-out mechanism are the two things I'd want addressed before merging. The side-effect nit is worth fixing but not a blocker. Thanks for this — SQL + dbt support has been one of the most-requested features.

I ended up merging my context providers branch to my main fork early forgetting it would impact this PR. I'll get it cleaned up and rebase my PR and make the tweaks mentioned above. Thank you!

PR #107, opened 2026-03-16
Add support for extracting const declarations as constants in JS/TS

AntonEriksson978:
## Summary
This PR adds support for extracting `const` variable declarations as constants in JavaScript and TypeScript files. Previously, only certain naming conventions (UPPER_CASE) were recognized as constants. Now the parser properly identifies and indexes `const` declarations, including exported ones, while intelligently skipping function expressions assigned to constants.

## Key Changes
- **Enhanced constant extraction in `extractor.py`**: Added logic to handle `lexical_declaration` and `variable_declaration` nodes in JS/TS, extracting const assignments as constants
- **Smart function detection**: Arrow functions and function expressions assigned to constants are correctly skipped and handled by the existing `_extract_variable_function` logic instead
- **Relaxed naming conventions for JS/TS**: JavaScript and TypeScript now accept any identifier name for constants (including camelCase), while other languages still enforce UPPER_CASE conventions
- **Test coverage**: Added comprehensive tests for both TypeScript and JavaScript const extraction, including edge cases with exported constants and function expressions
- **Fixture updates**: Updated sample files to include real-world examples of const declarations (Convex query/mutation patterns)

## Implementation Details
- The extraction logic checks the `value_node` type to distinguish between simple constant assignments and function expressions
- Function expressions (arrow functions, function expressions, generator functions) are explicitly skipped to avoid duplication with function extraction
- Content hashing and proper line/byte offset tracking are maintained for all extracted constants
- The implementation respects language-specific naming conventions while providing sensible defaults for JS/TS

https://claude.ai/code/session_01MzbXTW1mdugr134hQ6YMop

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -152,13 +152,14 @@ def test_typescript_const_declarations():
     assert "defaultConfig" in names
     assert "emailRegex" in names
     assert "allowedOrigins" in names
-    # Arrow function should be extracted as a function, not a constant
-    const_names = {s.name for s in symbols if s.kind == "constant"}
-    assert "greet" not in const_names
-    # Verify kind
+    # Verify non-function consts are kind=constant
     for s in symbols:
         if s.name in ("defaultConfig", "emailRegex", "allowedOrigins", "MAX_TIMEOUT"):
             assert s.kind == "constant"
+    # Arrow function const is indexed as a function (not a constant)
+    assert "greet" in names
+    greet_sym = next(s for s in symbols if s.name == "greet")
+    assert greet_sym.kind == "function"
 
 
 JS_CONST_SOURCE = """\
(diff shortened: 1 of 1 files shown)


jgravelle:
Thanks for the detailed follow-up @AntonEriksson978 — that narrows it down perfectly.

You're right: the Convex pattern is a distinct case that's currently a gap. To be precise:

```typescript
export const getTaskList = query({ handler: async (ctx, args) => { ... } });
```

The right-hand side is a **call expression** (`query({...})`), not an arrow function or function expression. The current parser skips these because it can't tell at parse time whether `query()` returns a callable — it could be anything.

**Your PR as written** handles non-function constants (objects, primitives, regex) and correctly skips arrow/function expressions to avoid double-extraction. But it also skips call expressions, so the Convex case wouldn't be caught.

**The actual fix** would be: when a `const` is assigned to a `call_expression`, extract it as a `function` symbol (since in practice these are always endpoint/handler wrappers). This is a small addition either to your PR or to main.

A few options:
1. **Update this PR** to also extract `call_expression` assignments as `function` symbols — would catch Convex `query`/`mutation`/`action` and similar patterns in other frameworks (tRPC, Hono, etc.)
2. **Close this PR** and I'll land a targeted fix on main that covers the Convex case specifically
3. **Merge this PR as-is** (non-function const extraction is still useful) and follow up with a separate issue for the Convex pattern

What would you prefer? Happy to implement option 2 if you don't want to continue iterating on this. Either way, appreciate you tracking this down — Convex users will thank you.

PR #143, opened 2026-03-21
fix: shutdown WAL checkpoint now works for hyphenated GitHub owners

MariusAdrian88:
## Summary

  - `IndexStore.close()` used `slug.split('-', 1)` to reverse-engineer owner/name from `*.db` filenames, then reconstructed the path via `_db_path()`.
  This silently failed for GitHub owners with hyphens (e.g., `my-org/my-repo`), meaning the shutdown WAL checkpoint never fired for those repos.
  - Added `checkpoint_db(db_path)` to `SQLiteIndexStore` that accepts the database path directly, bypassing the fragile owner/name round-trip.
  - `IndexStore.close()` now passes each discovered `*.db` file straight through — reliable for all repo types (local, GitHub, hyphenated owners).

  ## Test plan

  - [x] Existing `test_checkpoint_compacts_wal` still passes
  - [x] New `test_checkpoint_db_by_path` — verifies `checkpoint_db()` works with a path directly
  - [x] New `test_index_store_close_checkpoints_all` — verifies `IndexStore.close()` checkpoints without slug parsing
  - [x] Full suite: 825 passed, 0 failures

### DIFF
--- a/src/jcodemunch_mcp/storage/index_store.py
+++ b/src/jcodemunch_mcp/storage/index_store.py
@@ -217,12 +217,7 @@ class IndexStore:
         if not self.base_path.exists():
             return
         for db_file in self.base_path.glob("*.db"):
-            slug = db_file.stem  # e.g. "local-myrepo-abc123"
-            # Extract owner and name from the slug: first segment = owner, rest = name
-            parts = slug.split("-", 1)
-            if len(parts) == 2:
-                owner, name = parts
-                self._sqlite.checkpoint_and_close(owner, name)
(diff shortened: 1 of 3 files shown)


jgravelle (approved):
Exactly the right fix — adding checkpoint_db() that accepts a path directly is cleaner than any attempt to make the slug parsing more robust. The close() simplification from 6 lines to 1 says it all.

Two targeted tests that verify the actual behavior (WAL compacted to 0 bytes) rather than just calling the method — good test discipline.

Thanks for the quick turnaround on this.

PR #150, opened 2026-03-22
feat: add Razor .cshtml mixed-language parsing and fixture coverage

outoftheblue9:
## Summary

  - Adds a dedicated `CshtmlExtractor` for parsing ASP.NET Core Razor (`.cshtml`) files, which embed C# within HTML
  templates
  - Registers `.cshtml` as a recognized extension mapped to the `razor` language profile
  - Adds two fixture files and a full test suite validating symbol extraction

  ## Motivation

  `.cshtml` files are ubiquitous in .NET web projects and contain a mix of C# logic and HTML markup. Without
  language-aware parsing, jCodeMunch treats them as opaque blobs — none of the symbols, methods, or directives are
  indexed. This PR closes that gap so .NET Razor projects get the same token-savings benefits as other supported
  languages.

  ## What changed

  | File | Change |
  |------|--------|
  | `src/jcodemunch_mcp/parser/extractor.py` | New `_parse_razor_symbols()` function (~300 lines) handling `@{ }` code
  blocks, `@functions { }` / `@code { }` blocks, C# method/class extraction, `@model` / `@using` / `@inject` directives,
   `<script>` and `<style>` blocks, and HTML element ID extraction |
  | `src/jcodemunch_mcp/parser/languages.py` | Registers `.cshtml` → `razor` in the extension map |
  | `src/jcodemunch_mcp/server.py` | Minor follow-on update |
  | `tests/fixtures/razor/sample.cshtml` | Minimal fixture for unit tests |
  | `tests/fixtures/razor/dotnet_aspnetcore_layout.cshtml` | Full real-world ASP.NET Core layout page fixture |
  | `tests/test_razor.py` | 151-line test file covering symbol extraction, directive parsing, method detection, index
  integration, and edge cases |

  ## Test plan

  - [ ] `pytest tests/test_razor.py` — all Razor-specific tests pass
  - [ ] `pytest tests/test_server.py` — no regressions in server tests
  - [ ] Manually verify `.cshtml` files are discovered and indexed via `index_folder`

  ## Notes

  No new dependencies introduced. The extractor uses only stdlib `re` patterns, consistent with the existing custom
  extractor approach used for Vue, Nix, and other languages in this codebase.

### DIFF
--- a/src/jcodemunch_mcp/parser/extractor.py
+++ b/src/jcodemunch_mcp/parser/extractor.py
@@ -30,6 +30,8 @@ def parse_file(content: str, filename: str, language: str) -> list[Symbol]:
         symbols = _parse_elixir_symbols(source_bytes, filename)
     elif language == "blade":
         symbols = _parse_blade_symbols(source_bytes, filename)
+    elif language == "razor":
+        symbols = _parse_razor_symbols(source_bytes, filename)
     elif language == "nix":
         symbols = _parse_nix_symbols(source_bytes, filename)
     elif language == "vue":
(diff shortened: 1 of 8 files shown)


jgravelle (approved):
Really solid contribution — Razor/cshtml is a genuine gap for .NET users and this fills it cleanly.

A few things I appreciated:
- Stays true to the stdlib-regex approach used for Vue, Nix, and EJS — no new dependencies
- Two fixtures: a minimal unit-test case AND a real ASP.NET Core layout page, so the tests have actual signal
- The @functions/@code block distinction is handled correctly, and the @model/@using/@inject directives are first-class symbols rather than an afterthought

Full test suite (852 tests) passes clean. Merging.

Issue #151, opened 2026-03-22
fix broken index cache + optimize BM25 warm-path (~90% faster repeat searches)


MariusAdrian88:
## Problem

Two performance issues in the search pipeline:

### 1. In-memory LRU cache never hits (v1.10.0 regression)

The cache restored in PR #148 has a key mismatch: `save_index` and `incremental_save` store entries with raw `name`, but `load_index` looks them up with sanitized `safe_name` (via `_safe_repo_component`). This means every tool call pays the full SQLite cold-read cost (~110ms on a 5,800-symbol repo), defeating the entire cache optimization.

### 2. BM25 scoring recomputes everything on every search

Even with a working cache, `search_symbols` recomputes three query-independent operations on every call:
- IDF/avgdl corpus stats — tokenizes all symbols (~58K regex operations on repo_A)
- Centrality scores — iterates all imports
- Per-symbol token bags — `_sym_tokens()` called redundantly in both `_compute_bm25` and `_bm25_score`

Additionally, the two-pass pipeline (heuristic pre-filter -> BM25 re-rank) iterates the full symbol list twice and sorts all candidates before truncating to `max_results`.

## Impact

On repo_A (5,847 symbols) and repo_B (1,782 symbols), every `search_symbols` call takes ~110-200ms regardless of whether the index changed. For AI agents generating tool calls at 30-40 tokens/second, a typical 15-call exploration sequence wastes ~1.7 seconds on redundant computation.

## Proposed Fix

1. **Fix cache key consistency** — `save_index`, `incremental_save`, and `delete_index` use `safe_name` for cache operations, matching `load_index`
2. **Cache IDF/avgdl/centrality on CodeIndex** — computed once per index load, reused across all queries
3. **Cache per-symbol token bags** — `_sym_tokens()` result stored on the symbol dict, eliminating ~58K regex calls per search
4. **Single-pass BM25 with bounded heap** — replaces the two-pass heuristic + BM25 pipeline with one pass over `index.symbols`, using a min-heap of size `max_results`

## Expected Results

| Scenario | Before | After | Improvement |
|---|---|---|---|
| Cache hit (index load) | ~110ms every call | ~0ms (cache hit) | ~100% |
| Warm search (repeat query, same index) | ~116ms | ~25-35ms | ~75-80% |
| Cold search (first after reindex) | ~197ms | ~197ms | ~0% (same work, front-loaded) |
[...]

PR #152, opened 2026-03-22
perf: search up to 25x faster, find_references 14x, watcher reindex 30x (3s → 96ms)

MariusAdrian88:
## Summary

End-to-end performance optimization across the three hottest paths: BM25 symbol search, import reference lookups, and watcher-driven reindexing. Benchmarked on a 5,847-symbol / 365-file Python repo (repo_A).

This is a BIG PR... maybe too BIG.

### 1. Watcher reindex: 2,903ms → 96ms (30x)

**The problem.** The file watcher debounce is 200ms, but every reindex took ~3 seconds — even when nothing changed. Profiling revealed:

| Step | Time | What it does |
|------|------|-------------|
| `discover_local_files` | 2,748ms | Walks entire directory tree |
| `Path.resolve()` x 16,796 | 1,981ms | NTFS path resolution per candidate file |
| `discover_providers` | 520ms | Scans for dbt/terraform ecosystem markers |
| stat + filter + change detection | ~100ms | The actual useful work |

The watcher already knows exactly which files changed (from `watchfiles` events), but was throwing that information away and calling `index_folder` which re-walked 16,796 files from scratch.

**The fix.** Added a `changed_paths` parameter to `index_folder`. When the watcher provides it:

- Skip `discover_local_files` entirely — no directory walk, no `Path.resolve()` on 16K files
- Skip `discover_providers` — providers don't change between file edits
- Hash-check before parsing — if content hash matches stored hash (touch/mtime-only), update mtime in DB and return early
- Only read/parse/save the 1-2 affected files

| Scenario | Before | After | Speedup |
|----------|--------|-------|---------|
| Touch / mtime-only (most common) | 2,903ms | 96ms | 30x |
| Real edit of largest file (150KB, 178 symbols) | 2,903ms | 217ms | 13x |
| MCP tool call / initial index (fallback) | 2,903ms | 2,903ms | unchanged |

### 2. Filtered symbol search: up to 25x faster

**The problem.** `search_symbols` with `kind`, `file_pattern`, or `language` filters was falling back to a full scan of all symbols, bypassing the inverted index entirely.

**The fix.** The inverted index now narrows candidates first, then filters are applied on the smaller set. Falls back to full scan only when no query terms match any symbol (preserves centrality-only results).

### 3. find_references: 10-14x faster

**The problem.** Every `find_references` call iterated all files' imports, then all names per import — O(files x imports) brute force.

**The fix.** Build an inverted index on import names (cached on `CodeIndex`, built lazily on first call). Lookups are O(1).

### 4. BM25 cold path: 13-17% faster

Pre-compiled `_tokenize` regexes at module level (called ~9,000x during cold BM25 build). Baked min-length check into the regex pattern itself.

### 5. Bug fixes

- `get_symbol()` was leaking internal `_tokens`/`_tf`/`_dl` cache keys into API responses after the first search warmed the BM25 cache. Added `_INTERNAL_KEYS` stripping + `_get_symbol_raw()` for internal use.
- `save_index` / `incremental_save` / `delete_index` used raw `name` for cache keys but `load_index` used `safe_name` — permanent cache misses for repos with special characters.

### 6. Other optimizations

- `_build_index_from_rows`: single pass over file rows instead of 7 separate comprehensions
- `search_text`: pre-compiled fnmatch pattern, inline byte estimation (dropped `json.dumps` for size counting)
- `get_symbol` tool: compute `_content_dir` once instead of twice
- `_bm25_breakdown`: use cached `_dl`, pre-compute `query_set` outside per-field loop

## End-to-end profiling: all tools

Benchmarked on repo_A (5,847 symbols, 365 files). 5 passes x 5 iterations, median of medians.
But this data cames after many optimizations already included in commits in this PR. It just compares an already optimized state with the latest state. I know it is maybe not that relevant as comparison but the actual numbers are. I have lost the previous benchmarks so this is what it is.
We are talking now on duration's on the order milliseconds or tens of milliseconds for all tools. The re-indexing is the one that cost the most but even that is optimized (~220ms for a large file) as noted in the bottom note.

| Tool | Before (ms) | After (ms) | Delta | Speedup |
|------|-------------|------------|-------|---------|
| **reindex after file write** | 3135.74 | **3040.88** | -3% | — |
| list_repos | 4.62 | 4.29 | -7% | — |
| get_repo_outline | 10.13 | 7.04 | -30% | 1.4x |
| get_file_tree | 12.09 | 12.12 | 0% | — |
| get_file_tree (prefix) | 0.17 | 0.16 | -6% | — |
| get_file_outline | 3.47 | 2.73 | -21% | 1.3x |
| get_file_content | 1.13 | 1.23 | +9% | — |
| get_session_stats | 0.35 | 0.35 | 0% | — |
| get_symbol | 0.94 | 0.98 | +4% | — |
| get_symbols (3) | 0.85 | 0.86 | +1% | — |
| get_context_bundle | 4.09 | 4.63 | +13% | — |
| get_context_bundle (callers) | 14.55 | 12.83 | -12% | 1.1x |
| search_symbols | 1.19 | 1.19 | 0% | — |
| **search_symbols (kind)** | 3.16 | **0.27** | **-91%** | **11.7x** |
| **search_symbols (file)** | 24.65 | **0.96** | **-96%** | **25.7x** |
| search_symbols (compact) | 0.90 | 0.57 | -37% | 1.6x |
| search_symbols (full) | 4.68 | 5.24 | +12% | — |
| search_text | 67.90 | 66.07 | -3% | — |
| search_text (pattern) | 72.08 | 66.43 | -8% | 1.1x |
| search_text (regex) | 9.82 | 9.96 | +1% | — |
| search_text (context) | 17.51 | 17.27 | -1% | — |
| search_columns | 0.23 | 0.17 | -26% | — |
| suggest_queries | 18.45 | 14.05 | -24% | 1.3x |
| find_importers | 16.73 | 16.26 | -3% | — |
| **find_references** | 2.35 | **0.17** | **-93%** | **13.8x** |
| **find_references (batch)** | 2.63 | **0.21** | **-92%** | **12.5x** |
| check_references | 21.29 | 20.14 | -5% | — |
| get_blast_radius | 13.61 | 12.61 | -7% | — |
| get_class_hierarchy | 2.04 | 2.64 | +29% | — |
| get_dependency_graph | 8.86 | 8.17 | -8% | — |
| get_related_symbols | 27.97 | 25.37 | -9% | — |

Note: "reindex after file write" in the table above uses the standard path (no `changed_paths`). With the watcher fast path, it drops to **96ms** (touch) / **217ms** (real edit).

Small regressions in `get_context_bundle` (+13%), `get_class_hierarchy` (+29%), `search_symbols (full)` (+12%) are from the `get_symbol()` dict-copy overhead that strips internal BM25 cache keys from API responses.

## Commits

- `65f7beb` fix: use safe_name for cache keys in save/incremental_save/delete
- `94591e6` test: add BM25 search regression + perf baseline
- `4ba51bb` perf: cache per-symbol token bags to avoid re-tokenization
- `8b4a0f8` perf: cache BM25 IDF/avgdl/centrality on CodeIndex instance
- `ace93fb` perf: single-pass BM25 with bounded heap, drop heuristic pre-filter
- `4be4d31` test: finalize BM25 perf regression tests, clean up temp files
- `bedd8a0` refactor: hoist imports, remove redundant counter, add safety comment
- `dcc1729` perf: carry forward BM25 token bags across incremental reindex
- `17230d6` perf: inverted index + pre-computed TF for ~90% faster warm searches
- `22100ed` fix: strip internal BM25 keys from API responses, use inverted index with filters
- `5247a74` perf: pre-compile regexes, single-pass file rows, inverted import index
- `2ae9f07` fix: revert search_text to `in` operator for substring mode
- `45a67c0` perf: use _get_symbol_raw in search, cache dl in breakdown, dedupe content_dir
- `1fd5d8f` perf: watcher fast path skips full directory discovery (3s → 96ms)

Solves Issue #151


jgravelle (approved):
Marius, this is genuinely impressive work. The diagnosis is thorough, the fixes are surgical, and the numbers speak for themselves.

A few things that stood out:

The watcher fast path is the right architecture — the watcher already knew which files changed and we were throwing that away on every reindex. The changed_paths parameter is the clean solution: it threads that knowledge all the way into index_folder without contaminating the normal code path.

The safe_name cache key bug is the kind of thing that causes permanent misses that are nearly impossible to notice — nice catch. Same with the _INTERNAL_KEYS BM25 leak into API responses; that was a correctness issue hiding in plain sight.

Carrying forward token bags across incremental reindexes is clever — avoids re-tokenizing symbols that haven't changed and keeps BM25 warm through edits.

The benchmark table is a good format for the PR record. The missing pre-optimization baseline for some rows is noted but understandable — the individual commit benchmarks tell the story.

856 tests passing. Merging.

PR #163, opened 2026-03-24
feat: merge get_symbol+get_symbols into get_symbol_source

MariusAdrian88:
## Summary

- **Merges `get_symbol` and `get_symbols` into a single `get_symbol_source` tool** — reduces tool count from 27 to 26 and eliminates the choice between two identically-scoped tools
- **Shape-follows-input:** `symbol_id` (string) returns a flat symbol object; `symbol_ids` (array) returns `{symbols, errors}` — agents get the natural response shape for each use case without extra nesting
- **Batch mode now supports `verify` and `context_lines`** — previously only available on single-symbol calls
- **Mutual exclusion enforced:** passing both `symbol_id` and `symbol_ids` returns an error
- **Fixed config template bug:** missing comma after the `descriptions` block caused a JSON parse error when any key in subsequent sections (Transport, Watcher, etc.) was uncommented — silently fell back to all defaults
- **Improved `disabled_tools` template:** tool names now live inside the list as commented entries — just uncomment to disable, no separate reference block needed

## Files changed

| File | Change |
|------|--------|
| `src/jcodemunch_mcp/tools/get_symbol.py` | Replaced `get_symbol` + `get_symbols` with `get_symbol_source` |
| `src/jcodemunch_mcp/server.py` | One Tool entry + one dispatcher branch (was two each) |
| `src/jcodemunch_mcp/config.py` | Template: comma fix + disabled_tools as inline commented list |
| `tests/test_cleanup.py` | Updated + 2 new mutual-exclusion tests |
| `tests/test_server.py` | Updated tool count assertions and tool name references |
| 9 doc files | All references updated (AGENT_HOOKS, ARCHITECTURE, CLAUDE, SPEC, USER_GUIDE, README, QUICKSTART, TOKEN_SAVINGS, INSTALL_FROM_SOURCE) |

## Test plan

- [x] All 1104 tests pass (`pytest tests/ -q`)
- [x] `get_symbol_source` with `symbol_id` returns flat object
- [x] `get_symbol_source` with `symbol_ids` returns `{symbols, errors}`
- [x] Passing both `symbol_id` and `symbol_ids` returns error
- [x] Passing neither returns error
- [x] Config template parses cleanly after `_strip_jsonc` with any key uncommented


jgravelle (approved):
Excellent work, Marius! The shape-follows-input design is exactly right — agents get the natural response shape without extra nesting, and closing the verify/context_lines gap in batch mode was a long-overdue fix. The config comma bug is a real catch; silent fallback to defaults is the worst kind of failure. Docs are thorough and consistent throughout. Merging as 1.11.0.

PR #179, opened 2026-03-28
perf: retire unused staleness tools/fields; fix watcher config layering, hash-cache skip, test isolation, and config-driven watcher

### LINKED ISSUE #177: Index stale
Hi,

Was something changed to the indexing done by the file watcher?
Because it used to work very well, yet today I am seeing a lot of switching back to GREP due of index stale issues...
Retrieving data with jcodemunch is no longer in sync with the files state after EDIT.

Looks like there is a big regression in that area, I do not have now time to investigate/fix, just had to report it.

An example:

```
jcodemunch is named jcode in here (simpler)

All 124 tests pass. Let me verify the final state of the changes:

⚙ jcode_get_file_content [end_line=45, file_path=.tests/sqlite_migration/test_database_operators.py, 
repo=local/REPO-acfbc7e7, start_line=1]

Thinking: Hmm, the Qt patch I added earlier seems to have been removed or overwritten. Let me check 
what's happening. Let me search for the patch in the file.

# Check if Qt patch is in test file
$ grep -n "QEventLoop\|QtCore" .tests/sqlite_migration/test_database_operators.py | head -10
14:from PySide6 import QtCore
18:_app = QtCore.QCoreApplication.instance()
20:    _app = QtCore.QCoreApplication(_sys.argv)
22:# Patch QEventLoop to prevent hangs in tests
31:QtCore.QEventLoop = _MockEventLoop

Thinking: The Qt patch is there at lines 14-31. The jcode tool showed lines 1-45 but it seems the patch might 
be after the imports. Let me check the actual content around line 13-35.

⚙ jcode_get_file_content [end_line=50, file_path=.tests/sqlite_migration/test_database_operators.py, repo=local/REPO-acfbc7e7, start_line=10]

Thinking: That's strange - the jcode tool is showing different content than grep. Let me check the file directly 
with bash.

# Check actual file content
$ sed -n '10,40p' .tests/sqlite_migration/test_database_operators.py
Coverage target: >90% for Database.py
"""
# MUST PATCH QT BEFORE ANY OTHER IMPORTS TO PREVENT HANGS
from PySide6 import QtCore
import sys as _sys
# Ensure QCoreApplication exists for QObject creation
_app = QtCore.QCoreApplication.instance()
if _app is None:
    _app = QtCore.QCoreApplication(_sys.argv)
…
Click to expand

Thinking: Good, the Qt patch is there. The jcode index might be stale. Let me trigger a refresh and then 
provide the final summary.
```

LE: BTW, I do not use the hooks for indexing, because after profiling against the previous file watcher indexing (the one that worked), the hooks indexing lost in the performance battle. At least for my use case.

MariusAdrian88:
## Summary

Remove client-facing staleness machinery (2 MCP tools + 3 _meta fields) that no client consumed, fix three watcher regressions introduced since 979f5b5, and make all watcher parameters configurable via `config.jsonc` so users can drop CLI flags from their MCP client configs.

### 1. Retire staleness _meta injection and client-facing freshness tools

**Problem:** The previous design exposed reindex state to clients via:
- Three fields injected into every tool response `_meta`: `index_stale`, `reindex_in_progress`, `stale_since_ms`
- Two dedicated MCP tools: `wait_for_fresh` and `check_freshness`

No client ever acted on these fields — they were annotated noise on every single response.

**Token cost:** Two extra tools in the schema cost ~400 tokens per call just in definitions. Three staleness fields in `_meta` added ~30-50 tokens per response. In a typical 50-100 tool call session: 25,000-45,000 wasted tokens for information nobody was consuming.

**Better model:** Freshness management belongs on the server side. The config key `freshness_mode` already covers both cases:
- `"relaxed"` (default) — serve cached index immediately, zero overhead
- `"strict"` — block silently in `call_tool` via `await_freshness_if_strict()` until watcher finishes, then serve fresh results

Either way, the client receives a clean response. The server decides transparently.

**Removed:**
- `wait_for_fresh` MCP tool (schema + `call_tool` handler)
- `check_freshness` MCP tool + `tools/check_freshness.py`
- Entire `_meta` staleness injection block in `call_tool` (~50 lines)
- `_EXCLUDED_FROM_STRICT` entry for `wait_for_fresh`
- 670 lines of tests covering the removed surface area
- `get_reindex_status` / `wait_for_fresh_result` imports from `server.py`

**Preserved:**
- `reindex_state.py` machinery intact (watcher still needs it internally)
- `await_freshness_if_strict` in `call_tool` (strict mode still works)
- `freshness_mode` config key and `JCODEMUNCH_FRESHNESS_MODE` env var
- `meta_fields` config for partial `_meta` field selection

### 2. Fix watcher config layering

`_get_watcher_enabled()` bypassed `config_module.get()` and read `JCODEMUNCH_WATCH` env var directly, then fell back to hardcoded `False`. The config file `"watch"` key was silently ignored.

**Fix:** CLI flag > config (with env var as fallback only when key absent from `config.jsonc`).

### 3. Fix hash-cache reindex bug (silent skips)

When the watcher's in-memory hash cache missed, the fallback read the file from disk. But by the time `watchfiles` delivers the event the file already has new content, making `old_hash == new_hash` and silently skipping the change.

**Fix:** Use a `"__cache_miss__"` sentinel that keeps the fast path active (<200ms) while guaranteeing the file is re-parsed rather than skipped.

### 4. Fix flaky tests from SQLite WAL cache contamination

Tests that modify the DB directly (`index_version`, `indexed_at`, corruption) didn't invalidate the in-memory LRU cache. On Windows, WAL mode doesn't always update the DB file mtime on write, so the cache key still matched.

**Fix:** Added `tests/conftest.py` with autouse fixtures for cache clear and config reset, plus targeted `_cache_evict()` calls after direct DB modifications. Also fixed `test_openai_summarizer_timeout_config` that was reading `config.get()` for `allow_remote_summarizer` but the env var never flowed through `load_config()`.

### 5. Config-driven watcher parameters

**Problem:** Even though `"watch": true` in `config.jsonc` enables the embedded watcher, all other watcher parameters (`debounce`, `paths`, `idle_timeout`, `extra_ignore`, `follow_symlinks`, `log`) were only sourced from CLI flags. This forced verbose flags in MCP client configs:

```jsonc
// Before: every param must be a CLI flag
"command": ["uvx", "--with", "jcodemunch-mcp[watch]", "jcodemunch-mcp", "serve",
            "--watcher", "--watcher-no-ai-summaries"]
```

**Fix:** The `watcher_kwargs` block in `main()` reads each parameter as: **CLI flag > config key > default**. Same layering pattern already used for `watch` and `freshness_mode`.

New config keys (alphabetical after `watch`):
- `watch_debounce_ms` (default: 2000) — existed in config.py but was not wired to watcher_kwargs
- `watch_extra_ignore` — list of gitignore-style patterns (default: `[]`)
- `watch_follow_symlinks` — bool (default: `false`)
- `watch_idle_timeout` — minutes, `null` = disabled
- `watch_log` — file path or `"auto"` (default: `null`)
- `watch_paths` — list of folder paths (default: `["."]` i.e. CWD)

CLI flags (`--watcher-*`) remain as overrides for all keys.

**Result:** Users configure the watcher entirely in `config.jsonc`:

```jsonc
// config.jsonc
{
  "watch": true,
  "use_ai_summaries": false,
  "watch_debounce_ms": 2000
}
```

```jsonc
// MCP client config — simplified
"command": ["uvx", "--with", "jcodemunch-mcp[watch]", "jcodemunch-mcp"]
```

Note: `jcodemunch-mcp[watch]` extra is still required to install `watchfiles`. The server exits with a clear error if `watch: true` is set but `watchfiles` is not installed.

## Impact

| Metric | Before | After |
|--------|--------|-------|
| Schema tokens per call | +~400 (2 unused tools) | 0 |
| _meta tokens per response | +~30-50 (3 staleness fields) | 0 |
| Lines of code | +880 | -880 (net deletion) |
| Tests passing | 1256 | 1281, 0 regressions |
| Watcher config layering | Broken (config file ignored) | Fixed |
| Hash-cache miss handling | Silent skip | Re-parse guaranteed |
| Test isolation (Windows) | Flaky from cache bleed | Clean via autouse fixtures |
| Watcher config | CLI flags only | config.jsonc with CLI override |

## Files changed

```
# Deleted
src/jcodemunch_mcp/tools/check_freshness.py  deleted (108 lines)
tests/test_backpressure_regression.py        deleted (361 lines)
tests/test_check_freshness.py                deleted (144 lines)
tests/test_wait_for_fresh.py                 deleted (149 lines)

# Modified / added
AGENT_HOOKS.md                               doc cleanup (−2 lines)
CONFIGURATION.md                             new watcher config key docs (+11/−1)
README.md                                    doc cleanup (+2/−1)
SPEC.md                                      remove staleness/freshness tool specs (−92 lines)
USER_GUIDE.md                                doc cleanup (−5 lines)
src/jcodemunch_mcp/config.py                 new watcher config keys (+20)
src/jcodemunch_mcp/server.py                 staleness removal + config-driven watcher kwargs (−97/+39)
src/jcodemunch_mcp/watcher.py                cache miss sentinel (+19)
tests/conftest.py                            autouse cache/config fixtures (+48, new)
tests/test_config_watcher_keys.py            config key defaults/types (+36, new)
tests/test_find_importers.py                 isolation fix (+8)
tests/test_hardening.py                      isolation fix (+4)
tests/test_retrieval_tools.py                isolation fix (+4)
tests/test_server.py                         tool removal assertions (+19)
tests/test_storage.py                        isolation fix (+3)
tests/test_summarizer.py                     config flow fix (+10)
tests/test_watcher_memory_cache.py           cache miss sentinel tests (+64, new)
tests/test_watcher_serve.py                  config layering + CLI override tests (+304/−45)
```

Fixes #177


jgravelle (approved):
Solid contribution — five distinct improvements in one well-scoped PR.

The staleness tool retirement is the right call. Two tools costing ~400 tokens/call that no client ever acted on is pure overhead. The server-side freshness model (strict/relaxed via config) is cleaner and the _meta noise reduction is a real win.

The watcher config-layering fix is long overdue — silent config-file ignore was a confusing UX footgun. The new CLI > config > default precedence is exactly right, and the new watch_* config keys (paths, debounce, idle_timeout, extra_ignore, follow_symlinks, log) make the watcher genuinely usable without flag soup in MCP client configs.

The __cache_miss__ sentinel fix is subtle but important — the old behavior (read disk hash, compare against new content, conclude 'no change') was a correctness bug, not just a perf issue.

The conftest autouse fixtures for Windows WAL cache bleed are the right fix. Direct DB writes bypassing the LRU cache is exactly the class of flakiness that's hard to chase down without a systematic reset.

1281 tests, 0 regressions, net -880 lines. Approved.

PR #180, opened 2026-03-28
fix: cross-process LRU cache invalidation + config template improvements

MariusAdrian88:
## SUMMARY

SQLite WAL mode does not always update the main .db file mtime on commit. When the watcher (separate process) reindexes a file, the MCP server's in-memory LRU cache was serving stale data because its mtime-based cache key appeared unchanged. This caused agents to see outdated jcode results after editing files, leading to unnecessary fallbacks to grep/bash.

# Cache invalidation (hybrid A+C):
- Add _db_mtime_ns() helper: checks max(db, db-wal) mtime so the reader detects WAL writes from any external process
- Add os.utime(db_path) after save_index() and incremental_save() to force .db mtime update as a belt-and-suspenders measure
- Fix ordering: utime runs BEFORE _cache_put() so the cached mtime matches what cross-process readers compute via _db_mtime_ns()
- Wrap os.utime() in try/except OSError (best-effort, non-fatal)

# server.py bug fix:
- Wire max_files parameter through call_tool dispatch for get_file_tree (was silently ignored despite being in the MCP schema)

# get_file_tree hardening:
- max_files now defaults to None; resolved as: per-call arg → config file_tree_max_files → hardcoded 500 fallback
- Add max(1, max_files) guard against 0 or negative values

# Config system additions:
- file_tree_max_files: configurable cap for get_file_tree results
- gitignore_warn_threshold: configurable threshold for the missing .gitignore warning in index_folder (set 0 to disable)
- Both added to ENV_VAR_MAPPING, DEFAULTS, CONFIG_TYPES, and template

# Config template improvements:
- Languages list: each language on its own line, active (not commented)
- version field added to track config template version
- Added inline documentation for all config keys
- Added missing keys: trusted_folders_whitelist_mode, exclude_secret_patterns, path_map, and watcher params
- Removed stale tools (wait_for_fresh) and meta fields from template
- meta_fields and disabled_tools lists sorted alphabetically

Tests:
- test_db_mtime_ns_no_wal: returns .db mtime when WAL absent
- test_db_mtime_ns_wal_newer: returns WAL mtime when WAL is newer
- test_db_mtime_ns_wal_older: returns .db mtime when DB is newer
- test_cross_process_cache_invalidation: injects stale cache entry with old mtime, verifies load_index detects mtime change without explicit cache eviction (tests the actual cross-process scenario)
- test_template_languages_synced_from_registry: updated to verify active (not commented) entries via JSONC parsing

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>
Co-Authored-By: Minimax M2.7 <noreply@minimaxi.com>

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -26,6 +26,8 @@ ENV_VAR_MAPPING = {
     "JCODEMUNCH_MAX_INDEX_FILES": "max_index_files",
     "JCODEMUNCH_STALENESS_DAYS": "staleness_days",
     "JCODEMUNCH_MAX_RESULTS": "max_results",
+    "JCODEMUNCH_FILE_TREE_MAX_FILES": "file_tree_max_files",
+    "JCODEMUNCH_GITIGNORE_WARN_THRESHOLD": "gitignore_warn_threshold",
     "JCODEMUNCH_EXTRA_IGNORE_PATTERNS": "extra_ignore_patterns",
     "JCODEMUNCH_EXTRA_EXTENSIONS": "extra_extensions",
     "JCODEMUNCH_CONTEXT_PROVIDERS": "context_providers",
@@ -60,6 +62,8 @@ DEFAULTS = {
     "max_index_files": 10000,
     "staleness_days": 7,
     "max_results": 500,
+    "file_tree_max_files": 500,
+    "gitignore_warn_threshold": 500,
     "extra_ignore_patterns": [],
     "exclude_secret_patterns": [],
     "extra_extensions": {},
@@ -99,6 +103,8 @@ CONFIG_TYPES = {
     "max_index_files": int,
     "staleness_days": int,
     "max_results": int,
+    "file_tree_max_files": int,
+    "gitignore_warn_threshold": int,
     "extra_ignore_patterns": list,
     "exclude_secret_patterns": list,
     "extra_extensions": dict,
@@ -128,6 +134,7 @@ CONFIG_TYPES = {
     "summarizer_concurrency": int,
     "allow_remote_summarizer": bool,
     "path_map": str,
+    "version": str,
 }
 
 
@@ -299,6 +306,7 @@ def load_config(storage_path: str | None = None) -> None:
                                         "Config key 'trusted_folders' contains non-absolute path "
                                         f"'{folder}'"
                                     )
+
                             _GLOBAL_CONFIG[key] = list(valid_folders)
                         else:
                             _GLOBAL_CONFIG[key] = value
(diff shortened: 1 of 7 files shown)


jgravelle (commented):
## Review

**Core bug fix (sqlite_store.py)** — Well-targeted and correct.

The `_db_mtime_ns()` helper cleanly solves the WAL mtime problem. The `os.utime()` before `_cache_put()` with documented ordering rationale is solid defensive engineering. Tests cover all three branches (no WAL, WAL newer, DB newer) plus the end-to-end cross-process scenario.

**One concern: `meta_fields` default is a breaking change**

```jsonc
// Before (template):
"meta_fields": null,     // → all fields included (backward-compatible)

// After (template):
"meta_fields": [],       // → NO metadata in any response
```

Users who regenerate their config or new users copying the template will silently lose all `_meta` from responses. The existing behavior and comment both say `null` = backward-compatible default. Shipping `[]` as the new template default is a breaking UX change that should be called out explicitly in the PR title/description and gated on a major version bump, or reverted to `null`.

**Minor issues**

- `version` field (set to `__version__`) will immediately be stale for anyone who doesn't regenerate config on every release. If it's only used for future migrations, it's premature to add — migrations can check it when they land.

- `test_db_mtime_ns_no_wal` asserts the WAL file doesn't exist after `_connect()`, but depending on SQLite internals and OS, WAL mode initialization may create the `-wal` file. The assert will catch it, but this test could be flaky on some CI configurations.

- `Path(str(db_path) + "-wal")` in production vs `db_path.with_suffix(".db-wal")` in tests — both produce the same path for `.db` files, but it's an inconsistency worth noting.

**Everything else looks good** — the `get_file_tree` `max_files` wire-up fix is a clear correctness bug, the config additions are clean and well-documented, and the template inline docs are a real quality-of-life improvement.

**Summary:** Request changes on the `meta_fields: []` default — revert to `meta_fields: null` (or at minimum flag it as a breaking change in the changelog). Everything else is solid.

jgravelle (approved):
Another strong contribution, Marius. The WAL mtime issue is a real and subtle bug — SQLite WAL mode is well-documented to not flush mtime to the main .db file on every commit, so the old single-file stat() was genuinely unreliable for cross-process cache invalidation. _db_mtime_ns() is the right fix, the utime() belt-and-suspenders is a sensible addition, and the ordering — utime() BEFORE _cache_put() — is the detail that makes it correct rather than just mostly correct.

The server.py max_files passthrough is a real bug fix too. That param has been in the MCP schema since v1.12.1 but was silently swallowed in call_tool dispatch.

Config template cleanup is overdue — removing wait_for_fresh (gone since v1.12.0), sorting the tools/meta lists, and adding all the missing keys/docs makes the --init output actually trustworthy. One behavior note worth documenting: the new default template emits meta_fields: [] (empty list = no _meta) rather than the old meta_fields: null (all fields). It's better for token efficiency but users upgrading via config --init will lose _meta without noticing. Not blocking — just worth a mention in the release notes.

The cross-process cache test is exactly right: inject a stale entry, let load_index detect the mtime delta on its own, verify fresh data comes back without an explicit eviction call. That's the actual scenario we're protecting against.

1302 passed, 7 skipped on Windows/Python 3.10 locally. CI is green on all other matrix legs. Merging.

PR #185, opened 2026-03-29
feat: adaptive language config + review-feedback fixes

MariusAdrian88:
# SUMMARY

Add languages_adaptive config key that auto-manages the languages list in per-project .jcodemunch.jsonc based on detected languages. 

# HOW
- When enabled in the global config and no local config exists, copies the global config to the project and restricts languages to those actually found. 
- When enabled in the local config, surgically updates the languages array on every index_folder call (full and incremental), commenting out unused languages and uncommenting detected ones. 
Fast path (~0ms) when languages are unchanged; surgical update path ~1-5ms.

# Review-feedback fixes from PR #180:
- Fix broken test_db_mtime_ns_no_wal: _connect() creates WAL mode, so use plain sqlite3.connect() for the no-WAL test scenario
- Standardize WAL path construction in 3 tests to match production code (Path(str(db_path) + "-wal") instead of db_path.with_suffix())
- Align meta_fields Python default with template default (None → []) so users with no config file get the same behavior as config --init users
- Fix meta_fields display source detection in config subcommand

New helpers in config.py: 
_global_config_path(), 
_parse_active_languages(), 
_build_languages_block(), 
_check_raw_local_adaptive(), 
_apply_languages_adaptation(), 
invalidate_project_config_cache(), 
apply_adaptive_languages(). 

Hook via _maybe_apply_adaptive() at all 3 index_folder return points (fast-path, incremental, full).

16 new tests in test_adaptive_languages.py; 1341 passed, 4 skipped.

Co-Authored-By: [Executor] Minimax M2.7 <noreply@minimaxi.com>
Co-Authored-By: [Planner] Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -4,6 +4,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import threading
 from copy import deepcopy
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


jgravelle (approved):
Marius — this is excellent work on two fronts, and it's clear you've thought both through carefully.

**Adaptive languages** is a genuinely useful QoL feature. The decision tree (no local + global adaptive → create from copy; local + raw adaptive → surgical update; otherwise no-op) is clean and the invariant that we read the *raw* local value (not the merged config) is exactly the right call — good instinct to document *why* in the docstring. The fast-path guard in `_maybe_apply_adaptive` (never raises, skips on !success) means this hook can't break an index that was otherwise working, which is the right tradeoff.

The regex limitation with `]` inside comments is correctly called out in two places and is not a realistic concern for auto-generated content.

**PR #180 feedback fixes** are all correct:
- `sqlite3.connect()` instead of `store._connect()` for the no-WAL scenario — spot on.
- `Path(str(db_path) + "-wal")` path consistency — matches production code exactly.
- `meta_fields` default `None → []` is a semantic breaking change, but it's the right default for new users (token-efficient by default) and is well-documented in the template comment.

16 new tests, full suite 1338 passed / 7 skipped. Merging.

PR #218, opened 2026-04-07
feat: AST call graph, decorator awareness, negative evidence, framework providers + perf optimizations

MariusAdrian88:
## Summary

This PR addresses several gaps in jcodemunch's code analysis capabilities that caused LLM consumers to hallucinate about features that don't exist, miss cross-cutting concerns like decorators, and produce inaccurate call graphs. It also adds framework-aware route/middleware detection for 18 web frameworks and includes targeted performance optimizations to offset the cost of new features.

**42 files changed** | **5,209 insertions** | **130 deletions** | **2,310 tests passing**

---

## Why

Three problems were identified in real-world usage:

1. **LLM hallucination on missing features.** When `get_ranked_context` or `search_symbols` returned empty or low-confidence results, there was no structured signal telling the LLM "this doesn't exist." The LLM would then fabricate an answer based on nearby-but-unrelated code.

2. **Call graph accuracy was too low to be useful.** The call graph relied entirely on text-heuristic word-boundary regex (`\b` + name + `\b`), which produced false positives from comments, strings, and variable names. Confidence was self-reported as "low."

3. **No framework awareness for most web frameworks.** Route/middleware metadata was only available for Laravel, Next.js, and Nuxt. Projects using Express, Flask, FastAPI, Django, Spring Boot, NestJS, ASP.NET, Go routers, or Rails got no route detection, making endpoint discovery and dead-code analysis incomplete.

Additionally, incremental feature development over several releases had accumulated performance overhead: multiple O(N) passes during indexing, eager computation of indexes that may never be used, and O(N x M) file I/O in dead-code analysis.

---

### 1. AST-based call graph

**Problem:** Call relationships were detected at query time by scanning file contents for symbol names using regex. This was slow (loaded every importer's source) and imprecise (matched names in comments/strings).

**Fix:** Extract `call_expression` AST nodes during tree-sitter parsing and store them as `call_references` per symbol. 13 languages supported including `new_expression` for JS/TS/Java constructors. INDEX_VERSION bumped from 7 to 8 with full backward compatibility (v7 indexes gracefully degrade to text heuristic).

| Aspect | Before | After |
|--------|--------|-------|
| Call detection method | Runtime text regex | Pre-computed AST call_expression nodes |
| Languages with call extraction | 0 | 13 (Python, JS, TS, Go, Rust, Java, PHP, Ruby, C#, Kotlin, Dart, Swift, TSX) |
| Constructor calls (`new Foo()`) | Missed entirely | Captured via `new_expression` / `object_creation_expression` |
| Query-time call graph speed | O(importers x symbols) with file I/O | O(1) index lookup |
| Reported confidence | "low" | "medium" (AST) / "low" (fallback) |
| Caller precision | False positives from comments/strings | Only actual call sites |
| Index storage overhead | 0 | ~100-200 bytes/symbol (JSON array in existing `data` column) |

### 2. Decorator awareness

**Problem:** Tree-sitter already extracted decorators during parsing and stored them in `Symbol.decorators`, but no tool surfaced or filtered by them. Cross-cutting concerns (CSRF protection, auth, rate limiting) were invisible.

**Fix:** Surface decorators in `get_file_outline` and `search_symbols` results. Add `decorator` filter parameter to `search_symbols` (case-insensitive substring match). Add `decorator_filter` parameter to `get_blast_radius`.

| Aspect | Before | After |
|--------|--------|-------|
| Decorator data in get_file_outline | Hidden | Shown per symbol |
| Decorator data in search_symbols results | Hidden | Shown per symbol |
| Filter symbols by decorator | Not possible | `search_symbols(decorator="csrf_protect")` |
| Filter blast radius by decorator | Not possible | `get_blast_radius(decorator_filter="route")` |
| Cross-cutting concern workflow | Manual file-by-file inspection | Set difference: all functions - filtered functions |

### 3. Negative evidence + enforcement signals

**Problem:** When queries didn't match anything, tools returned empty results with no signal. LLMs treated silence as "I didn't look hard enough" and hallucinated implementations. Advisory signals like `budget_warning` were buried in `_meta` where LLMs often ignored them.

**Fix:** Add structured `negative_evidence` to `get_ranked_context` and `search_symbols`. Add top-level `warning` strings that are hard to miss. Add `action: "STOP_AND_REPORT_GAP"` to `plan_turn` for low/none confidence. Record dead-end searches in session journal for cross-tool escalation.

| Aspect | Before | After |
|--------|--------|-------|
| Empty `get_ranked_context` result | `{"context_items": [], ...}` (no signal) | `{"negative_evidence": {...}, "warning": "No implementation found..."}` |
| Empty `search_symbols` result | `{"results": [], ...}` (no signal) | Same structured negative_evidence + warning |
| `plan_turn` on missing feature | `confidence: "low"` (advisory text) | `confidence: "low", action: "STOP_AND_REPORT_GAP"` (machine-readable directive) |
| Dead-end search visibility to plan_turn | Invisible | Recorded in session journal, escalates to `confidence: "none"` |
| `budget_warning` location | Buried in `_meta` only | Duplicated at top level |

### 4. Framework-aware route/middleware providers

**Problem:** Only Laravel, Next.js, and Nuxt had context providers. The majority of web frameworks — Express, Flask, FastAPI, Django, Spring Boot, NestJS, ASP.NET, Go routers, Rails — had no route or middleware detection. This meant `search_symbols` couldn't find endpoints, dead-code analysis missed route handlers, and the dependency graph lacked route-to-handler edges.

**Fix:** Added 6 new context provider files covering 18 frameworks, grouped by detection mechanism. Each provider detects the framework from manifest files, scans source for route patterns, and enriches symbols with searchable route metadata via the existing `FileContext` / `enrich_symbols` pipeline.

| Provider file | Frameworks | Detection mechanism |
|--------------|------------|-------------------|
| `decorator_routes.py` | Flask, FastAPI, Spring Boot, NestJS, ASP.NET | Decorator/annotation regex on source |
| `express.py` | Express, Fastify, Hono, Koa | Call-expression regex (`app.get()`, `router.post()`) |
| `go_routers.py` | Gin, Chi, Echo, Fiber | Call-expression regex (`r.GET()`, `r.Use()`) |
| `django.py` | Django (+ DRF) | `urls.py` DSL parsing + `@api_view` decorators |
| `rails.py` | Rails | `routes.rb` DSL parsing + resource expansion |
| `_route_utils.py` | (shared) | Dependency detection, tag builders, consolidated entry-point regex |

Also added 8 new `FrameworkProfile` definitions with layer/entry-point patterns, and consolidated the duplicated `_ENTRY_POINT_DECORATOR_RE` from `find_dead_code.py` / `get_dead_code_v2.py` into a single source of truth.

| Aspect | Before | After |
|--------|--------|-------|
| Frameworks with route detection | 3 (Laravel, Next.js, Nuxt) | 21 (+ 18 new) |
| Express.js route discovery | None | `app.get()`, `router.use()`, mount detection |
| Flask/FastAPI route discovery | None | `@app.route()`, `@router.get()` |
| Django URL detection | None | `path()`, `include()`, DRF `@api_view` |
| Spring Boot endpoint detection | None | `@GetMapping`, `@PostMapping`, `@RestController` |
| Go router detection | None | `r.GET()`, `r.Use()`, `r.Group()` |
| Rails route detection | None | `resources`, verb routes, namespace nesting |
| Framework profiles defined | 5 | 13 |
| Entry-point decorator regex | Duplicated in 2 files | Single source in `_route_utils.py` |

### 5. Performance optimizations

**Problem:** Incremental feature development accumulated overhead: double AST walk during indexing, eager reverse-index computation on every load, O(N x M) file I/O in dead-code analysis, redundant provider calls per symbol, and unnecessary scans.

| Optimization | Before | After | Estimated gain |
|-------------|--------|-------|---------------|
| AST walk during indexing | 2 full passes (symbols + calls) | Single pass (calls collected during `_walk_tree`) | ~15-30% faster per-file parse |
| `_callers_by_name` index | Built eagerly on every index load | Lazy — built on first call graph query | 0ms load time when call graph unused |
| `get_dead_code_v2` Signal 2 | O(N x M) file I/O per symbol | AST call_references lookup (O(1)) + cached file fallback | ~10-100x faster on v8 indexes |
| `enrich_symbols` | O(symbols x providers) `get_file_context` calls | Pre-computed per unique file, then O(1) lookup | ~60-80% fewer provider calls |
| Disambiguation + complexity | 2 separate O(N) passes | Single merged pass with fast set-based duplicate check | ~33% fewer iterations |
| PHP detection in `__post_init__` | `any(f.endswith(".php") for f in source_files)` — O(N) | `"php" in self.languages` — O(1) | Eliminates O(N) scan on every load |
| Fuzzy search | Full O(N) scan over all symbols | Early exit after 5x max_results candidates | Bounds worst case on large repos |
| Semantic search negative evidence | Fragile nested ternary `(max_bm25 < t if not semantic_only else max_cos < t)` | Named `best_score` variable | Refactoring safety, no perf change |

### Bug fixes

- `get_blast_radius` docstring was missing the `decorator_filter` parameter documentation
- Semantic search path in `search_symbols` used a fragile ternary for threshold comparison that could break during refactoring — replaced with a named variable
- `_migrate_v7_to_v8` sets `call_refs_missing` metadata flag and logs a warning directing users to re-index (not a silent no-op)

---

## Test plan

- [x] `python -m pytest tests/ -x` — 2,310 passed, 5 skipped, 0 failed
- [x] All 11 new test files pass independently
- [x] Existing test_call_hierarchy.py passes with updated methodology assertions
- [x] Backward compatibility: v7 indexes load and degrade to text heuristic
- [x] No regressions in existing search, retrieval, or indexing tests

### New test files

| Test file | Coverage |
|-----------|---------|
| `test_call_extraction.py` | Per-language AST call extraction (Python, JS, TS, Go, Rust, Java, Ruby, C#) |
| `test_call_graph_ast.py` | E2E call hierarchy with AST methodology, JS variant, v7 fallback |
| `test_call_references_model.py` | Symbol field, CodeIndex lazy index, SQLite round-trip, v4/v7 compat |
| `test_decorator_awareness.py` | Outline surfacing, search filtering, multi-decorator, case-insensitive |
| `test_negative_evidence.py` | Empty results, low confidence, warning strings, related_existing |
| `test_plan_turn.py` | Action field for low/none confidence, absent for high |
| `test_decorator_routes.py` | Flask, FastAPI, Spring Boot, NestJS, ASP.NET route detection |
| `test_express_provider.py` | Express route/middleware/mount detection |
| `test_go_routers_provider.py` | Gin, Chi, Echo, Fiber route/group/middleware detection |
| `test_django_provider.py` | Django URL patterns, DRF decorators |
| `test_rails_provider.py` | Rails resources, verb routes, namespace nesting |
| `test_route_utils.py` | Shared utility functions, dependency detection, tag builders |


jgravelle (changes requested):
Thanks for this massive contribution — the PR description is one of the best I've seen, and the architecture (single-pass AST call collection, lazy reverse index, structured negative evidence) is well thought out. CI is green, test coverage is strong (11 new test files). A few issues to address before merging:

## Must fix

### 1. Bug: `_callers_from_references` lookup key is inverted (`_call_graph.py`)

The `_callers_by_name` index is keyed by `(caller_file, called_name)` — i.e., "file X contains a symbol that calls `foo`". But `_callers_from_references` looks up `(sym_file, sym_name)` — that's the **callee's** file, not the caller's. This will only find self-calls within the same file, not cross-file callers.

The fix should iterate over `importing_files` and check `(importer_file, sym_name)` for each:

```python
for importer_file in importing_files:
    for cid in callers_by_name.get((importer_file, sym_name), []):
        ...
```

### 2. `_CALL_NODE_TYPES` has duplicate dict keys (`extractor.py`)

The dict literal defines `"javascript"`, `"typescript"`, `"tsx"`, and `"java"` twice. Python silently keeps only the last value, so the behavior happens to be correct (the later entries are supersets), but the earlier entries are dead code. Please remove the duplicates to avoid confusion — a reader would reasonably assume all entries matter.

## Should fix

### 3. `get_blast_radius` decorator filter is O(N²) (`get_blast_radius.py`)

The post-filter does `[s for s in index.symbols if s.get("file") == imp_file]` for every confirmed file — that's O(confirmed × total_symbols). Pre-build a `file → symbols` map once:

```python
from ._call_graph import build_symbols_by_file
syms_by_file = build_symbols_by_file(index)
```

### 4. Non-English comment in `_call_graph.py`

Line ~209: `# Always use text heuristic as fallback/补充` — please use English.

### 5. `has_dependency` redundant branch (`_route_utils.py` ~line 277)

```python
if manifest == "package.json":
    packages = parser(content)
else:
    packages = parser(content)
```

Both branches are identical — the conditional is dead code from a refactor. Remove it.

### 6. Missing trailing newlines

`plan_turn.py` and `test_plan_turn.py` both end without a trailing newline.

---

Everything else looks great. The backward-compat migration path (v7→v8 with graceful degradation), the `enrich_symbols` caching, the fuzzy search early-exit cap, and the framework provider architecture are all solid. Happy to approve once the lookup key bug (#1) and the duplicate keys (#2) are fixed.

jgravelle (approved):
All 6 issues addressed, plus the contributor found and fixed two additional bugs I missed:

- `find_direct_callers` wasn't returning `ast_callers` (only tracked in `seen_ids` but never yielded)
- Same-file callers were excluded because `sym_file` wasn't in `importing_files`

Both are now fixed with tests. CI green. Ship it.

PR #233, opened 2026-04-11
Auto-Watch on Demand — Reindex & Watch Unwatched Repos Before Tool Dispatch

MariusAdrian88:
## Summary

When an LLM calls any jcodemunch MCP tool against a repo that isn't being file-watched, the server now **automatically reindexes and starts watching** that folder **before** the tool executes. This ensures the LLM always sees fresh results from the very first call — eliminating the silent-stale-data problem that causes LLMs to abandon jcodemunch tools for the rest of the session.

## Why

LLMs that use jcodemunch MCP tools can silently receive stale results when a repo isn't being file-watched. When results look wrong, the LLM falls back to `grep`/`glob`/`Read` — wasting **10-50x more tokens** for every subsequent call in the session. Once an LLM loses trust in jcode results, it doesn't come back.

The blocking reindex before tool dispatch is intentional: returning stale results on the first call is worse than a one-time indexing delay, because it triggers permanent tool abandonment.

### New: `WatcherManager` class (`watcher.py`)

A high-level manager that replaces direct task manipulation in `watch_folders()`:

- **`add_folder(path)`** — acquire file lock, start watch task, mark as watched
- **`remove_folder(path)`** — cancel task, release lock, clean up
- **`is_watched(folder)`** — O(1) set lookup with path normalization
- **`list_folders()`** — sorted list of watched folders
- **`ensure_indexed(folder)`** — race-safe reindex using `asyncio.Condition`; concurrent callers wait and receive the actual result (only one reindex runs per folder)
- **`run()`** — main loop monitoring crashed tasks and waiting on stop event

### New: Pre-tool-dispatch hook (`server.py`)

`_auto_watch_if_needed()` fires **before** every tool call:

1. No watcher running or `watch: false` config → skip (zero cost)
2. Folder already watched → skip (one O(1) set lookup)
3. Folder not watched → blocking `ensure_indexed` → `add_folder` → tool executes against fresh index

Meta-only tools (`list_repos`, `get_session_stats`, `get_session_context`, `get_session_snapshot`, `index_file`) are excluded.

### New: `get_source_root()` (`sqlite_store.py`)

Lightweight metadata-only SQLite query to resolve a repo ID to its folder path without loading the full `CodeIndex`.

### New: `watch` config key (`config.py`)

Opt-in via `watch: true` in config or `JCODEMUNCH_WATCH=1` env var. Default: `false`.

## Before vs After

| Scenario | Before | After |
|----------|--------|-------|
| LLM calls `search_symbols` on unwatched repo | Returns stale/empty results; LLM falls back to grep for the rest of the session | Auto-reindex fires, tool returns fresh results, LLM stays on jcodemunch tools |
| Two concurrent tool calls hit same unwatched repo | Both trigger independent reindexes (wasted work) | One reindex runs; second caller waits and gets the same result |
| Watcher running, repo already watched | Tool executes normally | Tool executes normally (one O(1) set check, no overhead) |
| No watcher running at all | Tool executes normally | Tool executes normally (`_watcher_manager is None` → immediate return) |
| `watch: false` in config | N/A | Auto-watch is a no-op even if watcher is running |
| `watch_folders()` shutdown | Individual tasks not explicitly cancelled | All `_watch_single` tasks cancelled and awaited cleanly |
| `_watch_single` idle timeout tracking | `on_reindex` callback updated last-reindex time | Same — `on_reindex` callback properly passed through WatcherManager |
| Token cost per session (stale repo) | 10-50x higher (grep/glob/Read fallback) | Baseline (jcodemunch tools used throughout) |

## Files Changed

| File | Change |
|------|--------|
| `src/jcodemunch_mcp/watcher.py` | `WatcherManager` class with race-safe `ensure_indexed`, clean shutdown, `on_reindex` passthrough |
| `src/jcodemunch_mcp/server.py` | `_watcher_manager` global, `_auto_watch_if_needed` pre-dispatch hook, `_get_source_root` helper |
| `src/jcodemunch_mcp/storage/sqlite_store.py` | `get_source_root(owner, name)` metadata query |
| `src/jcodemunch_mcp/config.py` | `watch` config key in DEFAULTS, CONFIG_TYPES, ENV_VAR_MAPPING |
| `tests/test_watcher_dynamic.py` | 15 new tests covering manager lifecycle, race guard, and auto-watch integration |
| `tests/test_watcher_serve.py` | Updated mocks to use `WatcherManager` |
| `CLAUDE.md` | Updated Key Files |
| `README.md` | Documented auto-watch behavior |

## Test Plan

- [x] `pytest tests/test_watcher_dynamic.py -v` — 15 tests pass
- [x] `pytest` full suite — 2654 passed, 5 skipped
- [ ] Integration: start server with `--watcher`, call a tool against an unwatched repo, verify "Auto-watch: reindexing" in logs
- [ ] Race test: two concurrent `search_symbols` to same unwatched repo — only one reindex in logs
- [ ] Config off: set `watch: false`, verify no auto-watch triggers


jgravelle (approved):
Solid work, Marius. The race-safe Condition-based ensure_indexed is well designed, and the auto-watch hook is cleanly gated (zero cost when disabled).

I pushed one fix commit to the branch:

1. **on_reindex missing in run() restart** -- restarted _watch_single tasks weren't getting the on_reindex callback, so idle-timeout tracking would silently break after a crash recovery. Same bug class your second commit fixed for the initial path.

2. **_pending_results memory leak** -- changed .get() to .pop() so entries are removed after the concurrent waiter consumes them. Without this, the dict grows unbounded in long-running sessions with many auto-watched repos.

Full suite: 2654 passed, 5 skipped.

jgravelle:
Merged and shipped as v1.33.0.

- Fix commit pushed before merge: on_reindex callback in run() restart path + _pending_results .get() changed to .pop()
- Full suite: 2654 passed, 5 skipped
- Published to PyPI: https://pypi.org/project/jcodemunch-mcp/1.33.0/
- GitHub release: https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.33.0

Thanks again Marius -- great feature.

jgravelle:
Opus is acting weird this morning.  Going over everything manually.  Thanks... -jjg

PR #238, opened 2026-04-12
fix: invalidate_cache reliability + meta_fields + watcher self-restart + orphan cleanup

MariusAdrian88:
## Summary

Four reliability fixes for the MCP server:

**1. `invalidate_cache` → `index_folder` returns "No changes detected"**

Calling `invalidate_cache` followed by `index_folder` (default `incremental=true`) would silently skip the reindex and return `"No changes detected"`. This confused tool-chaining workflows that expected a full reindex after invalidation.

Root causes:
- **Windows WAL-mode file locking**: SQLite's WAL mode creates `-wal`/`-shm` sidecar files that can remain locked by in-process connections (the watcher), causing `db_path.unlink()` to fail silently on Windows. The next `index_folder` call would find the intact DB and correctly report no changes.
- **Legacy JSON auto-migration**: `delete_index` preserved `.json` when no `.db` existed (sole-copy guard), and `load_index` auto-migrated it back into a fresh `.db`, making the invalidation invisible.

Fixes:
- Added `_unlink_retry()` in `sqlite_store.py` with retry logic for `PermissionError` on Windows (3 attempts, 100 ms delay)
- Added `force=True` to `IndexStore.delete_index` — deletes legacy JSON unconditionally when called from `invalidate_cache`
- Added module-level `_force_full_reindex` set in `invalidate_cache.py` as a belt-and-suspenders flag: even if deletion silently fails (Windows WAL race), the next `index_folder` call for that repo is guaranteed to do a full reindex

**2. `meta_fields` config ignored for batch tool results**

The server-level `meta_fields` filter only stripped the top-level `result["_meta"]`. Tools returning a `results` array — specifically `get_file_outline` with `file_paths=[...]` — include a per-item `_meta` on each element that was left untouched regardless of config.

Fix: Extend both branches of the `meta_fields` filter in `server.py` to walk `result["results"]` and apply the same strip/filter logic to each item. Future batch tools following the same `{results: [...]}` shape benefit automatically.

**3. WatcherManager self-restarts on crash**

If the embedded watcher's monitoring loop crashes, it now auto-restarts with 100ms backoff, up to 5 consecutive attempts. After 5 crashes, it logs and exits cleanly.

Root cause:
- `WatcherManager.run()` was a single `while` loop — any unhandled exception killed the entire watcher silently. The server had no mechanism to detect or restart it, leaving auto-reindex permanently disabled until server restart.

Fix:
- Wrapped the monitoring loop in a `while True` with `try/except` — crashes are caught, logged with attempt count, and the loop restarts after 100ms
- Restart counter resets only on graceful shutdown (when `stop_event` is set and the inner loop exits normally) — only truly consecutive crashes count toward the limit
- Removed redundant 0.1s startup check in `_run_server_with_watcher` (manager now self-manages)
- Graceful stop still works: `stop_event.set()` → inner loop exits → outer loop detects `is_set()` and breaks

**4. Orphan index cleanup on startup**

On serve startup, scan all indexed repos and delete any whose `source_root` folder no longer exists on disk. Remote repos (GitHub) are skipped since they have no filesystem path.

Root cause:
- Indexed folders can be deleted externally (e.g. `rm -rf project/`) while their index remains in `~/.code-index/`. These orphan indexes accumulate, waste space, and cause confusing `list_repos()` output.

Fix:
- `SQLiteIndexStore.cleanup_orphan_indexes()` (delegated via `IndexStore.cleanup_orphan_indexes()`) runs before `load_all_project_configs()` at startup
- For each indexed repo with a non-empty `source_root`, check `Path(source_root).is_dir()` — if `False`, call `delete_index(owner, name)`
- Logged: `"Deleted orphan index: %s (source_root: %s)"` per deletion, `"Cleaned up %d orphan index(es)"` summary

## Files Changed

- `src/jcodemunch_mcp/storage/sqlite_store.py` — `_unlink_retry()` helper, WAL sidecar cleanup, `cleanup_orphan_indexes()`
- `src/jcodemunch_mcp/storage/index_store.py` — `force` param on `delete_index`; delegate `cleanup_orphan_indexes()` to `_sqlite`
- `src/jcodemunch_mcp/tools/invalidate_cache.py` — `_force_full_reindex` set, `force=True` call
- `src/jcodemunch_mcp/tools/index_folder.py` — check `_force_full_reindex` before incremental guard
- `src/jcodemunch_mcp/server.py` — extend `meta_fields` filter to nested `results` items; remove redundant watcher startup check; call `cleanup_orphan_indexes()` at startup
- `src/jcodemunch_mcp/watcher.py` — `WatcherManager.run()` self-restart with 5-attempt limit and 100ms backoff

## Test Plan

- [x] All 2995 tests pass, 5 skipped
- [ ] Manual: `invalidate_cache` → `index_folder` (incremental=true) performs full reindex
- [ ] Manual: `get_file_outline` with `file_paths=[...]` and `meta_fields: []` returns no `_meta` on any item
- [ ] Manual: watcher crash → auto-restarts within 100ms, survives 4 crashes, abandons on 5th
- [ ] Manual: delete indexed folder externally → restart server → orphan index is deleted at startup

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -2980,18 +2980,11 @@ async def _run_server_with_watcher(
 
     _watcher_manager = manager
 
-    # Create manager run task
+    # Create manager run task (self-restarts on crash)
     manager_task = asyncio.create_task(
         manager.run(),
         name="watcher-manager",
     )
-
-    # Give watcher a moment to start; detect early failures before blocking on server
-    await asyncio.sleep(0.1)
-    if manager_task.done() and not manager_task.cancelled():
-        exc = manager_task.exception()
-        if exc is not None:
-            logger.warning("Embedded watcher failed to start: %s", exc)
 
     try:
         await server_coro_func(*server_args)
@@ -4464,6 +4457,20 @@ def main(argv: Optional[list[str]] = None):
         # Re-run load_config() after _setup_logging() so config warnings/errors
         # go to the configured log destination (the early call at startup ran before logging was set up)
         config_module.load_config()
+
+        # Clean up orphan indexes whose source_root no longer exists
+        try:
+            from .storage import IndexStore
+
+            storage_path = os.environ.get("CODE_INDEX_PATH")
+            store = IndexStore(base_path=storage_path)
+            cleaned = store.cleanup_orphan_indexes()
+            store.close()
+            if cleaned:
+                logger.info("Cleaned up %d orphan index(es)", cleaned)
+        except Exception:
+            logger.debug("Orphan index cleanup failed", exc_info=True)
+
         config_module.load_all_project_configs()
         from .reindex_state import set_freshness_mode
         # Apply config default if --freshness-mode was not explicitly provided
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -2878,6 +2878,10 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
             meta_fields = config_module.get("meta_fields")
             if meta_fields == [] or arguments.get("suppress_meta"):
                 result.pop("_meta", None)
+                # Also strip nested _meta from batch tools (e.g. get_file_outline batch)
+                for _item in result.get("results", []):
+                    if isinstance(_item, dict):
+                        _item.pop("_meta", None)
             elif isinstance(meta_fields, list):
                 # Partial field inclusion — keep only the fields listed in meta_fields,
                 # preserving tool-generated fields (timing_ms, tokens_saved, etc.)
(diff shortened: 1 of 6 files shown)


MariusAdrian88:
Will add another fix into this PR. Stay tuned..

MariusAdrian88:
Hi @jgravelle 
This PR is ready for review/merge.

jgravelle (approved):
Marius, excellent work -- four solid reliability fixes with thorough commit messages. All 2995 tests pass. Approving with one bug to fix before merge:

BUG: cleanup_orphan_indexes() is unreachable

server.py line 4463 does store = IndexStore(...) then store.cleanup_orphan_indexes(). But cleanup_orphan_indexes() is defined on SQLiteIndexStore, not IndexStore. IndexStore has no __getattr__ delegation. This raises AttributeError, caught silently by the except Exception block -- orphan cleanup never actually runs.

Fix: add a one-line delegate to IndexStore:

    def cleanup_orphan_indexes(self) -> int:
        return self._sqlite.cleanup_orphan_indexes()

MINOR: PR description says "Restart counter resets on each successful monitoring cycle" but the code resets _restart_count only when the inner while loop exits normally (stop_event set), not per-iteration. The actual behavior is correct and reasonable -- just the description is slightly off.

Everything else looks clean. The _unlink_retry Windows WAL handling, force param on delete_index, _force_full_reindex belt-and-suspenders flag, and the meta_fields batch filtering are all well-reasoned. Push the IndexStore fix and I will merge.

MariusAdrian88:
@jgravelle 

Thank you for the review!
Done.

PR #245, opened 2026-04-18
Add `open_in_viewer` to `render_diagram` — config-gated mmd-viewer spawn

MariusAdrian88:
## Summary

Adds an optional `open_in_viewer` boolean parameter to `render_diagram`. When set to `true`, the
produced Mermaid markup is written to a temp file and opened in the local `mmd-viewer` binary for
instant visual preview — no copy-paste required. The parameter is hidden from the tool schema by
default and only surfaced when the user explicitly opts in via config, so LLM clients see no change
unless the feature is turned on.

---

## Why

`render_diagram` already produces rich, annotated Mermaid markup, but there was no way to preview
it visually without manually copying the string into an external tool. Developers running jCodemunch
locally have `mmd-viewer` available and wanted a single call to go from "I want to see this
diagram" to "diagram is open in front of me". The feature is gated so it stays invisible to
cloud/headless deployments where spawning a GUI process makes no sense.

### Getting `mmd-viewer`

`mmd-viewer` is a companion binary that renders Mermaid markup in a browser with cursor-anchored
zoom/pan. Download it from:

> **https://github.com/MariusAdrian88/mmd-viewer**

Once installed, point `mermaid_viewer_path` at the executable or drop it on `$PATH` — the
integration will pick it up automatically.

---

### New module — `tools/mermaid_viewer.py`
- `resolve_viewer_path()` — returns the configured executable path, or falls back to `mmd-viewer`
  on `$PATH`. Returns `None` when neither resolves.
- `open_diagram(mermaid, storage_path?)` — writes a `jcm-`-prefixed `.mmd` file under
  `<index_storage>/temp/mermaid/` and spawns the viewer with the file on stdin. Non-fatal on
  spawn failure — returns `{opened, path, error?}`.
- `cleanup_temp_dir(storage_path?)` — removes only `jcm-`-prefixed files, leaving anything else
  in the folder untouched. On Windows, retries once after 500 ms to handle viewer process locks.
- `was_viewer_used()` — session flag; used by the shutdown hook to skip cleanup when the viewer
  was never invoked.

### `config.py`
- Two new keys in `DEFAULTS` and `CONFIG_TYPES`:
  - `render_diagram_viewer_enabled` (bool, default `false`) — schema gate.
  - `mermaid_viewer_path` (str, default `""`) — absolute path to the executable; empty = `$PATH`.
- Full commented block added to `generate_template`.

### `tools/render_diagram.py`
- New `open_in_viewer: bool = False` parameter.
- After result assembly, if `open_in_viewer=True` and the config gate is on, calls `open_diagram`.
  Failure adds `viewer_error` to the result; success adds `viewer_path`. Mermaid is always returned.

### `server.py`
- Schema: `open_in_viewer` property injected into `render_diagram`'s `inputSchema` only when
  `render_diagram_viewer_enabled` is `true` — invisible to LLMs otherwise.
- Dispatcher: passes `open_in_viewer=arguments.get("open_in_viewer", False)`.
- Lifecycle: `_cleanup_mermaid_temp_startup()` runs at module load; `_cleanup_mermaid_temp_shutdown()`
  registered via `atexit` (skipped if viewer was never used this session).

### `tests/test_render_diagram_viewer.py`
285-line test suite covering `resolve_viewer_path`, `open_diagram`, `was_viewer_used`,
`cleanup_temp_dir`, the config gate on `render_diagram`, and the schema gate in `_build_tools_list`.

---

### Before

```python
result = render_diagram(call_hierarchy_data)
# result["mermaid"] contains the diagram string
# → developer manually copies it into mermaid.live or another tool
```

### After

```python
# config.jsonc
# "render_diagram_viewer_enabled": true,
# "mermaid_viewer_path": "C:/tools/mmd-viewer.exe",  # or leave empty for $PATH

result = render_diagram(call_hierarchy_data, open_in_viewer=True)
# → mmd-viewer opens in the browser automatically
# result["viewer_path"] = "~/.code-index/temp/mermaid/jcm-diagram-1234-98765.mmd"

# If viewer is missing or fails, mermaid is still returned:
# result["viewer_error"] = "viewer_not_found"
# result["mermaid"] = "flowchart TD\n  ..."
```

### Schema (feature off — default)

```json
{
  "name": "render_diagram",
  "inputSchema": {
    "properties": {
      "source": { ... },
      "theme": { ... },
      "max_nodes": { ... }
    }
  }
}
```

### Schema (feature on)

```json
{
  "name": "render_diagram",
  "inputSchema": {
    "properties": {
      "source": { ... },
      "theme": { ... },
      "max_nodes": { ... },
      "open_in_viewer": {
        "type": "boolean",
        "default": false,
        "description": "When true, also open the rendered mermaid in the local mmd-viewer..."
      }
    }
  }
}
```

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -316,6 +316,9 @@ DEFAULTS = {
     "agent_selector": {},
     # LSP enrichment
     "enrichment": {},
+    # Mermaid Viewer
+    "render_diagram_viewer_enabled": False,
+    "mermaid_viewer_path": "",
 }
 
 CONFIG_TYPES = {
@@ -384,6 +387,8 @@ CONFIG_TYPES = {
     "session_max_queries": int,
     "agent_selector": dict,
     "enrichment": dict,
+    "render_diagram_viewer_enabled": bool,
+    "mermaid_viewer_path": str,
 }
 
 
@@ -1456,5 +1461,24 @@ def generate_template() -> str:
   // "path_map": "",
   //   Cross-platform path remapping. Format: "orig1=new1,orig2=new2".
   //   Allows indexes built on Linux to work on Windows and vice versa.
+
+  // === Mermaid Viewer Integration ===
+  // "render_diagram_viewer_enabled": false,
+  //   When true, render_diagram exposes an extra boolean parameter
+  //   `open_in_viewer` in its tool schema. When the caller sets
+  //   open_in_viewer=true, the produced mermaid is written as a
+  //   self-contained HTML file under <index_path>/temp/mermaid/ and
+  //   opened with the viewer resolved via `mermaid_viewer_path`.
+  //   When false (default), the parameter is hidden from the schema.
+  //   The temp folder is cleaned on server startup and shutdown.
+
+  // "mermaid_viewer_path": "",
+  //   Absolute path to the mmd-viewer executable. Used only when
+  //   render_diagram_viewer_enabled is true and the caller requests
+  //   open_in_viewer=true.
+  //   - Explicit path: used as-is (e.g. "C:/tools/mmd-viewer.exe").
+  //   - Empty string: falls back to "mmd-viewer" on $PATH.
+  //   If neither resolves, render_diagram still returns the mermaid
+  //   markup and adds a non-fatal `viewer_error` field to the result.
 }}
 '''
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -280,6 +280,36 @@ def _save_session_state() -> None:
 atexit.register(_save_session_state)
 
 
+def _cleanup_mermaid_temp_startup() -> None:
+    """Clean stale mermaid viewer temp files from previous sessions."""
+    if not config_module.get("render_diagram_viewer_enabled", False):
+        return
+    try:
+        from .tools.mermaid_viewer import cleanup_temp_dir
+        cleanup_temp_dir()
+    except Exception as e:
+        logger.debug("Mermaid temp startup cleanup failed: %s", e, exc_info=True)
+
+
+def _cleanup_mermaid_temp_shutdown() -> None:
+    """Clean mermaid viewer temp files only if viewer was used this session."""
+    if not config_module.get("render_diagram_viewer_enabled", False):
+        return
+    try:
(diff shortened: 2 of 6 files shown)


MariusAdrian88:
Hi @jgravelle ,

If depending on an external tool does not feel good, feel free to fork the repo with it and use it as you see fit.

MariusAdrian88:
There are also those 2 options to replace the usage of an external binary:

- CDN-loaded (recommended, zero bloat): replace <script>{mermaid_js}</script> with <script
  src="https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.min.js"></script>. Total added code: ~80 LOC, pure stdlib. Tradeoff:
  requires internet on first open (browser caches after).

- Bundled: ship mermaid.min.js in src/jcodemunch_mcp/assets/. Works offline, but +3 MB wheel.

jgravelle (approved):
Beautifully thought-out PR, Marius. This hits every checkbox I care about for a feature like this:

- **Double gate.** Config flag hides the schema from LLMs entirely when off; the parameter itself defaults to false. Cloud/headless deployments see zero change.
- **Non-fatal everywhere.** Viewer missing, spawn failure, path unresolved — mermaid still returns. `viewer_error` surfaces the reason without crashing the call.
- **Safe cleanup.** The `jcm-` prefix discipline means we can never nuke a user's unrelated files in that temp dir. The integration test specifically verifies an `other-tool.txt` survives cleanup — chef's kiss.
- **Windows-aware.** Retry-after-500ms on the cleanup unlink path is exactly the kind of detail that bites you in production if you forget it.
- **Session flag.** `_viewer_used` skipping shutdown cleanup when the viewer was never invoked — small, smart.

Tests all green locally (18/18 new unit, integration test skips cleanly without the binary, 284 adjacent tests still pass). No regressions.

Merging. Thanks for the clean seam into the existing config system, and for the companion `mmd-viewer` tool — happy to link it from the jCodeMunch docs once it has a release tag.

One ask only for follow-up (no blocker): the hardcoded `D:\1.Development\mmd-viewer\target\debug\mmd-viewer.exe` in the integration test will skip for everyone but you. If you want it to run on other devs' machines, reading from `MMD_VIEWER_PATH` env var (with your path as a fallback) would make it portable. Totally fine as-is though.

Shipping this in the next release. And please keep telling people about jCodeMunch — word of mouth from folks who actually build with it is worth more than any amount of README copy.

PR #256, opened 2026-04-21
fix(encoder): align all MUNCH tier-1 schemas with actual tool responses + add `__stypes`

MariusAdrian88:
## Summary

Six of the tier-1 custom MUNCH encoders shipped in v1.55.0 were written against assumed response shapes that diverged from actual tool output. This caused silent data loss across multiple tools: empty table rows, wrong column names, dicts serialized as Python repr strings, and all numeric scalars decoded as strings instead of their native types.

This PR fixes all six broken encoders and adds the missing `__stypes` type-preservation infrastructure to `schema_driven.py`, bringing it to parity with what `generic.py` already implemented.

## Why

While using the mcp server I saw errors with 'get_blast_radius' usage. 
`get_blast_radius` returned MUNCH output with `enc=br1` where:
- The `symbol` field was serialized as `"{'name': 'foo', 'kind': 'method', ...}"` (Python `str(dict)`) instead of properly flattened scalars
- Table rows for `affected_symbols` and `importer_files` were declared but always empty
- The pattern reproduced across unrelated symbols in different repos

Investigation revealed the root cause was a schema mismatch between the encoder's declared keys and the actual tool response keys. Auditing all 15 tier-1 encoders found 6 with the same class of bug, plus a foundational gap where `schema_driven.encode()` never wrote `__stypes` type hints, causing all schema-driven encoders to lose numeric/bool types on round-trip.

### Infrastructure: `__stypes` in `schema_driven.py`
- `encode()`: derives type hints from non-string scalar values and writes `__stypes` into the payload (same format as `generic.py`)
- `decode()`: reads `__stypes` from the payload and merges into the coercion map (caller-supplied `scalar_types` takes precedence)
- Backward-compatible: old payloads lacking `__stypes` fall back to string decoding

### Per-Encoder Fixes

| Encoder | Fix | Pattern |
|---------|-----|---------|
| `get_blast_radius` | Moved `symbol` from `_SCALARS` to `_NESTED`, renamed tables `confirmed`/`potential`, fixed scalar names | Schema realignment |
| `find_importers` | Fixed cols to `[file, specifier, has_importers]`, scalar `file` to `file_path`, batch via JSON blob | Schema realignment |
| `find_references` | Added `_flatten()`/`_regroup()` following `search_text.py` pattern for nested `matches[]` | Flatten/regroup transform |
| `get_signal_chains` | Union column set for 3 response modes, `_meta_keys()` dispatch per mode | Multi-mode support |
| `get_tectonic_map` | Fixed plate/drifter cols, moved `isolated_files` to JSON, `_prune_optional_plate_fields()` in decode | Schema realignment + prune |
| `get_dependency_cycles` | Transform `list[list[str]]` to `{length, files}` with `\x1f` separator (path-safe) | List-to-dict transform |

### Test Coverage
- All fixtures re-anchored to actual tool response shapes (cross-referenced against `test_render_diagram.py` realistic fixtures)
- Added: batch round-trip tests, empty-matches edge case, all 3 signal-chain modes, realistic multi-plate tectonic map
- Type assertions validate `__stypes` end-to-end (float, int, bool survive round-trip)
- 48 encoding tests pass

### `get_blast_radius`

| Aspect | Before (br1) | After (br2) |
|--------|-------------|-------------|
| `symbol` field | `"{'name': 'foo', 'kind': 'method', ...}"` (Python repr) | `symbol.name=foo symbol.kind=method symbol.file=x.py ...` |
| Confirmed files table | Declared as `affected_symbols` — **always empty** (key doesn't exist in response) | `confirmed` table with `file\|references\|has_test_reach` rows |
| Potential files table | Declared as `importer_files` — **always empty** | `potential` table with `file\|reason` rows |
| Scalars | `direction`, `importer_file_count`, `affected_symbol_count` (none exist) | `importer_count`, `confirmed_count`, `potential_count`, `overall_risk_score` |

### `find_importers`

| Aspect | Before (fi1) | After (fi2) |
|--------|-------------|-------------|
| Table columns | `file, specifier, line, column` (`line`/`column` don't exist) | `file, specifier, has_importers` |
| Scalar key | `file` (wrong — tool returns `file_path`) | `file_path` |
| Batch mode | Stale `results` table with wrong cols | Preserved via JSON blob |

### `find_references`

| Aspect | Before (fr1) | After (fr2) |
|--------|-------------|-------------|
| Table structure | Flat cols `file, line, column, specifier, kind` — none match actual nested response | Flatten/regroup: `references[].matches[]` exploded to flat `(file, specifier, match_type)` rows, regrouped on decode |
| Empty matches | N/A (table was wrong anyway) | Files with `matches: []` preserved via `__empty_groups__` JSON blob |
| Batch mode | Stale `results` table | Preserved via JSON blob |

### `get_signal_chains`

| Aspect | Before (sc1) | After (sc2) |
|--------|-------------|-------------|
| Column names | `gateway_kind, leaves, symbol_path` (none exist) | `gateway_name, kind, label, reach, file_count, chain_reach, depth_from_gateway` |
| Response modes | Discovery only (missing lookup + no-gateway scalars) | All 3 modes: discovery, lookup, no-gateway |
| `_meta` keys | Single union (discovery keys leaked into lookup output) | `_meta_keys()` dispatch: each mode encodes only its own metadata |
| Scalars | Missing `symbol, symbol_id, on_no_chain, gateway_warning` | All mode-specific scalars declared |

### `get_tectonic_map`

| Aspect | Before (tm1) | After (tm2) |
|--------|-------------|-------------|
| Plate columns | `label, file_count, representative` (wrong names) | `plate_id, anchor, file_count, cohesion, majority_directory, drifter_count, nexus_alert` |
| Drifter summary | `file, score, nearest_plate` (wrong names) | `file, current_directory, belongs_with, plate_anchor` |
| `isolated_files` | Declared as dict table (it's `list[str]`) — **always empty** | Moved to `_JSON` |
| Optional fields | `drifter_count: None` / `nexus_alert: None` on plates without drifters | Pruned in decode — absent keys instead of `None` |

### `get_dependency_cycles`

| Aspect | Before (dc1) | After (dc2) |
|--------|-------------|-------------|
| Cycle data | `list[list[str]]` silently skipped by `isinstance(row, dict)` guard — **always empty** | Transformed to `{length, files}` on encode, split back to `list[str]` on decode |
| Separator | N/A | `\x1f` (ASCII unit separator) — safe for paths containing `->` |

### Scalar Type Preservation (`__stypes`)

| Aspect | Before | After |
|--------|--------|-------|
| `overall_risk_score: 0.75` | Decoded as `"0.75"` (string) | Decoded as `0.75` (float) |
| `importer_count: 5` | Decoded as `"5"` (string) | Decoded as `5` (int) |
| `has_test_reach: true` | Decoded as `"T"` or `"true"` (string) | Decoded as `True` (bool) |
| `truncated: false` | Decoded as `"F"` or `"false"` (string) | Decoded as `False` (bool) |

---

### What

Expose MUNCH encoding behavior through `config.jsonc` with user-facing mode names (`raw`, `encoded`, `adaptive`) while keeping the internal/legacy aliases (`json`, `compact`, `auto`) backward-compatible. Make encoding decisions repo-aware: project config can override the global output mode and savings threshold per-repo.

### Why

Some LLM models may not ingest well the encoded tool output so at least we have a turn-off button

### Config Surface

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `server_output` | `str` | `"adaptive"` | Output mode: `raw` (always JSON), `encoded` (always MUNCH), `adaptive` (encode when savings clear threshold) |
| `server_output_threshold` | `float` | `0.15` | Minimum savings ratio for adaptive mode to emit MUNCH (e.g. 0.15 = 15% smaller) |

Legacy env vars `JCODEMUNCH_DEFAULT_FORMAT` and `JCODEMUNCH_ENCODING_THRESHOLD` map through the config fallback pipeline with deprecation warnings.

### Alias Mapping

| User-facing | Legacy | Internal |
|-------------|--------|----------|
| `adaptive` | `auto` | `auto` |
| `encoded` | `compact` | `compact` |
| `raw` | `json` | `json` |

Normalization happens at two layers:
- **config.py** (`_normalize_server_output`): canonicalizes to user-facing names (`adaptive`/`encoded`/`raw`) at load time
- **encoding/__init__.py** (`_FORMAT_ALIASES`): maps both user-facing and legacy names to internal format names (`auto`/`compact`/`json`) at encode time

### Repo-Aware Threading

- `server.py:call_tool()` passes `repo=repo_arg` to `encode_response()`
- `encode_response()` passes `repo` to `default_format()` and `gate.passes()`
- `default_format(repo)` reads `server_output` from project config (falling back to global)
- `gate.threshold(repo)` reads `server_output_threshold` from project config (falling back to global)
- Explicit per-call `requested_format` still takes precedence over config

### Validation

- `server_output`: must be a recognized alias (rejects unknown strings like `"banana"`)
- `server_output_threshold`: must be a number in `[0.0, 1.0]` (rejects negatives)
- Both fall back to defaults on invalid input

### Files Changed

| File | Changes |
|------|---------|
| `config.py` | `_SERVER_OUTPUT_ALIASES`, `_normalize_server_output()`, DEFAULTS/CONFIG_TYPES entries, type validation, env var mapping, template commentary |
| `encoding/__init__.py` | `_FORMAT_ALIASES`, `_normalize_format()`, `default_format(repo)`, `encode_response(repo)` |
| `encoding/gate.py` | `threshold(repo)`, `passes(repo)` |
| `server.py` | Thread `repo=repo_arg` into `encode_response()` call |

### Test Coverage (207 new lines)

| Test file | Tests added |
|-----------|-------------|
| `test_dispatcher.py` | Format alias support, config-driven default, project-over-global output mode, project-over-global threshold |
| `test_config.py` | Default values/types, alias normalization (6 aliases), template documentation, legacy env var fallback, type validation rejection, string/float key round-trips |

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -59,7 +59,24 @@ ENV_VAR_MAPPING = {
     "JCODEMUNCH_PATH_MAP": "path_map",
     "JCODEMUNCH_TRUSTED_FOLDERS_ENV": "trusted_folders",
     "JCODEMUNCH_CROSS_REPO_DEFAULT": "cross_repo_default",
+    "JCODEMUNCH_DEFAULT_FORMAT": "server_output",
+    "JCODEMUNCH_ENCODING_THRESHOLD": "server_output_threshold",
 }
+
+_SERVER_OUTPUT_ALIASES = {
+    "raw": "raw",
+    "encoded": "encoded",
+    "adaptive": "adaptive",
+    # Legacy aliases kept for backward compatibility.
+    "json": "raw",
+    "compact": "encoded",
+    "auto": "adaptive",
+}
+
+
+def _normalize_server_output(value: str) -> str | None:
+    """Normalize server_output/format aliases to canonical config values."""
+    return _SERVER_OUTPUT_ALIASES.get(value.strip().lower())
 
 
 def _global_config_path() -> Path:
@@ -316,6 +333,8 @@ DEFAULTS = {
     },
     "adaptive_tiering": False,
     "compact_schemas": False,
+    "server_output": "adaptive",  # "raw", "encoded", or "adaptive"
+    "server_output_threshold": 0.15,  # Minimum savings ratio for adaptive mode
     "disabled_tools": ["test_summarizer"],
     "descriptions": {},
     "transport": "stdio",
@@ -390,6 +409,8 @@ CONFIG_TYPES = {
     "model_tier_map": dict,
     "adaptive_tiering": bool,
     "compact_schemas": bool,
+    "server_output": str,
+    "server_output_threshold": float,
     "disabled_tools": list,
     "descriptions": dict,
     "transport": str,
@@ -551,6 +572,10 @@ def _validate_type(key: str, value: Any, expected_type: type | tuple) -> bool:
         if isinstance(value, str):
             return value.lower() in {"true", "false", "auto"}
         return False
+    if key == "server_output":
+        return isinstance(value, str) and _normalize_server_output(value) is not None
+    if key == "server_output_threshold":
+        return isinstance(value, (int, float)) and 0.0 <= float(value) <= 1.0
     if isinstance(expected_type, tuple):
         return isinstance(value, expected_type)
     return isinstance(value, expected_type)
@@ -615,6 +640,10 @@ def load_config(storage_path: str | None = None) -> None:
                                     )
 
                             _GLOBAL_CONFIG[key] = list(valid_folders)
+                        elif key == "server_output" and isinstance(value, str):
+                            normalized = _normalize_server_output(value)
+                            if normalized is not None:
+                                _GLOBAL_CONFIG[key] = normalized
                         else:
                             _GLOBAL_CONFIG[key] = value
                         _explicit_keys.add(key)  # Track explicitly set keys
@@ -647,6 +676,8 @@ def _parse_env_value(value: str, expected_type: type | tuple, key: str | None = 
     # generic bool parsing would coerce "auto" to False.
     if key == "use_ai_summaries":
         return value.strip().lower()
+    if key == "server_output":
+        return _normalize_server_output(value)
     try:
         if isinstance(expected_type, tuple):
             for t in expected_type:
@@ -869,6 +900,10 @@ def load_project_config(source_root: str) -> None:
                                         )
                                     valid_folders.add(expanded_folder)
                                 merged[key] = list(valid_folders)
+                            elif key == "server_output" and isinstance(value, str):
+                                normalized = _normalize_server_output(value)
+                                if normalized is not None:
+                                    merged[key] = normalized
                             else:
                                 merged[key] = value
                         else:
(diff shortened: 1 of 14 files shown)


jgravelle:
Merged — thank you, @MariusAdrian88. Excellent debugging work here: the root-cause trace from the `{'name': ..., 'kind': ...}` repr string back to the schema/response mismatch was spot-on, and catching the missing `__stypes` parity with `generic.py` was a nice bonus find. The realistic test fixtures (especially the multi-mode signal chains and multi-plate tectonic map) will pay dividends the next time an encoder drifts.

Really appreciated the clear before/after tables in the PR description too. Thanks for shipping this.

PR #257, opened 2026-04-21
Add generic tsconfig discovery walker with extends-chain support

bertPB:
## Summary
Adds generic depth-limited tsconfig discovery with `extends`-chain following to `_load_tsconfig_aliases`, so path aliases resolve correctly in any monorepo layout. Previously only the root `tsconfig.json`/`jsconfig.json` and `.svelte-kit/tsconfig.json` were scanned.

## What Changed
- Added `_TSCONFIG_SKIP_DIRS` module-level constant (node_modules, dist, build, .git, .turbo, .vercel, framework caches)
- Added a recursive walker (depth ≤ 5) that finds all `tsconfig*.json` / `jsconfig*.json` files in the repo tree
- Walker follows each file's `extends` chain, picking up aliases from `tsconfig.base.json` / `tsconfig.paths.json` shared configs
- Handles TypeScript 5+ array-form `extends`
- `seen_cfg` set prevents double-ingestion and circular `extends` loops
- Package references like `@tsconfig/recommended` are safely skipped via `relative_to(root)` boundary check

## Why
Without this, any monorepo using path aliases defined in sub-app tsconfigs (or centralised in a root `tsconfig.base.json`) gets empty alias maps. Import graph tools (`get_blast_radius`, `find_importers`, `get_dependency_graph`) then fail to resolve aliased imports, returning incomplete or empty results.

## Test Coverage
6 new tests in `TestTsconfigWalker`:
- **extends chain** — base config in hidden dir (`.config/`), only reachable via extends
- **array extends (TS 5+)** — both entries in hidden dirs, verifies all array elements are followed
- **circular extends** — A→B→A terminates cleanly via `seen_cfg` deduplication
- **out-of-root extends** — real file outside repo root, blocked by `relative_to` guard
- **precedence** — pins first-alphabetical-wins walk order so future refactors are visible
- **Nx libs/ layout** — `libs/shared/ui/tsconfig.json` at depth 3, asserts normalised path value

## Risk Assessment

| Category | Risk | Notes |
|----------|------|-------|
| Backwards Compatibility | Low | First-definition-wins logic unchanged; existing repos gain more aliases, never lose them |

**Overall Risk**: Low

### DIFF
--- a/src/jcodemunch_mcp/parser/imports.py
+++ b/src/jcodemunch_mcp/parser/imports.py
@@ -700,6 +700,12 @@ def _clear_python_roots_cache() -> None:
 _alias_map_cache: dict[str, dict[str, list[str]]] = {}
 _ALIAS_MAP_LOCK = threading.Lock()
 
+# Directories to skip when walking for tsconfig files.
+_TSCONFIG_SKIP_DIRS = frozenset({
+    "node_modules", ".git", "dist", "build", "out", ".cache",
+    ".next", ".nuxt", ".svelte-kit",
+})
+
 
 def _norm_alias_replacement(rep: str, tsconfig_dir_rel: str = "") -> str:
     """Normalize one tsconfig paths replacement to a repo-root-relative prefix.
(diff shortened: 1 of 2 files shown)


bertPB:
Let me know if something isn't clear or if you feel this change is redundant for some reason.
I did need this change for it to work in my monorepo setup.

jgravelle:
Thanks for digging into this, @bertPB — the underlying problem is real and the direction is right. A couple of things to sort out before merge:

**1. The "before" state in the description doesn't match `main`.**

The PR opens with *"Replaces three hardcoded glob patterns (`apps/*/tsconfig.json`, `packages/*/tsconfig.json`, `src/tsconfig.json`)..."* — but those globs aren't in `imports.py` at HEAD. The current `_load_tsconfig_aliases` handles only the root `tsconfig.json` / `jsconfig.json` and `.svelte-kit/tsconfig.json`. Could you update the description to reflect that this is **adding** generic monorepo + `extends`-chain support where previously there was none? The shape of the review changes depending on whether we're replacing existing logic or adding net-new capability.

**2. This needs tests.**

Per our CLAUDE.md maintenance practice, PRs touching parser behavior need at least one test, and 70 lines of recursion + extends-chain handling is more than enough surface to warrant several. Please add coverage for:

- **Nested monorepo layout** — `apps/web/tsconfig.json` extending a root `tsconfig.base.json` that holds the actual `paths`. This is the motivating case and should be the headline test.
- **Array-form `extends`** (TS 5+) — one `extends: ["./a.json", "./b.json"]` with paths defined in both.
- **Circular `extends`** — A extends B extends A; walker should terminate, no `RecursionError`.
- **Out-of-root reference** — `extends: "@tsconfig/recommended"` (or `../../other-repo/tsconfig.json`) should be silently skipped via the `relative_to(root)` boundary.
- **Precedence** — when two sibling workspaces both define `@/*` with different replacements, pin the resolution so a future refactor of walk order doesn't silently change semantics.

**3. Minor**

- `depth ≤ 4` is a reasonable default but hardcoded. Nx repos sometimes push per-package tsconfigs to depth 5 (e.g., `apps/foo/frontend/packages/bar/tsconfig.json`). Not blocking — just worth a comment in the code noting the tradeoff.
- The skip-list is good; consider adding `.turbo` and `.vercel` alongside `.next` / `.nuxt`.

Once tests land and the description is squared away, this should be ready to go.

bertPB:
Updated the PR
Fixed the description, added 6 tests for the cases you listed, bumped depth to 5, added .turbo/.vercel to the skip list.

jgravelle:
Merged — thanks for the thorough follow-up, @bertPB. Tests are particularly well-shaped: `test_extends_chain_resolves_hidden_base` deliberately using a dotted dir so the walker *can't* find the base directly is a clean way to prove the extends-chain branch is actually exercised, and pinning first-alphabetical-wins in `test_precedence_first_alphabetical_wins` saves us from a future silent regression. Appreciated.

One leftover nit (not worth a follow-up PR on its own, we'll pick it up in passing): the comment above the walker still says *"depth ≤ 4"* while the code now enforces `> 5`. Stale doc, trivially addressed next time that file gets touched.

Thanks again.

If you have any more problems with the software, let me know.

If you have no problems whatsoever, let the planet know... 😎

PR #258, opened 2026-04-21
fix: restore exact-match identity scoring and v8 SQLite metadata

MariusAdrian88:
## Summary

- **Identity score channel was dead** for all snake_case and camelCase queries on the main BM25 path — exact name matches got zero identity bonus (should be 50 points)
- **v8 SQLite rows lost metadata** (language, decorators, keywords) on reload — symbols silently dropped by kind/language filters
- Both bugs combined produced false negatives in `search_symbols`, `plan_turn`, and `get_ranked_context` across all indexed repos

## Why

I saw the search_symbols fail and after investigation, there were three failures on an indexed Python repo:

1. `search_symbols(query="_build_left_pane_cache", kind="method", language="python")` returned **0 results** — the method exists in source
2. `search_symbols(query="_render_row", kind="method", language="python")` returned **0 results** — same
3. `search_symbols(query="build_ui")` returned `set_ui` instead of `build_ui` — **wrong match**
4. `search_symbols(query="filter_keys_by_page")` worked correctly — **control case**

Investigation revealed two independent bugs that conspired:

**Bug A — Identity score compared tokenized query against raw name:**
`_bm25_score` tokenizes the query (`_build_left_pane_cache` → `["build", "left", "pane", "cache"]`), joins with spaces (`"build left pane cache"`), and passes that to `_identity_score`. The function then compared `"build left pane cache"` against the raw symbol name `"_build_left_pane_cache"` — guaranteed mismatch. Every snake_case and camelCase exact match scored 0 on the identity channel.

**Bug B — v8 `_row_to_symbol_dict` read metadata from an empty dict:**
When the `data` column is a JSON array (v8 call_references format), the old code did `data = {}` then `data.get("language", "")` — always returning empty strings. After SQLite reload, symbols had `language=""`, so a `language="python"` filter silently excluded them.

### `_identity_score` — raw query threading (search_symbols.py:269-302)

Added `raw_query` parameter. Checks raw query first (preserves underscores/case), falls back to tokenized `query_joined` for backward compat. Both prefix and ID-segment checks now test both forms.

### `_bm25_score` / `_bm25_breakdown` — parameter threading (search_symbols.py:305, 336)

Added `raw_query` keyword, forwarded to `_identity_score`.

### All call sites updated

| File | Function | Line |
|------|----------|------|
| `search_symbols.py` | `search_symbols` (main BM25 path) | ~728 |
| `search_symbols.py` | `_search_symbols_semantic` (hybrid path) | ~1073 |
| `search_symbols.py` | debug breakdown in main loop | ~769 |
| `get_ranked_context.py` | `get_ranked_context` | ~200 |
| `plan_turn.py` | `plan_turn` | ~98 |
| `signal_fusion.py` | `build_identity_channel` (consistency) | ~189 |

### `_row_to_symbol_dict` v8 metadata fix (sqlite_store.py:1655-1680)

When `data` is a JSON array, read `language`, `qualified_name`, `decorators`, `keywords`, `content_hash`, `ecosystem_context` from row columns instead of the empty-dict fallback. Added `logger.warning` for corrupted JSON in the v8 path (matching existing v5+ behavior).

### Dead code removal (get_ranked_context.py:296-301)

Removed `semantic_only` strategy guard around negative-evidence threshold. `get_ranked_context` only accepts `"combined"`, `"bm25"`, `"centrality"` — the guard was unreachable.

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| `search_symbols(query="_build_left_pane_cache", kind="method", language="python")` | 0 results (v8 metadata: `language=""` → filtered out) | Correct result, identity score = 50.0 |
| `search_symbols(query="_render_row", kind="method", language="python")` | 0 results (same cause) | Correct result, identity score = 50.0 |
| `search_symbols(query="build_ui")` | Returns `set_ui` (identity bonus = 0 for both; BM25 noise decides) | Returns `build_ui` first (identity = 50.0 vs 0.0 for `set_ui`) |
| `search_symbols(query="ProcessData")` | Identity score = 0 (`"process data" != "processdata"`) | Identity score = 50.0 (raw query `"ProcessData"` matches) |
| `plan_turn(query="renderRow")` | `renderRow` may not rank first (no identity bonus) | `renderRow` ranks first (identity = 50.0) |
| `get_ranked_context(query="build_ui", strategy="bm25")` | `build_ui` may not rank first | `build_ui` ranks first |
| v8 SQLite reload with decorators `["@cache"]` | `decorators = []` (read from `{}`) | `decorators = ["@cache"]` (read from row column) |
| v8 SQLite reload with `language="python"` | `language = ""` (read from `{}`) | `language = "python"` (read from row column) |
| Corrupted decorators/keywords JSON in v8 row | Silent empty fallback | `logger.warning` + empty fallback |

## Test Coverage

| Test | File | What it verifies |
|------|------|------------------|
| `test_raw_query_exact_snake_case_match` | test_bm25_correctness.py | `_identity_score` with `raw_query` on snake_case |
| `test_raw_query_exact_camel_case_match` | test_bm25_correctness.py | `_identity_score` with `raw_query` on camelCase |
| `test_breakdown_uses_raw_query_for_snake_case_exact_match` | test_bm25_correctness.py | `_bm25_breakdown` with `raw_query` shows identity=50.0 |
| `test_bm25_score_exact_snake_case_match` | test_bm25_correctness.py | BM25 integration: `_bm25_score` with `raw_query` scores >= 50 |
| `test_bm25_score_exact_camel_case_match` | test_bm25_correctness.py | BM25 integration: camelCase via `raw_query` |
| `test_exact_name_beats_partial_in_bm25` | test_bm25_correctness.py | `build_ui` outranks `set_ui` with identity bonus |
| `test_search_symbols_exact_snake_case_survives_sqlite_reload_with_language_filter` | test_search_symbols_defaults.py | End-to-end: snake_case + SQLite reload + language filter |
| `test_exact_snake_case_query_ranks_matching_symbol_first` | test_token_budget_context.py | `get_ranked_context` ranks `build_ui` first |
| `test_exact_camel_case_query_ranks_matching_symbol_first` | test_plan_turn.py | `plan_turn` ranks `renderRow` first |
| `test_v8_row_preserves_metadata` | test_call_references_model.py | v8 row columns read correctly for all metadata fields |
| `test_v8_row_qualified_name_from_row_not_name` | test_call_references_model.py | qualified_name comes from row, not fallback to name |
| `test_v8_row_logs_invalid_decorators_and_keywords_json` | test_call_references_model.py | Corrupted JSON produces warning + empty fallback |

---
Co-Authored-By: GPT 5.4 XHIGH <noreply@openai.com>

### DIFF
--- a/src/jcodemunch_mcp/retrieval/signal_fusion.py
+++ b/src/jcodemunch_mcp/retrieval/signal_fusion.py
@@ -186,7 +186,7 @@ def build_identity_channel(
     query_lower = query.lower()
     scored: list[tuple[float, str]] = []
     for sym in symbols:
-        s = _identity_score(sym, query_lower)
+        s = _identity_score(sym, query_lower, raw_query=query_lower)
         if s > 0:
             scored.append((s, sym["id"]))
 
--- a/src/jcodemunch_mcp/storage/sqlite_store.py
+++ b/src/jcodemunch_mcp/storage/sqlite_store.py
@@ -1662,11 +1662,13 @@ class SQLiteIndexStore:
                 try:
                     decorators = json.loads(deco_raw) if deco_raw and deco_raw != "[]" else []
                 except (json.JSONDecodeError, ValueError):
+                    logger.warning("Corrupted decorators JSON for symbol %s", row["name"])
                     decorators = []
                 kw_raw = row["keywords"]
                 try:
                     keywords = json.loads(kw_raw) if kw_raw and kw_raw != "[]" else []
                 except (json.JSONDecodeError, ValueError):
+                    logger.warning("Corrupted keywords JSON for symbol %s", row["name"])
                     keywords = []
                 content_hash = row["content_hash"] or ""
                 ecosystem_context = row["ecosystem_context"] or ""
--- a/src/jcodemunch_mcp/storage/sqlite_store.py
+++ b/src/jcodemunch_mcp/storage/sqlite_store.py
@@ -1646,7 +1646,6 @@ class SQLiteIndexStore:
     def _row_to_symbol_dict(self, row: sqlite3.Row) -> dict:
         """Convert a database row to a symbol dict (matches CodeIndex.symbols format)."""
         call_references: list[str] = []
-        # v5: read directly from columns. Fallback to data JSON for mid-migration rows.
         if row["data"]:
             try:
                 data = json.loads(row["data"])
@@ -1655,15 +1654,30 @@ class SQLiteIndexStore:
                 data = {}
             if isinstance(data, list):
                 # v8: data column contains call_references as JSON array
+                # Read metadata from row columns (not from data, which is an array)
                 call_references = data
-                data = {}
-            # else: legacy v4 row (data is a JSON object)
-            qualified_name = data.get("qualified_name", row["name"])
-            language = data.get("language", "")
-            decorators = data.get("decorators", [])
-            keywords = data.get("keywords", [])
-            content_hash = data.get("content_hash", "")
-            ecosystem_context = data.get("ecosystem_context", "")
+                qualified_name = row["qualified_name"] or row["name"]
+                language = row["language"] or ""
+                deco_raw = row["decorators"]
+                try:
+                    decorators = json.loads(deco_raw) if deco_raw and deco_raw != "[]" else []
+                except (json.JSONDecodeError, ValueError):
+                    decorators = []
+                kw_raw = row["keywords"]
+                try:
+                    keywords = json.loads(kw_raw) if kw_raw and kw_raw != "[]" else []
+                except (json.JSONDecodeError, ValueError):
+                    keywords = []
+                content_hash = row["content_hash"] or ""
+                ecosystem_context = row["ecosystem_context"] or ""
+            else:
+                # legacy v4 row (data is a JSON object)
+                qualified_name = data.get("qualified_name", row["name"])
+                language = data.get("language", "")
+                decorators = data.get("decorators", [])
+                keywords = data.get("keywords", [])
+                content_hash = data.get("content_hash", "")
+                ecosystem_context = data.get("ecosystem_context", "")
         else:
             # v5/v6/v7 row — direct column reads, no JSON parsing
             qualified_name = row["qualified_name"] or row["name"]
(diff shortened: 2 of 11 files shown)


MariusAdrian88:
@jgravelle 

BTW, there is a version 1.72.0 in CHANGELOG.md but I see no release with that number, the latest is v 1.71.0

jgravelle (approved):
Excellent work, Marius. This is a tight, well-diagnosed PR — both bugs are real and your root-cause analysis nails them.

**Bug A** is especially satisfying to see caught: `_tokenize` strips underscores and splits camelCase, so `\" \".join(query_terms)` for `_build_left_pane_cache` becomes `\"build left pane cache\"` and the identity check at line 288 can never match the raw name. Threading `raw_query` through is the right shape, and I like that you updated all five BM25 call sites plus `build_identity_channel` for consistency rather than patching just the loudest failure.

**Bug B** is a nasty silent-data-loss bug that would have kept biting users on every v8 reload. The `data = {}` fallthrough on array-shaped rows guaranteed empty language/decorators/keywords, so any `language=\"python\"` filter quietly dropped matching symbols — exactly the kind of thing a user would blame on \"search is broken\" without ever suspecting the loader. The fix pulls from row columns in the array branch and preserves the v4 JSON-object path in a separate else — clean.

Test coverage is strong:
- `test_raw_query_exact_snake_case_match` / `camel_case_match` lock down the unit fix
- `test_bm25_score_exact_*` verify the BM25 integration actually threads raw_query
- `test_exact_name_beats_partial_in_bm25` is the one I'd have asked you to add if it weren't already there
- `test_v8_row_preserves_metadata` + `test_v8_row_qualified_name_from_row_not_name` cover the SQLite path end-to-end
- `test_search_symbols_exact_snake_case_survives_sqlite_reload_with_language_filter` ties it all together through the real search pipeline

Minor notes (not blocking):

1. **Corrupt-JSON edge in the v8 path**: if `row[\"data\"]` fails JSON decode entirely, the code sets `data = {}`, then `isinstance(data, list)` is False, so it falls into the v4 else branch and reads metadata from that empty dict instead of the row columns. For a truly corrupt v8 row, we'd still lose language/decorators despite the row columns being intact. Rare enough that it's not worth holding the PR — worth a follow-up issue if you want to chase it.

2. **`semantic_only` dead-code removal** is a legitimate cleanup given `get_ranked_context` only accepts `combined`/`bm25`/`centrality` today. Good that you called it out in the PR body — it's a behavior change in principle, just unreachable in practice.

3. The `INDEX_VERSION 8 → 9` docstring tweak in `test_call_references_model.py` is slightly out of scope but accurate (index is at 9), so it's fine.

Approving. Thanks for the thorough write-up — the before/after table made review almost mechanical.

— J.

PR #259, opened 2026-04-22
fix: preserve v8 symbol metadata when data JSON is corrupt

MariusAdrian88:
## Summary
- Preserve row-backed v8 symbol metadata when `symbols.data` contains corrupt JSON.
- Add a regression test covering malformed v8 rows and the expected fallback behavior.

## Why
Malformed v8 `data` payloads were falling back to `{}`, which made the loader treat the row like a legacy dict row and discard metadata that was still present in dedicated columns. That turned a recoverable corruption case into silent data loss for fields the row still had intact.

## What We Did
- Changed the corrupt-JSON fallback in `SQLiteIndexStore._row_to_symbol_dict` from `{}` to `[]` so the loader stays on the v8 array code path.
- Added a regression test that inserts a malformed v8 row and asserts metadata preservation, empty `call_references`, and warning emission.

## Before / After
| Scenario | Before | After |
| --- | --- | --- |
| Corrupt v8 `symbols.data` JSON | Loader fell back to `{}` and was treated like a legacy dict row | Loader falls back to `[]` and is still treated like a v8 row |
| Row-backed metadata (`qualified_name`, `language`, `decorators`, `keywords`, `content_hash`, `ecosystem_context`) | Could be dropped during recovery | Preserved from dedicated columns |
| `call_references` recovery | Recovery followed the wrong row shape | Explicitly defaults to `[]` while keeping the remaining metadata |
| Warning behavior | Warning was emitted, but recovery could still lose metadata | Warning is still emitted and recovery keeps intact metadata |

## Test Plan
- `& ".venv\\Scripts\\python.exe" -m pytest tests/ -v --tb=short --cov=src --cov-report=term-missing`

### DIFF
--- a/src/jcodemunch_mcp/storage/sqlite_store.py
+++ b/src/jcodemunch_mcp/storage/sqlite_store.py
@@ -1651,6 +1651,7 @@ class SQLiteIndexStore:
                 data = json.loads(row["data"])
             except (json.JSONDecodeError, ValueError):
                 logger.warning("Corrupted JSON in symbol data column for row %s, skipping legacy fields", row["name"])
+                # Keep corrupt v8 rows on the array path so row-backed metadata still loads.
                 data = []
             if isinstance(data, list):
                 # v8: data column contains call_references as JSON array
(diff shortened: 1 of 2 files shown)


jgravelle (approved):
Tight, correct follow-up to #258. The {} → [] fallback is exactly right: on v8 rows the metadata lives in dedicated columns, so keeping corrupt rows on the array path preserves recovery data instead of silently dropping it via the v4 dict read. Regression test is thorough — malformed data column, intact metadata columns, asserts preservation and warning emission. CI green across the full matrix, CLA signed. Nice catch. Merging.

Thank you for the continued sharpening of this area, Marius. Please help us spread the word about jCodeMunch — word of mouth from contributors who know the internals means more than any amount of marketing. 🙏

— J.

PR #262, opened 2026-04-25
fix(parser): extract Scala 3 abstract defs, vals, vars, and type aliases

irreversible-paths:
The Scala extractor only recognized function_definition nodes, missing:
- function_declaration (abstract defs in traits/classes)
- val_definition / var_definition (fields)
- type_definition (type aliases)

tree-sitter-scala names the identifier field 'pattern' for val/var nodes, not 'name'.

Fixes SCALA_SPEC symbol_node_types, name_fields, param_fields, return_type_fields, container_node_types, and type_patterns.

Adds tests/test_scala_parser.py covering Scala 3 significant- indentation syntax: traits, classes, objects, enums, vals, vars, type aliases, and nested methods.

Symbol count on a real Scala 3 + Spring Boot project went from 734 -> 1,475 after this fix.

### DIFF
--- a/src/jcodemunch_mcp/parser/languages.py
+++ b/src/jcodemunch_mcp/parser/languages.py
@@ -1197,26 +1197,36 @@ SCALA_SPEC = LanguageSpec(
         "object_definition": "class",
         "trait_definition": "type",
         "enum_definition": "type",
+        "type_definition": "type",
         "function_definition": "function",
+        "function_declaration": "function",
+        "val_definition": "constant",
+        "var_definition": "constant",
     },
     name_fields={
         "class_definition": "name",
         "object_definition": "name",
         "trait_definition": "name",
         "enum_definition": "name",
+        "type_definition": "name",
         "function_definition": "name",
+        "function_declaration": "name",
+        "val_definition": "pattern",
+        "var_definition": "pattern",
     },
     param_fields={
         "function_definition": "parameters",
+        "function_declaration": "parameters",
     },
     return_type_fields={
         "function_definition": "return_type",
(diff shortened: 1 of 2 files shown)


jgravelle (approved):
Tested locally on the PR branch — all 3676 tests pass (2 new Scala tests + 3674 baseline), 7 skipped, 0 failures. Diff is confined to the Scala language spec and a new test file; no shared parser changes. Verified against tree-sitter-scala node names. Thanks for the fix — the 734→1475 symbol jump on real Scala 3 + Spring Boot is a meaningful coverage improvement.

jgravelle:
Shipped in **v1.80.1** — thanks for the sharp diagnosis and the clean fix, @irreversible-paths.

- PyPI: https://pypi.org/project/jcodemunch-mcp/1.80.1/
- Release notes: https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.80.1

The 734 → 1,475 symbol jump on a real Scala 3 + Spring Boot project is exactly the kind of silent-coverage hole that's hardest to catch without someone running it on a non-trivial codebase, so this is a meaningful contribution. The fact that you found and fixed both the missing node types **and** the `pattern` vs `name` field-name mismatch in tree-sitter-scala in one pass — and shipped it with a Scala-3-significant-indentation test file — saved a maintenance round trip.

If there's a domain or language you've been working in where you'd like to see jCodeMunch do better (or a use case you're hitting friction on), please open an issue — extractor coverage in the real-world long tail is exactly where outside contributions land best.

Thanks again. 🪲➡️🦋

PR #266, opened 2026-04-28
Add text-only Markdown indexing

### LINKED ISSUE #265: Index Markdown documentation as text-only files
## Problem

jCodeMunch currently filters repository files by supported language extensions before storing raw file content. Markdown documentation files such as `README.md`, `AGENTS.md`, and `docs/*.md` are skipped as unsupported extensions, so agents cannot query indexed documentation through `search_text`, `get_file_content`, or `get_file_tree`.

## Proposed approach

Add Markdown as a text-only indexed language for `.md`, `.markdown`, and `.mdx` files:

- include Markdown files in local and GitHub discovery;
- store their raw content and language metadata;
- make them available to text-oriented retrieval tools;
- intentionally emit no heading symbols and avoid changing code graph / call graph behavior.

This keeps jCodeMunch focused on structured code retrieval while making repository documentation discoverable through the existing text retrieval surfaces.

JayceeB1:
Fixes #265.

## Concept

This PR treats Markdown documentation as repository context that should be indexed and retrievable, but not as code.

Currently `.md`/`.markdown`/`.mdx` files are filtered out during discovery because they are not registered as supported language extensions. That means docs such as `README.md`, `AGENTS.md`, `CHANGELOG.md`, and project guides are unavailable to existing indexed-file tools, even though jCodeMunch already stores raw file contents and already has `search_text` for non-symbol lookups.

The implementation adds Markdown as a text-only language:

- local and GitHub discovery now accept `.md`, `.markdown`, and `.mdx`;
- Markdown files are stored in the raw file cache with `language = markdown`;
- `search_text`, `get_file_content`, `get_file_tree`, repo outlines, and file summaries can now see documentation;
- `parse_file(..., "markdown")` intentionally returns no symbols;
- Markdown is exempt from refactoring import/definition pattern coverage because it has no code imports or symbol definitions.

This keeps the code graph clean: Markdown docs do not affect call graph, import graph, dead-code detection, impact analysis, or AST search. They simply become searchable indexed text.

## Why text-only first

A heading-as-symbol parser would be a larger behavior change: it would require deciding symbol kinds for sections, section offsets, search ranking implications, and how those pseudo-symbols should interact with code-focused tools. Text-only indexing is the smallest useful change and matches existing text-only behavior for formats that are useful to index but not useful to parse into code symbols.

## Tests

- `PYTHONPATH=src uv run --group dev python -m pytest tests/test_markdown_language.py -q`
- `PYTHONPATH=src uv run --group dev python -m pytest tests/test_markdown_language.py tests/test_tools.py::test_discover_source_files tests/test_plan_refactoring.py::TestLanguageCoverage tests/test_get_file_tree.py::test_get_file_tree_falls_back_to_extension_without_symbol_language tests/test_server.py::test_language_enum_all_languages_when_config_none -q`
- `PYTHONPATH=src uv run --group dev python -m pytest tests/test_languages.py tests/test_json.py tests/test_get_file_tree.py tests/test_tools.py tests/test_plan_refactoring.py::TestLanguageCoverage tests/test_markdown_language.py -q`
- `uv run python -m compileall -q src tests/test_markdown_language.py`
- `PYTHONPATH=src uv run --group dev python -m pytest tests -q` (`3607 passed, 14 skipped`)
- `git diff --check`
- `uvx ruff check tests/test_markdown_language.py`

Note: running ruff against broader pre-existing files still reports unrelated existing lint issues in `extractor.py`, `sqlite_store.py`, `test_plan_refactoring.py`, and `test_tools.py`; this PR does not address those existing lint findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to jcodemunch-mcp are documented here.
 
+## Unreleased
+
+### Added
+- Markdown documentation files (`.md`, `.markdown`, `.mdx`) are now indexed
+  as text-only files. They are available to `search_text`, `get_file_content`,
+  `get_file_tree`, and file summaries without emitting heading symbols or
+  affecting code graph analysis.
+
 ## [1.80.1] — 2026-04-26 — Scala 3 parser fix (PR #262)
 
 ### Fixed
--- a/LANGUAGE_SUPPORT.md
+++ b/LANGUAGE_SUPPORT.md
@@ -79,6 +79,7 @@ These languages are fully indexed and searchable via `search_text`. Symbol extra
 
 | Language | Extensions     | Notes                                                              |
 | -------- | -------------- | ------------------------------------------------------------------ |
+| Markdown | `.md`, `.markdown`, `.mdx` | Documentation is indexed for `search_text`, `get_file_content`, `get_file_tree`, and file summaries; headings are not emitted as symbols |
 | TOML     | `.toml`        | Tables indexed; key-as-symbol extractor planned                    |
 
 ---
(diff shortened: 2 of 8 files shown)


jgravelle:
Thanks for the well-scoped patch — but Markdown indexing is out of scope for jcodemunch-mcp. Prose/doc indexing lives in our companion server **jdocmunch-mcp** (https://github.com/jgravelle/jdocmunch-mcp, https://pypi.org/project/jdocmunch-mcp/), which already covers `.md`/`.mdx` with section-aware retrieval, TOC, backlinks, and term lookup — much richer than text-only file caching. The two servers are designed to compose: jcodemunch-mcp owns symbols/code, jdocmunch-mcp owns prose/docs. Adding Markdown here would duplicate that surface.

Closing — but if there's a specific doc-retrieval workflow jdocmunch doesn't cover yet, please open an issue on that repo and we'll prioritize it.

MariusAdrian88:
Dear Mr. J ( @jgravelle ),

I see value in this PR, even if configured with a default value False in the config, but allowing users to opt in.
I was thinking to come up with a PR myself on the same subject.

I find it harder to work with Claude models. No matter how detailed is the Claude.md file. No matter the memories. 
Once a Claude model tries to read a file with the jcodemunch and errors because it cannot read, it defaults to READ/GLOB/GREP. GPT models are much more well-behaved.

If I mention to use jcodemunch tools and also mention to read from a plan file (markdown usually) it usually tries (wrongly) to read with jcodemunch first leading to this wrong behavior for the rest of the session unless I stop it and correct it, meaning I have to keep my eyes on what is doing instead of coding elsewhere at the same time.

I guess I could turn the prompt into a novel, really hammering to use jdocmunch but from experience Claude (especially Opus) does what it wants not what it is instructed. And it has short memory, it does respect the instructions for a few tool calls and then again it defaults to its training,

It will really help to have a minimal capability to read the markdown files in jcodemunch especially when working both with code and markdown plans. It does not have to replicate the capabilities in the jdocmunch.

Thank you,
Marius

jgravelle:
Marius — thanks, this is a genuinely useful diagnosis. The "Claude errors on `.md`, falls back to Read/Grep, contaminates the rest of the session" pattern is the actual user pain, and you're right that no amount of CLAUDE.md hammering reliably fixes it.

That said, I still don't want to index markdown text in jcodemunch — jdocmunch already does it with section-aware retrieval, and a parallel text cache here would drift. But the fix you're really asking for isn't "index markdown," it's "stop training the model to give up on our tools." Those are different.

What I'd rather ship:

1. **Structured redirect errors.** When `get_file_content` / `get_file_outline` / `search_symbols` / etc. are called on `.md` / `.mdx` / `.markdown`, return a typed error that names the correct jdocmunch tool (`get_doc`, `get_section`, `search_sections`) instead of a generic "unsupported." That gives the model a concrete next action in the same turn, which is what keeps it from defaulting to Read.
2. **`jcodemunch_guide` cross-reference.** Extend the existing guide tool so the markdown-lives-in-jdocmunch boundary is one tool call away when the model gets confused.

This addresses the real failure mode (silent fallback) without duplicating jdocmunch's surface. Mind opening an issue titled something like *"Structured redirect errors for non-code file types"* on this repo? I'll prioritize it — easier to track than a closed-PR thread.

 -jjg

PR #268, opened 2026-05-01
fix(server): return no-op sentinel from streamable-http handler

### LINKED ISSUE #267: streamable-http: TypeError 'NoneType' object is not callable on every request
## Description

Running jcodemunch-mcp with `--transport streamable-http` produces `TypeError: 'NoneType' object is not callable` errors in the ASGI layer on POST and DELETE requests.

## Root Cause

In `run_streamable_http_server()`, `handle_mcp` is registered as a Starlette **endpoint** (`endpoint=handle_mcp`). Starlette's endpoint wrapper calls the function and then invokes the return value as a `Response`:

```python
response = await func(request)
await response(scope, receive, send)  # TypeError when response is None
```

`handle_mcp` writes directly to the ASGI `send` callable via `transport.handle_request()` and returns `None`, so Starlette tries to call `None(scope, receive, send)`.

## Traceback

```
File ".../starlette/routing.py", line 62, in app
    await response(scope, receive, send)
          ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not callable
```

## Fix

Return a no-op ASGI callable sentinel (`_AlreadySent`) from `handle_mcp` instead of `None`. Since `transport.handle_request()` already wrote the full HTTP response directly to the ASGI `send` callable, the sentinel's `__call__` is a no-op — Starlette invokes it without error and no duplicate response is sent.

## Note

The existing `test_streamable_http_sessions.py` tests did not catch this because they re-implement routing logic with mocks rather than exercising the actual Starlette app. The fix includes integration tests that hit the real routing stack via `httpx.ASGITransport` and reproduce the exact TypeError when the sentinel is removed.

DrHayt:
## Summary

- Fix `TypeError: 'NoneType' object is not callable` on streamable-http POST/DELETE requests
- `handle_mcp` returns a no-op ASGI callable sentinel instead of `None`, so Starlette's endpoint wrapper can invoke it safely after `transport.handle_request()` has already written the response
- Add integration tests that exercise the actual Starlette routing stack via `httpx.ASGITransport` — the existing mock-based tests did not catch this regression

## What changed

**`server.py`** — `run_streamable_http_server()`:
- Add `_AlreadySent` sentinel class (no-op `async __call__`)
- Return `_ALREADY_SENT` from all `handle_mcp` code paths instead of implicit `None`
- Return `StarletteResponse` directly from the timeout path (instead of manually calling it on `send`)

**`tests/test_streamable_http_integration.py`** (new):
- Builds the same Starlette app structure as `run_streamable_http_server` with a lightweight fake transport
- Tests POST, DELETE, session reuse, 405 on disallowed methods, and sequential requests
- Confirmed to reproduce the exact original TypeError when the sentinel return is reverted to `None`

## Test plan

- [x] Existing `test_streamable_http_sessions.py` passes (9/9)
- [x] New `test_streamable_http_integration.py` passes (5/5)
- [x] Reverting fix to `return None` reproduces the exact original traceback in the new tests

Fixes #267

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -4287,21 +4287,30 @@ async def run_streamable_http_server(host: str, port: int):
     _sessions: dict[str, StreamableHTTPServerTransport] = {}
     _session_tasks: dict[str, asyncio.Task] = {}  # type: ignore[type-arg]
 
-    async def handle_mcp(scope, receive, send):
-        request = Request(scope, receive, send)
+    # Sentinel response: transport.handle_request() already wrote to the ASGI
+    # send callable, so Starlette's endpoint wrapper must not send anything
+    # else.  Returning this instead of None prevents the "NoneType is not
+    # callable" TypeError.
+    class _AlreadySent:
+        async def __call__(self, scope, receive, send):
+            pass
+
+    _ALREADY_SENT = _AlreadySent()
+
+    async def handle_mcp(request: Request):
         session_id = request.headers.get(MCP_SESSION_ID_HEADER)
 
         # Route to existing session if client sent a session ID we recognise.
         if session_id and session_id in _sessions:
             transport = _sessions[session_id]
-            await transport.handle_request(scope, receive, send)
+            await transport.handle_request(request.scope, request.receive, request._send)
             # Clean up terminated sessions (e.g. after DELETE).
             if transport._terminated:
                 _sessions.pop(session_id, None)
                 task = _session_tasks.pop(session_id, None)
                 if task and not task.done():
                     task.cancel()
-            return
+            return _ALREADY_SENT
 
         # New session — generate a unique ID so the transport enforces it on
         # all subsequent requests, preventing cross-session pollution.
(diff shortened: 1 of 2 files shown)


DrHayt:
Something not right, hit submit too early.

DrHayt:
We were missing some real tests that exercise this functionality, added those and now things are much better.

jgravelle (approved):
Solid root-cause analysis and minimal fix. The sentinel pattern is the right call here — Starlette's `request_response` wrapper calls the return value, so returning `None` from a write-through endpoint is broken by design. The integration test is the real value-add: hitting the actual ASGI stack via `httpx.ASGITransport` catches what mock-based tests miss. Timeout-path cleanup (returning the response directly instead of manually invoking it) is a nice bonus. All 14 streamable-http tests pass on my end.

jgravelle (approved):
Solid root-cause analysis and minimal fix. The sentinel pattern is the right call — Starlette's request_response wrapper calls the return value, so returning None from a write-through endpoint is broken by design. The integration test is the real value-add: hitting the actual ASGI stack via httpx.ASGITransport catches what mock-based tests miss. Timeout-path cleanup (returning the response directly instead of manually invoking it) is a nice bonus. All 14 streamable-http tests pass locally.

PR #270, opened 2026-05-01
fix(hooks): WorktreeRemove with worktree_path now actually removes the worktree

### LINKED ISSUE #269: WorktreeRemove hook skips git worktree remove when worktree_path is in payload
## Bug

When Claude Code fires the `WorktreeRemove` hook, worktrees are never actually removed from disk. Two independent bugs cause this:

### Bug 1: Legacy early-return skips git commands entirely

When `worktree_path` is present in the stdin payload (which Claude Code sends for `WorktreeRemove`), the handler takes a legacy early-return that only writes the manifest entry — it never runs `git worktree remove` or `git branch -D`.

`hook_event.py` lines 93-101:

```python
if legacy_path:
    resolved = str(Path(legacy_path).resolve())
    _append_manifest(event_type, resolved, manifest_path)
    print(resolved, flush=True)
    return  # <-- skips git worktree remove
```

### Bug 2: git commands run from inside the worktree being removed

Even after fixing Bug 1, `git -C cwd worktree remove` fails silently because `cwd` from the payload is the worktree's own path (where Claude Code's session was running). Git cannot remove a worktree from inside itself.

## Expected behavior

1. The legacy shortcut should only apply to `create` (caller already owns the worktree). For `remove`, git commands must still execute.
2. For `remove`, the handler should resolve the main repo root (via `git rev-parse --git-common-dir`) and run removal from there.

DrHayt:
Fixes #269

## Summary

Two bugs prevented `WorktreeRemove` from actually removing worktrees:

- **Legacy early-return skipped git commands:** The `worktree_path` field in the payload triggered a code path that only wrote the manifest entry and returned — never running `git worktree remove`. Fixed by limiting the early-return to `create` events only.
- **git commands ran from inside the worktree being removed:** `cwd` from the payload is the worktree itself (where Claude Code's session ran). `git worktree remove` cannot remove a worktree from inside itself. Fixed by resolving the main repo root via `git rev-parse --git-common-dir` and running removal from there.

## Test plan

- [x] All 15 `TestHookEvent` tests pass (2 new tests added)
- [x] Manual: confirmed orphaned worktrees are cleaned up on session exit

### DIFF
--- a/src/jcodemunch_mcp/hook_event.py
+++ b/src/jcodemunch_mcp/hook_event.py
@@ -45,6 +45,26 @@ def _resolve_worktree_path(cwd: str, name: str) -> str:
     return str(Path(cwd) / ".claude" / "worktrees" / name)
 
 
+def _resolve_main_repo(cwd: str) -> str:
+    """Return the main repo root, even when *cwd* is inside a worktree.
+
+    Uses ``git rev-parse --git-common-dir`` to find the shared ``.git``
+    directory, then takes its parent.  Falls back to *cwd* on any error.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "-C", cwd, "rev-parse", "--path-format=absolute",
+             "--git-common-dir"],
+            capture_output=True, text=True,
+        )
+        if result.returncode == 0:
+            git_common = result.stdout.strip()
+            return str(Path(git_common).parent)
+    except Exception:
+        pass
+    return cwd
+
+
 def _append_manifest(event_type: str, resolved: str, manifest_path: Path) -> None:
     """Append a single event line to the JSONL manifest."""
     entry = {
@@ -130,8 +150,11 @@ def handle_hook_event(event_type: str, manifest_path: Path | None = None) -> Non
             sys.exit(1)
 
     elif event_type == "remove":
+        # cwd may be the worktree itself — resolve to the main repo so
+        # git worktree remove doesn't run from inside the tree it's deleting.
+        repo_root = _resolve_main_repo(cwd) if cwd else _resolve_main_repo(resolved)
         result = subprocess.run(
-            ["git", "-C", cwd, "worktree", "remove", resolved, "--force"],
+            ["git", "-C", repo_root, "worktree", "remove", resolved, "--force"],
             capture_output=True,
             text=True,
         )
@@ -142,7 +165,7 @@ def handle_hook_event(event_type: str, manifest_path: Path | None = None) -> Non
         # Clean up the branch created during worktree add.
         branch_name = f"worktree-{name}"
         subprocess.run(
-            ["git", "-C", cwd, "branch", "-D", branch_name],
+            ["git", "-C", repo_root, "branch", "-D", branch_name],
             capture_output=True,
             text=True,
         )
--- a/src/jcodemunch_mcp/hook_event.py
+++ b/src/jcodemunch_mcp/hook_event.py
@@ -67,8 +67,10 @@ def handle_hook_event(event_type: str, manifest_path: Path | None = None) -> Non
     For 'remove': determines path, runs ``git worktree remove``, records
     to manifest.
 
-    When ``worktreePath`` is provided directly (legacy), the caller already
-    owns the worktree — only manifest recording is performed, no git commands.
+    When ``worktreePath`` is provided directly for a 'create' event, the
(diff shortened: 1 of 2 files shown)


DrHayt:
I need to start creating these as drafts :). I am doing some manual testing.

DrHayt:
All good now.

jgravelle (approved):
Clean fix for both bugs. Bug 1 — gating the legacy early-return on event_type=='create' is the right call: the legacy contract was 'caller already owns the worktree, just record it,' which only ever made sense for create. Bug 2 — _resolve_main_repo via git rev-parse --git-common-dir is exactly the right primitive (handles bare repos, worktrees, submodules uniformly) and the fallback-to-cwd on failure keeps existing setups working. Name-derivation from path basename for branch cleanup when name is missing is a nice touch. New tests pin both regressions. Tests pass locally.

jgravelle (approved):
Clean fix for both bugs. Bug 1 — gating the legacy early-return on event_type=='create' is the right call: the legacy contract was 'caller already owns the worktree, just record it,' which only ever made sense for create. Bug 2 — _resolve_main_repo via git rev-parse --git-common-dir is the right primitive (handles bare repos, worktrees, submodules uniformly); fallback-to-cwd on failure keeps existing setups working. Name-derivation from path basename for branch cleanup when name is missing is a nice touch. New tests pin both regressions. 35/35 hook tests pass locally.

Doelfke:
I have a question. Can this cause problems since multiple “code bases” share one index? 

If agent one changes a function, it will get reindexed with hooks.  If agent 2 is changing the same function, it will also reindex that file.  So we end up with clashing indexes. 

It probably wouldn’t be a huge deal, because the agent can recognize the tool isn’t finding the correct symbols and fall back to other tools, but that would waste turns and tokens.  

I could be misunderstand how the system works though.  If I am not, I think a fix would be to rely on the file system watcher to update the indexes, instead of the hooks, because they will only fire once the worktree is merged into the main branch.

jgravelle:
Good question, and it's exactly the kind of thing worth checking before trusting hooks in a multi-agent workflow. Short answer: by construction, separate worktrees don't share an index, so the clash scenario you sketched doesn't actually happen — but there's a real subtlety worth knowing.

**The repo-id is path-derived.** `_compute_repo_id` hashes the resolved filesystem path (`hashlib.sha1(str(resolved).encode()).hexdigest()[:8]`), so a checkout at `~/work/repo` and a worktree at `~/work/repo-feature-x` get different `local/<name>-<hash>` ids and write to different SQLite databases under `~/.code-index/`. Two agents editing the same logical file via two worktrees → two independent index updates against two separate stores. No interleaving, no race.

**Same-path concurrent writes are safe too** — the SQLite backend runs in WAL mode (`storage/sqlite_store.py:113`), so even if the file watcher and a PostToolUse hook fire simultaneously on the same worktree, they serialise at the writer side and readers aren't blocked. Tested under that exact pattern in the watcher's incremental save path.

**The subtlety you're sensing is real, but it's a UX problem not a correctness problem.** Two parallel agents on two worktrees produce two independent indexes, which means agent 2's `search_symbols` won't see agent 1's still-uncommitted edit. We shipped v1.82.0 last night specifically to soften that — when you call `resolve_repo` against a worktree that isn't indexed, the response now includes `canonical_candidates` (other indexed checkouts that share the same `--git-common-dir`), and the hint suggests using one of those for read-only lookups. So the *read* side can be unified across worktrees without forcing the *write* side to share a path. Look at `tools/resolve_repo.py:_find_canonical_candidates`.

**On your fix suggestion (file watcher instead of hooks):** the watcher already exists (`jcodemunch-mcp watch-all` subcommand) and is what you'd want for "passive" reindex on save without depending on Claude Code being the editor. But it has the same per-path semantics as the hooks — both index whatever path the change happened in, neither magically routes edits in worktree-A to the index of canonical checkout B. So swapping them wouldn't change the model you're worried about. What it WOULD change is reach: hooks only fire from Claude Code, the watcher fires from any editor, so if your second agent isn't a Claude Code session that's the right thing to use anyway.

If you have a concrete two-agents repro where the indexes actually got out of sync (not just *theoretically* could), I'd love to see it — that'd be a real bug. But the clash you described should be prevented by the path-keyed index model.

  -jjg

Doelfke:
I see, thanks for the explanation. I appreciate it!

PR #278, opened 2026-05-08
fix(get_file_outline): emit flat list with parent ids so nested symbo…

sanyapuer:
**What this PR does / why we need it:**

`get_file_outline` is the canonical "see the API surface of a file without reading it" tool. The producer in `tools/get_file_outline.py` builds a nested tree (`children` arrays under each class), but the compact `fo1` encoder schema in `encoding/schemas/get_file_outline.py` is designed for a **flat** list with `parent` ids — `parent` is one of its declared columns. The generic encoder loop in `encoding/schema_driven.py` only iterates top-level rows of each table, so every nested entry (methods, constructors, fields, properties, events) silently falls on the floor before reaching the wire. MCP clients have been seeing classes as if they had no methods since the `fo1` encoder shipped, which breaks the recommended "outline first, then drill down" workflow.

A second smaller bug rides along on the same encoder schema: `signature` is always emitted by the producer but is missing from the encoder's `cols` list, so it's silently nulled on the wire too.

**Reproduction (this is what every Claude Code / Cursor / Copilot agent sees today):**

C# — a class with a field, constructor, and method (4 symbols expected):

```csharp
public class OrderService : IOrderService
{
    private readonly IMediator _mediator;
    public OrderService(IMediator mediator) { _mediator = mediator; }
    public Task<Order> GetOrder(int id) => _mediator.Send(new GetOrderQuery(id));
}
```

`get_file_outline` over MCP returns this:

```text
#MUNCH/1 tool=get_file_outline enc=fo1
__tables="s:symbols:id|name|kind|line|end_line|parent|summary,..."
s,...::OrderService#class,OrderService,class,1,,,Class OrderService
```

→ Only the class survives. `_mediator`, the constructor, and `GetOrder` are silently dropped.

TypeScript — a class with a couple of methods plus a standalone helper:

```ts
export class HttpServer {
    constructor(private port: number) {}
    listen() { /* ... */ }
    close() { /* ... */ }
}

export function pickFreePort(start: number): number { /* ... */ }
```

`get_file_outline` over MCP returns this:

```text
#MUNCH/1 tool=get_file_outline enc=fo1
__tables="s:symbols:id|name|kind|line|end_line|parent|summary,..."
s,...::HttpServer#class,HttpServer,class,1,...
s,...::pickFreePort#function,pickFreePort,function,7,...
```

→ Top-level symbols only. `HttpServer`'s constructor, `listen`, and `close` are all missing — even though `search_symbols` finds them fine. They just never make it through `get_file_outline`'s encoder.

The Python in-process function returns the full tree correctly in both cases — it's purely the MCP-encoded transport that drops nested entries.

**Languages affected — all of them.**

The bug lives in two places, both language-agnostic:

1. `tools/get_file_outline.py:_node_to_dict` (the producer) — emits a nested tree regardless of source language; just walks the `SymbolNode` graph that `parser.build_symbol_tree` returns.
2. `encoding/schema_driven.py:encode` (the generic encoder) — iterates only top-level rows; identical code path for every tool that uses the schema-driven encoder.

So every language the indexer parses is affected — anywhere a class/module/namespace contains nested members: **Python, TypeScript, JavaScript, TSX, C#, Razor, Java, Kotlin, Go, Rust, C, C++, Objective-C, Swift, PHP, Ruby, Dart, Elixir, Erlang, Haskell, Scala, Julia, R, Perl, Lua, Luau, GDScript, AL, Blade, Vue, EJS, Verse, Fortran, Groovy, SQL, Proto, GraphQL, OpenAPI, Bash, AutoHotkey, Asm, Nix, HCL, TOML, XML, CSS**.

Verified live over MCP on a C# project and a TypeScript project during bug discovery. Python is covered by the new integration test (class + ctor + 2 methods round-trips correctly post-fix).

**Changes made:**

* `src/jcodemunch_mcp/tools/get_file_outline.py`: replaced the recursive `_node_to_dict` helper with `_flatten_tree_with_parents`, a DFS-flatten that threads `parent` ids onto each entry. The schema's `parent` column carries the hierarchy. DFS ordering preserves the "class first, then its members, then next top-level symbol" grouping that callers see today.
* `src/jcodemunch_mcp/encoding/schemas/get_file_outline.py`: added `signature` to `cols` so it stops getting silently nulled on the wire.
* Updated the `get_file_outline` docstring — it previously promised "hierarchical structure" with a `children` shape that was never delivered over MCP.

**Breaking change for direct Python callers** (documented under `## Changed` in CHANGELOG): the in-process response shape is also flat now. Any caller that walked `result["symbols"][i]["children"]` must group by `parent` — one line. The on-wire compact format already carried `parent`; this just makes the in-process shape match.

**Testing:**

* New `tests/test_get_file_outline_integration.py` — 6 tests, real-indexed Python fixture (class + ctor + 2 methods + standalone function), full producer → encoder → decoder round-trip. Asserts:
  * nested members (`__init__`, `count`) survive the wire (regression test for the dropped-children bug);
  * `parent` ids point each method at its enclosing class id;
  * DFS ordering keeps members grouped with their class, before the next top-level symbol;
  * `signature` round-trips (regression test for the silent-null bug);
  * in-process response is also flat (no `children` keys);
  * batch mode flattens too.
* Strengthened the existing tier-1 round-trip test in `tests/encoding/test_tier1_roundtrip.py`. The old version asserted only `len(out["symbols"]) == 2` against hand-rolled flat input with empty `parent=""` placeholders — which is exactly why this bug shipped: the encoder unit test never exercised the producer's actual output. Now asserts `parent` and `signature` survival on a class+methods+field example.
* Full suite: 3652 passed, 12 skipped. One pre-existing failure (`test_no_warning_when_config_key_present`) reproduces on pristine `upstream/main` — env-var leak from `JCODEMUNCH_MAX_INDEX_FILES` in the shell environment, unrelated to this change.
* Schema budget gate (`tests/test_schema_budget.py`): `git diff upstream/main..HEAD` on `src/jcodemunch_mcp/server.py` and `src/jcodemunch_mcp/config.py` is empty, so `tools/list` is byte-identical to upstream and the gate passes mathematically. The `core_compact <= 4000` sub-test passes locally.
* Retrieval-quality replay: zero diff vs `upstream/main` in `tools/search_symbols.py`, `retrieval/`, `embeddings/`, `storage/`, `benchmarks/replay/` — metrics cannot move.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to jcodemunch-mcp are documented here.
 
+## [Unreleased]
+
+### Fixed
+- **`get_file_outline` silently dropped every nested symbol** (methods,
+  constructors, fields, properties) on the MCP wire. The producer emitted a
+  nested tree (`children` arrays), but the compact `fo1` encoder schema was
+  designed for a flat list with `parent` ids and the generic encoder loop
+  iterated only top-level rows — so nested children fell on the floor. MCP
+  clients saw classes as if they had no methods, breaking the canonical
+  "outline first, then drill down" workflow. Reproduces on every language
+  (the failure lives in language-agnostic code), confirmed on C# and
+  TypeScript. Producer now DFS-flattens the tree and writes `parent` ids
+  directly, matching the encoder's existing schema intent.
+- **`signature` field was silently nulled on the wire** by `get_file_outline`.
+  The producer emitted it, but the encoder schema's `cols` list omitted it,
+  so it never made it into the compact payload. `signature` is now part of
+  the encoder schema.
+
+### Changed
+- **`get_file_outline` response shape is now flat with `parent` ids** in both
+  the in-process and over-the-wire response. The previous in-process tree
+  shape (nested `children` arrays) is removed. Direct Python callers that
+  walked `result["symbols"][i]["children"]` must rebuild the tree with a
+  `parent`-based group-by — one line. The on-wire compact format already
+  carried `parent`; this just makes the in-process shape match.
+
 ## [1.82.1] — 2026-05-08 — Handshake watchdog for stdio transport
 
 ### Added
--- a/src/jcodemunch_mcp/encoding/schemas/get_file_outline.py
+++ b/src/jcodemunch_mcp/encoding/schemas/get_file_outline.py
@@ -9,7 +9,7 @@ _TABLES = [
     sd.TableSpec(
         key="symbols",
         tag="s",
-        cols=["id", "name", "kind", "line", "end_line", "parent", "summary"],
+        cols=["id", "name", "kind", "signature", "line", "end_line", "parent", "summary"],
         intern=["id", "parent"],
         types={"line": "int", "end_line": "int"},
     ),
--- a/src/jcodemunch_mcp/tools/get_file_outline.py
+++ b/src/jcodemunch_mcp/tools/get_file_outline.py
@@ -6,7 +6,7 @@ import time
 from typing import Optional
 
 from ..storage import IndexStore, record_savings, estimate_savings, cost_avoided
-from ..parser import build_symbol_tree
+from ..parser import Symbol, build_symbol_tree
 from ._utils import resolve_repo
 
 
@@ -61,13 +61,11 @@ def _get_file_outline_single(
             },
         }
 
-    # Build symbol tree
-    from ..parser import Symbol
+    # DFS-flatten the symbol tree into a list with parent ids; class members
+    # follow their class in order.
     symbol_objects = [_dict_to_symbol(s) for s in file_symbols]
     tree = build_symbol_tree(symbol_objects)
-
-    # Convert to output format
-    symbols_output = [_node_to_dict(n) for n in tree]
+    symbols_output = _flatten_tree_with_parents(tree)
 
     elapsed = (time.perf_counter() - start) * 1000
     response_bytes = len(json.dumps(symbols_output).encode("utf-8"))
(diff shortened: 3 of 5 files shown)


jgravelle:
This is a fantastic bug report and a clean fix — thank you, @sanyapuer.

I verified the bug end-to-end before reviewing the diff:

- **Encoder loop**: `encoding/schema_driven.py:87-95` iterates only `response.get(t.key, [])` (the top-level list under `symbols`); children are never visited. So yes, every nested member fell on the floor for every language the parser handles, exactly as you wrote.
- **`signature` drop**: `cols` omitted it, so the schema-driven encoder skipped it. The producer was wasting cycles writing a field that never reached the wire.
- **In-tree fallout**: I grepped src/ and tests/ for callers that walk `result["symbols"][i]["children"]`. There are none — the only in-tree use is `server.py:3837` checking `"symbols" in result`, which doesn't care about shape. So the in-process flatten is safe.

The DFS-flatten with `parent=None` for roots, `parent=<class id>` for members, ordered so members follow their class — that matches the encoder's existing schema intent perfectly. And the new integration test that runs producer → encoder → decoder with a real-indexed Python fixture is exactly the kind of test that would have caught this when the `fo1` encoder originally shipped. Strengthening the existing tier-1 round-trip to use class+methods+field input instead of two empty-parent placeholders is the right call too — that test was passing for the wrong reason.

One ask before merge: please rebase onto current `main`. We shipped v1.82.0 (worktree-aware canonical index discovery) and v1.82.1 (handshake watchdog) since you branched, so the CHANGELOG hits a conflict. The fix-side files don't drift, just the CHANGELOG header. Once rebased I'll merge and cut a v1.83.0 — this is a semver-minor bump (in-process shape change is documented as breaking for direct Python callers).

For the release notes I'll credit you on both the fix and the test discipline — the "the unit test passed for the wrong reason" callout is the kind of thing that's worth more than the patch itself.

  -jjg

sanyapuer:
Rebase done

jgravelle:
Shipped as v1.83.0 — out on PyPI now.

- Release: https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.83.0
- PyPI: https://pypi.org/project/jcodemunch-mcp/1.83.0/

Thanks again, @sanyapuer. Two things genuinely stand out about this contribution beyond the patch itself:

1. **The bug was invisible from inside the project.** Producer was right, encoder schema was right, the failure was the *gap between them* — exactly the kind of bug that hides in a codebase for a long time because every layer's unit test passes in isolation. You caught it from outside the wire and traced it to both halves at once.
2. **Strengthening the tier-1 round-trip with a class+methods+field input** is the change that prevents this class of bug from ever shipping again. The old test passing for the wrong reason was the actual root cause behind the root cause; calling that out and fixing it is craftsmanship-level.

If you find more of these we'd love to see them — `get_file_outline` is the canonical "outline first, drill down" entry point and you've made it materially more useful for every MCP client out there. And if you wouldn't mind sharing on whichever socials you frequent (Reddit's `r/ClaudeAI` and `r/cursor`, Twitter/X, HN) that you landed a release fix in jCodeMunch — happy to amplify the post. The "compact-on-wire encoders silently drop nested symbols" finding is a good war story, and people seeing real contributors land real fixes is worth more than any marketing copy I could write.

  -jjg

PR #285, opened 2026-05-09
fix: allow OpenRouter managed endpoint and update default model

MariusAdrian88:
## Summary

- treat the built-in OpenRouter base URL as a managed OpenAI-compatible endpoint instead of blocking it as a generic remote URL
- switch the default OpenRouter summarizer model to `nvidia/nemotron-3-nano-30b-a3b:free`
- align the config output and regression tests with the new default behavior

## Why

OpenRouter provider detection already worked, but summarizer initialization still failed because the built-in OpenRouter endpoint was handled like an arbitrary remote `OPENAI_API_BASE`. That caused `test_summarizer` to report a misconfigured provider even when `OPENROUTER_API_KEY` was present.

Once that path was fixed, the previous default OpenRouter model was also a poor fit for this summary-generation workflow. The new default is faster and produced valid summaries in the local diagnostic path.

## What

- allow the built-in OpenRouter endpoint through the OpenAI-compatible safety gate without requiring `allow_remote_summarizer`
- keep the remote override requirement for custom non-managed endpoints
- change the OpenRouter default summarizer model to `nvidia/nemotron-3-nano-30b-a3b:free`
- update status output, config template text, and OpenRouter regression coverage

## Before and After

| Area | Before | After |
| --- | --- | --- |
| OpenRouter initialization | Built-in OpenRouter endpoint could be rejected as a non-local remote URL | Built-in OpenRouter endpoint is treated as a managed provider endpoint |
| Remote override requirement | OpenRouter could require `allow_remote_summarizer` even for the built-in base URL | Only custom remote endpoints require `allow_remote_summarizer` |
| Default OpenRouter model | `meta-llama/llama-3.3-70b-instruct:free` | `nvidia/nemotron-3-nano-30b-a3b:free` |
| `test_summarizer` outcome | Provider could be detected but still fail to initialize | Provider initializes and can produce a real summary |
| Diagnostics and tests | Reflected the old default model and behavior | Aligned with the managed-endpoint logic and new default model |

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -1639,7 +1639,7 @@ def generate_template() -> str:
   // Leave empty ("") to use the provider's default model.
   // Examples: "claude-haiku-4-5-20251001" (anthropic), "gemini-2.5-flash-lite" (gemini),
   //           "gpt-4o-mini" (openai), "minimax-m2.7" (minimax), "glm-5" (glm),
-  //           "meta-llama/llama-3.3-70b-instruct:free" (openrouter)
+  //           "nvidia/nemotron-3-nano-30b-a3b:free" (openrouter)
   // "summarizer_model": "",
   // "embed_model": "",
   //   Sentence-transformers model name for local (free) semantic embeddings.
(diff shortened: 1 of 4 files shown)


MariusAdrian88:
Looks like the Health / radar failure is not related to my PR:

1. The failing part isn’t radar generation. The logs show the action renders the comment successfully, then fails on gh api when trying to list or create PR comments.
2. The workflow runs on pull_request, and my branch lives in a fork. On fork PRs, the GITHUB_TOKEN is restricted, so issue-comment writes to the upstream PR return 403 Resource not accessible by integration.
3. That means the root cause is a permissions/event mismatch, not a bad radar payload or a broken gh command.

MariusAdrian88:
Looks like I misunderstood the: "allow_remote_summarizer" parameter.
Sorry for this.

PR #291, opened 2026-05-11
fix: report unloadable repo indexes truthfully

MariusAdrian88:
## Summary
- Distinguishes present-but-unloadable repository indexes from truly missing repositories.
- Keeps damaged SQLite indexes visible in `list_repos()` with machine-stable status fields and remediation hints.
- Updates query tools to return actionable loadability errors instead of generic `Repository not indexed` messages.

## Why
`resolve_repo()` could report a repository as indexed when only an index artifact existed, even if that index was not loadable by query tools. That created false positives: discovery said a repo was available, while follow-up tools failed with generic missing-index errors. Operators needed a truthful status layer that preserves visibility for broken indexes and tells users how to recover.

## What
- Added an index loadability status probe that preserves `load_index()`'s existing return contract.
- Updated `resolve_repo()` so `indexed` means queryable, while exposing `index_present`, `loadable`, `status`, `load_error`, `backend`, and `hint` for diagnostics.
- Updated `list_repos()` and SQLite listing paths so damaged indexes remain visible, including missing metadata, future versions, corrupt metadata, and corrupt database files.
- Added JSON fallback handling when a legacy JSON index is still valid but the SQLite metadata is damaged.
- Replaced generic missing-index responses across tools with shared actionable unloadable-index errors.
- Added regression coverage for loadability status, corrupt metadata, corrupt SQLite files, JSON fallback, and representative tool errors.

## Before and After

| Before | After |
|---|---|
| `resolve_repo()` could return `indexed: true` when an index file existed but could not be loaded. | `resolve_repo()` returns `indexed: false`, `index_present: true`, and loadability diagnostics for present-but-broken indexes. |
| Tools returned `Repository not indexed` for damaged indexes. | Tools return `Repository index is not loadable` with `status`, `load_error`, and `hint`. |
| `list_repos()` could hide corrupt SQLite databases. | `list_repos()` keeps corrupt or incompatible indexes visible with `loadable: false`. |
| Invalid SQLite `index_version` metadata could be treated as loadable or raise during tool execution. | Invalid `index_version` is reported as `sqlite_corrupt` and handled without crashing tools. |
| A damaged SQLite index could mask a valid legacy JSON fallback. | Valid legacy JSON fallback remains usable and can migrate back to SQLite. |

---

Co-authored-by: GPT 5.5 high

### DIFF
--- a/src/jcodemunch_mcp/storage/index_store.py
+++ b/src/jcodemunch_mcp/storage/index_store.py
@@ -47,6 +47,43 @@ logger = logging.getLogger(__name__)
 INDEX_VERSION = 16
 
 
+@dataclass(frozen=True)
+class IndexLoadStatus:
+    """Presence and queryability status for a repository index."""
+
+    repo: str
+    owner: str
+    name: str
+    backend: str
+    index_present: bool
+    loadable: bool
+    status: str
+    load_error: Optional[str] = None
+    hint: Optional[str] = None
+    indexed_at: str = ""
+    symbol_count: Optional[int] = None
+    file_count: Optional[int] = None
+    languages: Optional[dict[str, int]] = None
+    index_version: Optional[int] = None
+    git_head: str = ""
+    display_name: str = ""
+    source_root: str = ""
+
+    def as_fields(self, include_empty: bool = False) -> dict:
+        """Return status fields suitable for tool/listing payloads."""
+        fields = {
+            "index_present": self.index_present,
+            "loadable": self.loadable,
+            "status": self.status,
+            "backend": self.backend,
+        }
+        if include_empty or self.load_error:
+            fields["load_error"] = self.load_error
+        if include_empty or self.hint:
+            fields["hint"] = self.hint
+        return fields
+
+
 @functools.lru_cache(maxsize=16)
 def _load_index_json_cached(index_path: str, mtime_ns: int) -> Optional[dict]:
     """Cache parsed JSON keyed on (path, mtime). mtime_ns ensures
@@ -596,6 +633,48 @@ class IndexStore:
         """Return True if an index exists (SQLite or JSON)."""
         return self._sqlite.has_index(owner, name) or self._index_path(owner, name).exists()
 
+    def inspect_index(self, owner: str, name: str, branch: str = "") -> IndexLoadStatus:
+        """Return index presence/loadability without changing load_index's contract."""
+        sqlite_status = self._sqlite.inspect_index(owner, name, branch=branch)
+        if sqlite_status.loadable:
+            return sqlite_status
+
+        if not branch:
+            index_path = self._index_path(owner, name)
+            if index_path.exists():
+                data = _load_index_json_cached(str(index_path), index_path.stat().st_mtime_ns)
+                if data:
+                    return IndexLoadStatus(
+                        repo=data.get("repo", f"{owner}/{name}"),
+                        owner=owner,
+                        name=name,
+                        backend="json",
+                        index_present=True,
+                        loadable=True,
+                        status="loadable",
+                        indexed_at=data.get("indexed_at", ""),
+                        symbol_count=len(data.get("symbols", [])),
+                        file_count=len(data.get("source_files", [])),
+                        languages=data.get("languages", {}),
+                        index_version=data.get("index_version"),
+                        git_head=data.get("git_head", ""),
+                        display_name=data.get("display_name", ""),
+                        source_root=data.get("source_root", ""),
+                    )
+                return IndexLoadStatus(
+                    repo=f"{owner}/{name}",
+                    owner=owner,
+                    name=name,
+                    backend="json",
+                    index_present=True,
+                    loadable=False,
+                    status="json_invalid",
+                    load_error="json_invalid",
+                    hint="Re-index this repository to rebuild the unreadable JSON index.",
+                )
+
+        return sqlite_status
+
     def load_index(self, owner: str, name: str, branch: str = "") -> Optional[CodeIndex]:
         """Load index from storage. Prefers SQLite, auto-migrates from JSON.
 
@@ -759,6 +838,10 @@ class IndexStore:
             "file_count": file_count,
             "languages": data.get("languages", {}),
             "index_version": data.get("index_version", 1),
+            "index_present": True,
+            "loadable": True,
+            "status": "loadable",
+            "backend": "json",
         }
         if data.get("git_head"):
             repo_entry["git_head"] = data["git_head"]
@@ -775,6 +858,7 @@ class IndexStore:
         """List all indexed repositories (SQLite + legacy JSON)."""
         repos = []
         seen_slugs: set[str] = set()
+        json_files_to_migrate: list[Path] = []
         _pairs = parse_path_map()
 
         # Pass 1: SQLite databases
@@ -786,10 +870,42 @@ class IndexStore:
             seen_slugs.add(slug)
             try:
                 entry = self._sqlite._list_repo_from_db(db_file, _pairs)
+                json_path = self.base_path / f"{slug}.json"
+                if entry and entry.get("loadable") is False and json_path.exists():
+                    try:
+                        with open(json_path, "r", encoding="utf-8") as f:
+                            data = json.load(f)
+                        json_entry = self._repo_entry_from_data(data, _pairs)
+                        if json_entry:
+                            repos.append(json_entry)
+                            json_files_to_migrate.append(json_path)
+                            continue
+                    except Exception:
+                        pass
                 if entry:
                     repos.append(entry)
             except Exception:
                 logger.debug("Skipping corrupted DB: %s", db_file, exc_info=True)
+                parts = slug.split("-", 1)
+                owner, name = parts if len(parts) == 2 else ("local", slug)
+                status = self._sqlite.inspect_index(owner, name)
+                entry = {
+                    "repo": status.repo,
+                    **status.as_fields(include_empty=True),
+                }
+                json_path = self.base_path / f"{slug}.json"
+                if json_path.exists():
+                    try:
+                        with open(json_path, "r", encoding="utf-8") as f:
+                            data = json.load(f)
+                        json_entry = self._repo_entry_from_data(data, _pairs)
+                        if json_entry:
+                            repos.append(json_entry)
+                            json_files_to_migrate.append(json_path)
+                            continue
+                    except Exception:
+                        pass
+                repos.append(entry)
 
         # Pass 2: legacy JSON — eagerly migrate to SQLite so that every
         # repo is in the canonical format before any tool can interact with it.
@@ -800,8 +916,6 @@ class IndexStore:
         if not any(self.base_path.glob("*.json")):
             repos.sort(key=lambda r: r["repo"])
             return repos
-
-        json_files_to_migrate: list[Path] = []
 
         for meta_file in self.base_path.glob("*.meta.json"):
             slug = meta_file.name.removesuffix(".meta.json")
@@ -844,8 +958,8 @@ class IndexStore:
             if len(parts) != 2:
                 continue
             owner, name = parts
-            if self._sqlite.has_index(owner, name):
-                continue  # already migrated
+            if self._sqlite.inspect_index(owner, name).loadable:
+                continue  # already migrated and loadable
             try:
                 logger.info("Eager-migrating %s from JSON to SQLite", json_path)
                 self._sqlite.migrate_from_json(json_path, owner, name)
--- a/src/jcodemunch_mcp/storage/index_store.py
+++ b/src/jcodemunch_mcp/storage/index_store.py
@@ -854,6 +854,26 @@ class IndexStore:
                 repo_entry["source_root"] = remap(data["source_root"], _pairs if _pairs is not None else parse_path_map())
         return repo_entry
 
+    def _try_json_fallback(
+        self,
+        slug: str,
+        _pairs,
+        json_files_to_migrate: list[Path],
+    ) -> Optional[dict]:
+        """Return a legacy JSON listing entry when SQLite is unloadable."""
+        json_path = self.base_path / f"{slug}.json"
+        if not json_path.exists():
+            return None
+        try:
+            with open(json_path, "r", encoding="utf-8") as f:
+                data = json.load(f)
+            entry = self._repo_entry_from_data(data, _pairs)
+            if entry:
+                json_files_to_migrate.append(json_path)
+            return entry
+        except Exception:
+            return None
+
     def list_repos(self) -> list[dict]:
         """List all indexed repositories (SQLite + legacy JSON)."""
         repos = []
@@ -870,18 +890,10 @@ class IndexStore:
             seen_slugs.add(slug)
             try:
                 entry = self._sqlite._list_repo_from_db(db_file, _pairs)
-                json_path = self.base_path / f"{slug}.json"
-                if entry and entry.get("loadable") is False and json_path.exists():
-                    try:
-                        with open(json_path, "r", encoding="utf-8") as f:
-                            data = json.load(f)
-                        json_entry = self._repo_entry_from_data(data, _pairs)
-                        if json_entry:
-                            repos.append(json_entry)
-                            json_files_to_migrate.append(json_path)
-                            continue
-                    except Exception:
-                        pass
+                if entry and entry.get("loadable") is False:
+                    if json_entry := self._try_json_fallback(slug, _pairs, json_files_to_migrate):
+                        repos.append(json_entry)
+                        continue
                 if entry:
                     repos.append(entry)
             except Exception:
@@ -893,18 +905,9 @@ class IndexStore:
                     "repo": status.repo,
                     **status.as_fields(include_empty=True),
                 }
-                json_path = self.base_path / f"{slug}.json"
-                if json_path.exists():
-                    try:
-                        with open(json_path, "r", encoding="utf-8") as f:
-                            data = json.load(f)
-                        json_entry = self._repo_entry_from_data(data, _pairs)
-                        if json_entry:
-                            repos.append(json_entry)
-                            json_files_to_migrate.append(json_path)
-                            continue
-                    except Exception:
-                        pass
+                if json_entry := self._try_json_fallback(slug, _pairs, json_files_to_migrate):
+                    repos.append(json_entry)
+                    continue
                 repos.append(entry)
 
         # Pass 2: legacy JSON — eagerly migrate to SQLite so that every
--- a/src/jcodemunch_mcp/storage/sqlite_store.py
+++ b/src/jcodemunch_mcp/storage/sqlite_store.py
@@ -1278,7 +1278,11 @@ class SQLiteIndexStore:
             if not meta:
                 return None
 
-            stored_version = int(meta.get("index_version", "0"))
+            try:
+                stored_version = int(meta.get("index_version", "0"))
+            except (TypeError, ValueError):
+                logger.warning("Corrupt index version for %s/%s", owner, name)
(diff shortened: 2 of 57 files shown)


MariusAdrian88:
Hi @jgravelle 
Looks like the Health Radar / radar has a bug where it works for PR's created directly but not for PR's created from a fork (like mine).

MariusAdrian88:
```text
gh: Resource not accessible by integration (HTTP 403)
{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"403"}
Error: Process completed with exit code 1.

The failing step calls:
gh api "repos/${REPO}/issues/${PR_NUMBER}/comments" --method POST ...

That endpoint creates an issue comment on the PR timeline, so the workflow token needs:
permissions:
  contents: read
  pull-requests: write
  issues: write
  
Current workflow only grants:
permissions:
  pull-requests: write
  contents: read
  
So the radar calculation works, but the sticky-comment publish step fails with 403.

Suggested fix:
permissions:
  contents: read
  pull-requests: write
  issues: write
  
Optional hardening:
- Make the sticky-comment step best-effort so permission failures don't fail the whole radar job.
- Keep the fork-PR guard.
- Avoid switching to pull_request_target unless the action is carefully audited for untrusted PR code execution.

The reported regression line was generated successfully:
### Regressions
- churn_surface: raw 301.519 → 2338.2461

The actual CI failure is the missing permission for posting the sticky comment.

**Comment by jgravelle** [comment]
Thanks for catching this @MariusAdrian88 — great catch, and exactly the kind of CI gap that's only visible when you're contributing from a fork.

**Root cause:** GitHub fires `pull_request` events from forks with a read-only `GITHUB_TOKEN`, regardless of what the workflow's `permissions:` block declares. Fork code can't be trusted with write tokens, so the comment POST was returning 403 — confirmed in the run logs.

**Fix shipped** (`010fab8` on `main`) — the standard `pull_request` + `workflow_run` split:

1. `.github/actions/health-radar/action.yml` — the inline comment step is now skipped on fork PRs. Same-repo PRs are unaffected.
2. `.github/workflows/health-radar.yml` — for fork PRs, the rendered comment + PR number are staged as a `health-radar-payload` artifact.
3. `.github/workflows/health-radar-comment.yml` (new) — `workflow_run`-triggered companion that runs in the base repo's trusted-secret context, downloads the artifact, and posts/updates the sticky comment on the fork PR's behalf.

You should see the radar comment appear on this PR within a minute or two after the next radar run completes — feel free to push an empty commit (`git commit --allow-empty -m "ci: retrigger radar"`) to verify if you'd like.

On the substance of the PR itself: reviewing now, will follow up separately.

 -jjg

**Review by jgravelle — COMMENTED** [review]
Thanks for this @MariusAdrian88 — exactly the kind of correctness fix the index layer needs. The PR description's before/after table is honest, the bug is real (`resolve_repo` returning `indexed: true` for files-but-unloadable indexes was a legitimate UX trap), and the test coverage hits the actual failure modes. I ran the full suite on the PR branch locally: **4239 passed, 7 skipped, 0 failed**.

Below: what I liked, three things that need addressing before merge, and two nits.

## Strengths

- **`IndexLoadStatus` + `inspect_index()` + `as_fields()` is a clean shape.** Frozen dataclass, read-only probe, doesn't mutate `load_index()`'s contract. The status taxonomy (`loadable` / `missing` / `sqlite_missing_meta` / `sqlite_corrupt` / `sqlite_future_version` / `json_invalid`) is the right level of granularity — each maps to a different remediation path.
- **The actionable error helper (`index_status_to_tool_error`) is the right level of abstraction.** Single chokepoint for the error shape, easy to evolve in one place. The fact that ~25 tools migrate to it with a `+2/-2` diff each says the design fits.
- **JSON-fallback recovery in `list_repos`** is a nice safety net — broken SQLite + valid legacy JSON now successfully recovers and migrates rather than presenting the user with a dead repo.
- **Defensive metadata reads.** `try/except (TypeError, ValueError)` around `int(meta.get("index_version"))` and `(TypeError, json.JSONDecodeError)` around the languages JSON are the right belt-and-suspenders for a metadata layer that can legitimately drift.
- **Tests target real failure modes.** Missing meta, corrupt version int, future version, non-DB content masquerading as `.db`, JSON-fallback recovery, list_repos sort stability with damaged entries. Exactly the right matrix.
- **Semantic shift on `result[\"indexed\"]` (file-present → loadable)** propagates correctly to the 4 in-tree callers (`cli/digest.py:68`, `cli/file_risk.py:52`, `cli/health.py:55`, `tools/import_runtime_signal.py:81`) — all want the tighter \"queryable\" meaning anyway, so this is a latent-bug fix rather than a regression.

## Needs addressing before merge

### 1. Rebase needed — three tools shipped after your base point still use the old error shape

`main` has drifted forward since you branched. After rebase, the sweep needs to cover:

- `src/jcodemunch_mcp/tools/check_delete_safe.py:160` (v1.104.0)
- `src/jcodemunch_mcp/tools/find_implementations.py:170` (v1.104.0)
- `src/jcodemunch_mcp/tools/assemble_task_context.py:211` (v1.105.0)

All three still return `{\"error\": f\"Repository not indexed: {owner}/{name}\"}` and should switch to `index_status_to_tool_error(store.inspect_index(owner, name))` to match the rest of the sweep. Otherwise users hitting an unloadable index through these tools get the old generic miss while every other tool gives them the actionable diagnostic — UX inconsistency that's exactly what this PR is trying to fix.

### 2. `digest.py:135` `hasattr` guard is dead-defensive — fix the test mock instead

```python
if hasattr(store, \"inspect_index\"):
    return index_status_to_tool_error(store.inspect_index(owner, name))
return {\"error\": f\"Repository index is not loadable: {owner}/{name}\"}
```

Every `IndexStore` / `SQLiteIndexStore` instance has `inspect_index` after this PR — the `hasattr` exists only because `tests/test_digest.py:153`'s `_NullStore` mock doesn't define it. Cleaner: add a one-line `inspect_index` to the mock (returning a fake status with `loadable=False, status=\"missing\"`) and drop the guard. Otherwise we ship a branch whose only purpose is keeping a test green — invisible cruft.

### 3. JSON-fallback block in `IndexStore.list_repos()` duplicated verbatim

In `src/jcodemunch_mcp/storage/index_store.py`, the 12-line JSON-fallback block appears twice — lines 873-884 (success path, triggered by `entry.get(\"loadable\") is False`) and 896-907 (exception path, triggered by SQLite read raising). Same `open / json.load / _repo_entry_from_data / append + migrate / except: pass` sequence in both. Extract to a small helper:

```python
def _try_json_fallback(self, slug: str, _pairs, json_files_to_migrate: list[Path]) -> Optional[dict]:
    json_path = self.base_path / f\"{slug}.json\"
    if not json_path.exists():
        return None
    try:
        with open(json_path, \"r\", encoding=\"utf-8\") as f:
            data = json.load(f)
        entry = self._repo_entry_from_data(data, _pairs)
        if entry:
            json_files_to_migrate.append(json_path)
        return entry
    except Exception:
        return None
```

Halves the surface in `list_repos`, makes the two trigger paths read as `if (entry := self._try_json_fallback(...)): repos.append(entry); continue`.

## Nits (non-blocking)

- **`register_edit.py` quietly switched from `SQLiteIndexStore` to `IndexStore`.** Functionally fine — `IndexStore.load_index` tries SQLite first then falls back — but `register_edit` is called on every PostToolUse Edit/Write, so it's the hot path. Worth a sentence in the commit message acknowledging the change, in case anyone bisects a latency regression later.
- **PR description mentions \"GPT 5.5 high\" as co-author.** No issue with that — just noting we usually call out the AI model in the trailing co-author line for traceability, which you've done. Good.

## Summary

Substance: ✅ correct fix, right architecture, real test coverage. Three actual items to address (rebase + sweep three new tools + drop the `hasattr` guard) and one quality nit (dedupe the JSON-fallback block). Once those are in, this is a merge.

I shipped a separate CI fix on `main` (`010fab8`) that gets the Health Radar comment posting on fork PRs through a `workflow_run` companion — so once you rebase, you should see the radar comment appear on this PR too.

 -jjg

MariusAdrian88:
@jgravelle 

Thank you for the review. But there is actually a problem that might be deeper. 
Using the latest version of jcodemunch, I dropped the index and tried to use index_folder. It time outs.
I cannot index anymore (not using the PR above). 
I will use an older version and see if that is fixing the problem.

MariusAdrian88:
I can reindex using version 1.90.0

MariusAdrian88:
It may be a mix up of the local indexing with trying to index a GIT repo. I am not sure, but GPT says that there were some changes in v1.95 / 1.96

MariusAdrian88:
Actually it says:

"The clean fix is to add an index_folder parameter that bypasses Git-root identity for local indexing.
Recommended design:
1. Add use_git_root_identity: bool = True to index_folder.
2. When False, skip:
   - remote-derived identity like owner/repo
   - git-root retargeting
   - subdir merge behavior
3. The repo ID stays local-path based, like v1.90.0: local/<folder>-<hash>.
4. Keep the default as True so existing Git/monorepo behavior doesn't regress.
Alternative approaches:
1. Per-call parameter only: safest and most explicit.
2. Config-only setting: useful globally, but less obvious during tool calls.
3. Change default back to local-only: simpler for your case, but it would break the v1.95+ monorepo/git identity behavior.
I recommend option 1, with optional config fallback later if needed."


LE:

Current risk: watcher identity and index_folder identity are not fully aligned.
- Watcher helper _local_repo_id() still computes local/<folder>-<hash>.
- Current index_folder can resolve the same path to Git identity, like owner/repo.
- In watcher fast-path, index_folder now also calls _resolve_repo_identity(), so the save target can be Git-derived.
- That means watcher status/locks/staleness may track one repo ID while the actual index is written under another repo ID.
So if we add use_git_root_identity: false only to the manual index_folder MCP call, watcher auto-indexing may still use the Git-root behavior unless we wire it there too.
Better design:
1. Add the parameter to index_folder.
2. Add a config setting, maybe git_root_identity, already exists.
3. Make watcher call index_folder with the same effective setting.
4. Update watcher _local_repo_id() or replace it with the same identity resolver used by index_folder.
This keeps manual indexing and watcher auto-indexing from diverging.

MariusAdrian88:
More on that:

They changed index_folder to solve Git/monorepo identity problems, not because the old local indexing path was broken.
The old v1.90.0 behavior was simple:
- index_folder(path) created a local repo ID from the resolved path.
- Same repo cloned in two folders became two separate indexes.
- Indexing a subfolder created an index relative to that subfolder.
- It was predictable and fast for local-only workflows.
The newer behavior was added for these reasons:
1. Match index_folder with index_repo.
   A local clone of elastic/kibana should index as elastic/kibana, not local/kibana-somehash.
2. Support monorepo subdir indexing.
   index ./packages and index ./scripts should merge into one repo index instead of creating separate partial indexes.
3. Prevent accidental overwrite across worktrees.
   If two different local clones have the same remote, the code now tries to detect collisions instead of silently replacing the same owner/repo index.

The regression is the tradeoff: the local-only workflow got coupled to Git identity. That made index_folder do more work and made drop/reindex behavior less obvious. 

The old local-path identity should still be available as an explicit mode.

@jgravelle what do we do?
It is a matter of architecture....


LE: Suggestion from GPT 5.5 Xhigh

"Best design is identity_mode: "auto" | "local" | "git".
Recommended auto behavior:
1. Use existing identity first.
   If the exact path already has a local index, reindex local.
   If the Git identity already has an index, reindex Git identity.
2. Use local identity for filesystem-first workflows.
   If the user calls index_folder on a local path and no Git identity index exists yet, use local/<folder>-<hash>.
3. Use Git identity only when clearly intended.
   Use owner/repo when:
   - the repo was already indexed by Git identity
   - the caller explicitly sets identity_mode: "git"
   - config enables Git identity for that project
4. Watcher uses the same resolver.
   No separate _local_repo_id() logic. Manual index_folder, watcher auto-indexing, resolve_repo, and invalidate_cache should agree.
   
This gives safe defaults:
- Local projects stay local.
- Existing Git-indexed repos stay Git-indexed.
- Monorepo users can opt into Git identity.
- Dropping and reindexing won’t unexpectedly switch identity or trigger a larger Git-root rebuild.

I’d implement identity_mode="auto" as the new default, with identity_mode="git" preserving current behavior and identity_mode="local" restoring v1.90.0 behavior."

jgravelle:
Thanks for digging into this @MariusAdrian88 — and for separating the loadability fix from the indexing-behavior change. Both are real, but they're separate problems and I'd rather not let one stall the other.

**On PR #291:** still a merge once the three review items are in (rebase + sweep the three new tools + drop the `digest.py` `hasattr` guard + the JSON-fallback dedupe nit). The loadability work stands on its own.

**On the `index_folder` timeout — quick unblock:** the escape hatch you're describing already exists as a config knob, `git_root_identity` (default `true`, added in v1.95). Set it to `false` in your project or global config and `index_folder` reverts to pre-v1.95 `local/<folder>-<hash>` identity without git-root retargeting:

```jsonc
{ "git_root_identity": false }
```

That should restore the v1.90.0 behavior for your local-only workflow today.

**On the deeper question:** before we redesign the surface (per-call param + watcher alignment), I'd like to confirm what's actually slow. My suspicion is it's not identity resolution (`detect_git_root` is one `git remote get-url origin` call) but the v1.96 subdir-merge behavior — if your folder lives inside a larger git working tree, `index_folder` may now retarget to the full git root and walk a much bigger tree than it did in v1.90. That would explain "indexes in 1.90, times out in 1.96" without any identity confusion at all.

Could you open a separate issue with: the exact folder path, the parent git root (if any), the file count under each, and whether the timeout reproduces with `git_root_identity: false`? With that we can tell apart "identity resolution is slow" vs. "we're now walking 10× more files" vs. "watcher and manual disagree on repo ID" — three different fixes.

Once we know which one it is, the per-call parameter + watcher-alignment design you and GPT sketched is probably right, but I'd rather build it on a verified failure mode than a plausible one.

 -jjg

MariusAdrian88:
@jgravelle PR ready for merge.

The listed review items are fixed.
- Rebase: fixed. origin/main is an ancestor of HEAD, branch is 0 behind and 2 ahead, PR merge state is CLEAN.
- Three new tools sweep: fixed. check_delete_safe.py:160, find_implementations.py:170, and assemble_task_context.py:210 now use index_status_to_tool_error(store.inspect_index(owner, name)).
- digest.py hasattr guard: fixed. digest.py:135 directly calls inspect_index; the test mock now defines it at tests/test_digest.py:157.
- JSON fallback dedupe: fixed. _try_json_fallback() exists at index_store.py:857 and is reused in both fallback paths.
- Verification: focused tests passed, 103 passed in 10.69s. GitHub checks on PR #291 are also all green.

As for the rest, right now I am rather busy so I need to defer. Thank you for the "git_root_identity" hint (kind of cryptic parameter though), but for now I will stay on 1.90.0 until I get the time to dig more into this.

jgravelle:
Merged and shipped as **v1.107.1** on [PyPI](https://pypi.org/project/jcodemunch-mcp/1.107.1/) and [GitHub Releases](https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.107.1) — thank you @MariusAdrian88. Clean turnaround on all four review items (rebase, the three new-tool sweeps, the `digest.py` test-mock cleanup, and the `_try_json_fallback` extraction), and the regression-test matrix you added is exactly the right shape for this kind of correctness fix. Operators hitting a corrupt SQLite or stale legacy JSON now get a real diagnostic instead of "Repository not indexed" — a legitimately better UX layer.

Totally fair to defer the `index_folder` thread; staying on 1.90.0 in the meantime is a sensible call. When you have the bandwidth, an issue with the folder path / parent git root / file count will let us split it cleanly into the right fix.

And — if jCodeMunch is saving you time on your own work, a quick word about it (a star, a tweet, a mention to a teammate) goes further than you'd think for a project of this size. Either way, much appreciated.

 -jjg

MariusAdrian88:
Hi @jgravelle 

The parameter:
{ "git_root_identity": false }

is not present in the config file so it does not adhere to the convention that all settings needs to be surfaced as commented in the config file to serve as self documenting.

I've added it though and I still get timeout's.

AI model messages:
```
⚙ jcode_invalidate_cache [repo=local/MCS_ERP-acfbc7e7]
The old index was deleted. Reindexing the local folder now.

⚙ jcode_index_folder [path=., use_ai_summaries=true, follow_symlinks=false]

MCP error -32001: Request timed out

Thinking: Resolving index issues
I’m facing a timeout issue and need to check if I should continue indexing or if I should resolve the list first. I wonder if running with summaries set to false could help the situation. The user has specifically requested a reindex, but it seems I don’t need to use shell commands for this. I’m thinking I might need to call resolve to move forward. Let’s see what works best!
The reindex request timed out at the tool boundary, so I'm checking whether the index was created or if it needs a lighter retry.
```
Summaries by AI were turned Off in the config.
Something was broken since v1.94 with the git_root_indentity. 

What I would do, was to introduce this in the  production version as not active by default and set it as False.
Whoever wants to use the git identity, they can opt in.

This will make everyone happy and those of us that work locally don't get surprises.

jgravelle:
Thanks for the doc-convention catch @MariusAdrian88 — you're absolutely right. `git_root_identity` was already in `DEFAULTS`, the type-validation table, and had an env-var alias (`JCODEMUNCH_GIT_ROOT_IDENTITY`), but it was missing from the JSONC template — which is exactly the place an operator would look to discover it. That's a real doc bug; surfacing every default in the self-documenting template is the convention and we weren't following our own rule.

**Fixed and shipped as v1.108.1** ([PyPI](https://pypi.org/project/jcodemunch-mcp/1.108.1/) · [GitHub Releases](https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.108.1)) — pure documentation, no behavior change.

**On the still-timing-out:** that's a much more interesting signal. `git_root_identity: false` rules out git-root retargeting as the cause, which means the timeout is coming from something else in the indexing pipeline on your particular tree. AI summaries off rules that out too. So it's most likely sheer file count + first-pass parse cost on `MCS_ERP` exceeding the MCP client's `-32001` request-timeout budget (which is set by the client, not by jCodeMunch).

To narrow it down, could you run from a shell — outside the MCP client's request-timeout — and share a couple of numbers?

```bash
# rough file count + size of what would be walked
find . -type f \( -name '*.py' -o -name '*.js' -o -name '*.ts' -o -name '*.go' -o -name '*.cs' \) \
  ! -path '*/node_modules/*' ! -path '*/.git/*' | wc -l

# real index time, no MCP timeout in the loop
time jcodemunch-mcp index .
```

With (a) the file count and (b) how long the CLI run takes, we can tell apart: "the indexer itself is slow on a 50k-file tree" (we should add per-file progress notifications or stream a partial response), vs. "the indexer is fast but your client is timing out at 60s" (we should document `JCODEMUNCH_MCP_REQUEST_TIMEOUT` overrides per client). Two different fixes; both worth doing once we know which.

**On flipping the default to `false`:** I'm going to push back gently here, and not because your pain isn't real. Two reasons:

1. You confirmed that flipping it didn't fix your timeout — so flipping it for everyone would break the v1.96 monorepo / worktree coalescing for the majority of users *without* fixing your case. It treats a symptom that turned out not to be the symptom.
2. The pre-v1.96 `local/<folder>-<hash>` identity had its own surprises — two clones of the same repo collided silently on a path hash. The current default fixes that quietly for the median user. The escape hatch (now properly documented) is the right shape for the minority who want path-anchored identity.

Once we've isolated whether it's pipeline duration or client timeout, I think the cleaner fix is in that direction — not the identity default. If it turns out to be pipeline duration on huge trees, there's a real engineering item there worth doing for everyone.

Totally fair to stay on 1.90.0 until you've got the bandwidth to dig in. When you do, opening a separate issue with the file count + CLI timing would let us split the problem cleanly.

 -jjg

PR #293, opened 2026-05-12
fix: watcher standby failover takeover resilience

MariusAdrian88:
## Summary

Adds resilient standby failover for the embedded watcher so a second server watching the same folder can wait in standby and take over when the active watcher releases its lock.

## Why

Running multiple MCP servers against the same local repo used to leave only one process able to watch the folder. Other servers that lost the watcher lock had no reliable takeover path, so they could remain unwatched until restarted. This made index freshness depend on which server started first and whether that server stayed alive.

## What was done

- Added standby tracking to `WatcherManager` for folders that fail watcher-lock acquisition.
- Added per-folder release signal files so standby managers can wake up promptly after lock release.
- Added opportunistic takeover from `_auto_watch_if_needed()` before normal reindex/watch fallback.
- Preserved freshness by awaiting `ensure_indexed()` before returning from auto-watch after takeover.
- Hardened cleanup so `remove_folder()` and `stop()` clear standby state and release active locks.
- Reused `_start_watch_task()` for restart/takeover paths so `on_reindex` and watcher options stay consistent.
- Added regression coverage for standby registration, takeover, signal files, retry interval behavior, lock cleanup, and auto-watch freshness.

## Before And After

| Area | Before | After |
| --- | --- | --- |
| Second server startup | Lost the watcher lock and stayed unwatched. | Enters standby and can take over later. |
| Active watcher shutdown | Lock release did not notify standby watchers. | Lock release writes a per-folder signal file. |
| Standby takeover | Required another explicit watch attempt or process restart. | Standby managers retry and react to release signals. |
| Auto-watch freshness | Takeover could return before initial indexing finished. | Auto-watch awaits `ensure_indexed()` before tool dispatch. |
| Watcher restart path | Restart logic manually recreated watch tasks. | Restart uses the shared `_start_watch_task()` helper. |
| `stop()` cleanup | Active tasks and locks could remain tracked/held. | Active tasks are cancelled and locks are released. |
| Retry interval | Standby retry interval was capped at 5 seconds. | Standby retry uses `_takeover_retry_seconds`. |
| Test coverage | No standby failover regression coverage. | Added focused tests across watcher lock, serve, and dynamic watcher paths. |

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -3541,6 +3541,7 @@ async def _auto_watch_if_needed(name: str, arguments: dict, storage_path: Option
     if maybe_takeover is not None:
         result = await maybe_takeover(folder)
         if result.get("status") in {"started", "already_watched"}:
+            await _watcher_manager.ensure_indexed(folder)
             return
 
     # Race-safe reindex, then start watching
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -3536,6 +3536,13 @@ async def _auto_watch_if_needed(name: str, arguments: dict, storage_path: Option
     if _watcher_manager.is_watched(folder):
         return
 
+    # Opportunistic standby takeover before indexing
+    maybe_takeover = getattr(_watcher_manager, "maybe_takeover", None)
+    if maybe_takeover is not None:
+        result = await maybe_takeover(folder)
+        if result.get("status") in {"started", "already_watched"}:
+            return
+
     # Race-safe reindex, then start watching
     try:
         await _watcher_manager.ensure_indexed(folder)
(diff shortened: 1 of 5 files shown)


MariusAdrian88:
@jgravelle 

jgravelle (approved):
Thanks @MariusAdrian88, this is a really clean piece of work — and a surprisingly hard problem solved with the right primitives:

- Per-folder `.signal` file next to the lock so standby managers can wake on release instead of polling.
- `_start_watch_task()` extraction quietly DRYs the crash-restart path that previously duplicated the watch-task setup inline.
- `maybe_takeover()` with throttle + `_takeover_retry_seconds` gives the standby loop a real backoff knob.
- `_auto_watch_if_needed()` awaits `ensure_indexed()` after takeover so freshness is preserved — the tool call sees the index ready, not racing.
- `stop()` now actively cancels active tasks and releases locks; before, locks could outlive the manager. That's worth its own line in release notes.
- Test coverage hits every shape: signal-file payload, standby registration, lock cleanup, throttle, retry interval, `remove_folder` on a standby-only folder, and the full two-manager failover handoff.

CI green across all 9 matrix jobs (Linux + Windows × 3.10–3.13). Approving and merging.

If jCodeMunch keeps proving useful, a star on the repo or a mention to a teammate goes a long way for a project this size.

 -jjg

PR #295, opened 2026-05-12
fix: restore local-first index identity default, make git identity opt-in

MariusAdrian88:
## Summary

Restores `local` (path-hash) as the default index identity mode, making git-root identity an explicit opt-in via `identity_mode: "git"` or the legacy `git_root_identity: true` config knob. Existing indexes of either kind are automatically preserved -- no migration, no invalidation, no changed repo IDs for current users.

The goal is to prioritize compatibility with the working solution that served all users before v1.95. Users who benefit from git-root identity (monorepo subdir merging, `owner/repo` keying) can opt in through the config file without any loss of functionality. Users who work with simple git clones, local folders, or non-git projects get the behavior they relied on by default, without paying the cost of a `git` subprocess probe on every reindex.

## Why

v1.95 introduced git-root identity as a default-on feature. This improved the monorepo workflow (multiple subdirs of the same git root merge into a single index), but it changed the index keying behavior for every existing user -- including those who never needed it.

The consequences for local-first users were concrete:

- **Default cost.** In the default git-identity mode, every new `index_folder` and watcher reindex still needs a `git` subprocess to detect the git root. v1.108.2 fixed the probe-when-off ordering bug and bounded the slowest subprocess paths, but it intentionally did not change the default identity mode.

- **Identity instability.** A folder that was previously indexed as `local/myproject-a1b2c3d4` could silently re-key as `owner/repo` after a version upgrade, orphaning the old index. Users who scripted around repo IDs or relied on `resolve_repo` stability saw unexpected breakage.

- **Unnecessary coupling.** Non-git projects and local-only clones (no remote, no origin) gained nothing from the git-root probe but still paid its cost and complexity.

The principle here is straightforward: a default should work for everyone, and advanced features should be opt-in. Git-root identity is valuable, but it makes assumptions about the user's environment (has git, has an origin remote, wants monorepo merging) that don't hold universally.

### Centralized identity resolution

Added `resolve_index_identity()` in `storage/git_root.py` -- a single function that decides how a local path maps to an index key. It replaces the scattered identity logic that was duplicated across `index_folder.py`, `resolve_repo.py`, and `watcher.py`.

The resolver supports three modes:

| Mode | Behavior |
| --- | --- |
| `config` (default) | Reads `identity_mode` / `git_root_identity` from config. Falls back to `local` when neither is set. |
| `local` | Path-hash identity (`local/<basename>-<sha1[:8]>`). No git subprocess. |
| `git` | Git-root identity (`<owner>/<repo>`). Runs `detect_git_root()`. Enables monorepo subdir merging. |

### Existing index preservation

The resolver checks the index store before making any decision:

- If a `local/...` index already exists for this path, it's returned as-is -- even if the config says `git`.
- If a git-keyed index (`owner/repo`) already covers this path (via its stored `git_root`), it's returned as-is -- even if the config says `local`.
- If both forms exist (should not happen, but guarded), it raises `IdentityModeAmbiguous` rather than silently picking one.
- Explicitly requesting a mode that conflicts with an existing index raises `IdentityModeConflict` with a message telling the user to `invalidate_cache` first.

This means: **no existing user sees a changed repo ID after this change.** The preservation logic runs before any config check or git probe.

### Config surface

The config template now documents both knobs:

```jsonc
// "identity_mode": "git",
//   How index_folder derives the repo identifier for a local path.
//   Existing indexes keep their current identity regardless of this
//   setting -- this only affects NEW indexes.
//
//   Choices:
//     "local" (default) -- repo ID is `local/<basename>-<hash>`.
//       No git subprocess, no remote detection. Fast and portable.
//       Each folder gets its own index. Works for non-git projects,
//       local-only clones, and simple git workflows.
//
//     "git" -- repo ID is `<owner>/<repo>` derived from the origin
//       remote URL. Runs a git subprocess on every index/reindex.
//       Enables monorepo subdir merging (multiple subdirs of the
//       same git root share one index). Requires a git working tree
//       with an origin remote. Falls back to local when detection
//       fails.
//
//   To switch an existing index: run invalidate_cache first, then
//   re-index with the new mode.

// "git_root_identity": true,
//   Deprecated boolean alias for identity_mode. When identity_mode
//   is not set, `true` here is equivalent to `"identity_mode": "git"`.
//   Prefer identity_mode for new configurations.
```

The `index_folder` MCP tool also accepts an `identity_mode` parameter for per-call override.

### Unified call sites

`_local_repo_id()` in `watcher.py`, `_compute_repo_id()` in `resolve_repo.py`, and `_resolve_repo_identity()` in `index_folder.py` all delegate to the central `resolve_index_identity()`, passing the `store` so the preservation logic can inspect existing indexes.

### Test coverage

- **`test_identity_mode.py`** (296 lines, new): Exercises the resolver directly -- default-local without git subprocess, config-mode preservation, explicit mode conflicts, ambiguous state detection, config template content, and integration with watcher/resolve_repo delegation.
- **`test_git_root_identity.py`** (updated): Git-identity tests now pass `identity_mode="git"` explicitly, reflecting that git identity is opt-in. Added `test_knob_off_skips_retarget_git_root_probe` verifying that the git subprocess is never called when identity is local.

## Before and After

| Area | Before (v1.108.2) | After |
| --- | --- | --- |
| Default identity mode | `git` (`git_root_identity: true`) | `local` (path-hash) |
| Git subprocess on reindex | Runs in default git-identity mode; skipped when explicitly local | Only when `identity_mode: "git"` or existing git index |
| New index for simple clone | `owner/repo` (requires working git + origin) | `local/<basename>-<hash>` (no git probe) |
| New index for non-git folder | `local/<basename>-<hash>` (after failed git probe) | `local/<basename>-<hash>` (no git probe attempted) |
| Existing `local/...` index | Preserved | Preserved |
| Existing `owner/repo` index | Preserved | Preserved (auto-detected from stored `git_root`) |
| Monorepo subdir merging | On by default | Opt-in via `identity_mode: "git"` |
| Identity resolution logic | Duplicated in 3 files | Centralized in `resolve_index_identity()` |
| Config documentation | `git_root_identity` only, default `true` | `identity_mode` + legacy `git_root_identity`, default `local` |
| Conflict guard | None (silent re-keying possible) | `IdentityModeConflict` / `IdentityModeAmbiguous` errors |

## Verification

```
.venv\Scripts\python.exe -m pytest tests/test_identity_mode.py tests/test_git_root_identity.py tests/test_resolve_repo.py tests/test_server.py tests/test_watcher_lock.py tests/test_watcher_serve.py -q
```

---

Co-Authored-By: GPT 5.5 XHIGH <noreply@openai.com>

### DIFF
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -1407,9 +1407,11 @@ def generate_template() -> str:
 
   // "identity_mode": "git",
   //   How index_folder derives the repo identifier for a local path.
-  //   Default is local: `local/<basename>-<hash>`, with no git detection
-  //   or retargeting. Uncomment and set to "git" to opt in to v1.95+
-  //   git identity (`<owner>/<repo>`) and monorepo subdir merging.
+  //   Existing indexes keep their current identity. New indexes default
+  //   to local: `local/<basename>-<hash>`, with no git detection or
+  //   retargeting. Uncomment and set to "git" to opt in to v1.95+ git
+  //   identity (`<owner>/<repo>`) and monorepo subdir merging for new
+  //   indexes.
 
   // "git_root_identity": true,
   //   Deprecated legacy boolean alias for identity_mode. Uncomment and set
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -1405,6 +1405,16 @@ def generate_template() -> str:
   //   Enable context providers for enhanced AI summarization.
   //   Set false to disable (faster indexing, less context).
 
+  // "identity_mode": "git",
+  //   How index_folder derives the repo identifier for a local path.
+  //   Default is local: `local/<basename>-<hash>`, with no git detection
+  //   or retargeting. Uncomment and set to "git" to opt in to v1.95+
+  //   git identity (`<owner>/<repo>`) and monorepo subdir merging.
+
+  // "git_root_identity": true,
+  //   Deprecated legacy boolean alias for identity_mode. Uncomment and set
+  //   true to opt in to git identity when identity_mode is not set.
+
   // === Meta Response Control ===
   // Allowlist of _meta fields to include in responses.
   // [] (default) = no _meta at all (maximum token savings).
--- a/src/jcodemunch_mcp/config.py
+++ b/src/jcodemunch_mcp/config.py
@@ -1407,15 +1407,29 @@ def generate_template() -> str:
 
   // "identity_mode": "git",
   //   How index_folder derives the repo identifier for a local path.
-  //   Existing indexes keep their current identity. New indexes default
-  //   to local: `local/<basename>-<hash>`, with no git detection or
-  //   retargeting. Uncomment and set to "git" to opt in to v1.95+ git
-  //   identity (`<owner>/<repo>`) and monorepo subdir merging for new
-  //   indexes.
+  //   Existing indexes keep their current identity regardless of this
+  //   setting — this only affects NEW indexes.
+  //
+  //   Choices:
+  //     "local" (default) — repo ID is `local/<basename>-<hash>`.
+  //       No git subprocess, no remote detection. Fast and portable.
+  //       Each folder gets its own index. Works for non-git projects,
+  //       local-only clones, and simple git workflows.
+  //
+  //     "git" — repo ID is `<owner>/<repo>` derived from the origin
+  //       remote URL. Runs a git subprocess on every index/reindex.
+  //       Enables monorepo subdir merging (multiple subdirs of the
+  //       same git root share one index). Requires a git working tree
+  //       with an origin remote. Falls back to local when detection
+  //       fails.
+  //
+  //   To switch an existing index: run invalidate_cache first, then
+  //   re-index with the new mode.
 
   // "git_root_identity": true,
-  //   Deprecated legacy boolean alias for identity_mode. Uncomment and set
-  //   true to opt in to git identity when identity_mode is not set.
+  //   Deprecated boolean alias for identity_mode. When identity_mode
+  //   is not set, `true` here is equivalent to `"identity_mode": "git"`.
+  //   Prefer identity_mode for new configurations.
 
   // === Meta Response Control ===
   // Allowlist of _meta fields to include in responses.
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -782,6 +782,12 @@ def _build_tools_list() -> list[Tool]:
                         "type": "array",
                         "items": {"type": "string"},
                         "description": "Optional list of explicit paths to index. When provided, the directory walk is skipped; only these files (and the contents of any directories in the list) are indexed. Entries may be absolute or relative to `path`. Useful for batch-indexing exactly the files an agent already knows about — e.g. the source files git just touched, the changeset for a PR, or an rg / fd match list. Validation matches the walk path (outside-root, traversal, symlink-escape, oversize, unsupported-extension all warn-and-skip)."
+                    },
+                    "identity_mode": {
+                        "type": "string",
+                        "enum": ["config", "local", "git"],
+                        "description": "How to derive the local-folder repo identity. Default config keeps local-first behavior unless an existing git-keyed index or config opts in. Use 'git' to opt in to git-root identity and monorepo subdir merging.",
+                        "default": "config"
                     }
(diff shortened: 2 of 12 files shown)


MariusAdrian88:
@jgravelle 

I rebased the PR so this is ready for review/merge.

jgravelle:
Hi @MariusAdrian88 — thanks again for both this PR and #293. #293 is shipped as **v1.108.5** ([PyPI](https://pypi.org/project/jcodemunch-mcp/1.108.5/) · [Release](https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.108.5)). The standby/takeover design is exactly the right shape for the multi-server case — clean handoff, no polling, lifecycle properly hardened.

I read through #295 carefully and Claude's recommendation is to merge it as v1.109.0, but a couple of pre-merge notes before you rebase:

**1. Rebase needed.** main now contains the #293 watcher changes. Your #295 diff touches `_local_repo_id()` / `_do_reindex()` / `sync_folders()` / `_stop_watching()` while #293 added `_start_watch_task()`, `maybe_takeover()`, and standby state inside `WatcherManager`. The conflicts should be mechanical — both sets of changes are in the same file but mostly in different functions. Could you rebase on main and force-push?

**2. Framing tweak.** Your PR body says v1.108.2 only "mitigated" the probe-when-off bug, but [eb810c8](https://github.com/jgravelle/jcodemunch-mcp/commit/eb810c8) actually did fix the ordering — `detect_git_root()` no longer runs when `git_root_identity` is false. The real value of #295 is the **default flip + centralization + conflict guards**, not the probe fix. Worth mentioning in the body so the merge story is "safer default + cleaner architecture" rather than "still fixing probe regression."

**3. The default flip is the consequential change**, and I want to be straight about that: v1.95 → v1.108 spent several releases polishing git-root identity as the default, so flipping it back to local-first is a real philosophical reversal. The things that made me OK with it:

- `_local_identity_if_present` + `_existing_git_identity` short-circuit before any decision, so existing indexes stay put on upgrade
- `IdentityModeConflict` / `IdentityModeAmbiguous` mean silent re-key becomes a loud error with remediation
- Monorepo subdir-merging is preserved as `identity_mode: "git"` opt-in, not removed

After the rebase + small body tweak I'll run a smoke pass (index_folder on non-git folder, git clone with origin, same clone with `identity_mode: "git"`, then resolve_repo on each) and merge as **v1.109.0** (minor bump because of the default change).

 -jjg

MariusAdrian88:
@jgravelle 
Thank you!
All items in the review should be fixed now.

jgravelle:
Merged and shipped as **v1.108.6** on [PyPI](https://pypi.org/project/jcodemunch-mcp/1.108.6/) and [GitHub Releases](https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.108.6) — thank you @MariusAdrian88.

This is the principled close to the architectural thread you opened in #291. The right call is exactly what you proposed: defaults should serve everyone, advanced features should be opt-in. The pieces that landed in this PR that I want to call out specifically:

- **`resolve_index_identity()` as a single source of truth.** Replacing three duplicated identity-resolution paths (`_local_repo_id` in watcher.py, `_compute_repo_id` in resolve_repo.py, `_resolve_repo_identity` in index_folder.py) was the structural win the codebase needed. Future identity-related work has exactly one place to touch now.
- **The preservation contract is real and tested.** The "no existing user sees a changed repo ID" claim isn't just words — `_local_identity_if_present` + `_existing_git_identity` probes run *before* any config check, and the explicit `IdentityModeConflict` / `IdentityModeAmbiguous` errors guard the corners where silent re-keying would otherwise sneak through. 10 directly-targeted tests in the new `test_identity_mode.py` cover all four preservation corners + the ambiguous-state guard + the no-subprocess fast path. Exactly the right matrix.
- **The `git_root` round-trip through both index store backends** was a small detail that makes the existing-git-index probe cheap in the common case — nice touch.

I bundled one small additional change into the 1.108.6 release: an "honest-hint" caveat on `check_delete_safe` when the verdict is `safe_to_delete` but no runtime traces have been ingested. The runtime channel previously just didn't fire in that state, which let `safe_to_delete` implicitly claim coverage it couldn't prove. Back-ported from a UX pattern we just shipped in jdatamunch-mcp's `check_column_drop_safe`. Sibling-parity work paying it forward.

And — if jCodeMunch has been useful to you, a quick word about it goes further than you'd think for a project of this size. A star, a tweet, a mention to a teammate who deals with the same monorepo-identity headaches you've been working through. Either way, much appreciated.

 -jjg

PR #311, opened 2026-05-24
feat(parser): add Astro v6 (.astro) language support

gokhanozdemir:
## Summary.

Adds full parsing support for `.astro` files (Astro v6), bringing the
language count to 74. Astro components are mixed-language files —
TypeScript/JavaScript frontmatter between `---` fences, an HTML template
body, and optional `<script>` and `<style>` blocks — structurally aligned
with the existing Razor/Blazor mixed-language strategy.

This PR also hardens Astro behavior for malformed/real-world files and improves
template-aware import extraction.

## Implementation.

Rather than adding a new binary dependency, the parser follows the same
**recursive delegation pattern** already used for Razor:

- **Frontmatter** (`--- ... ---`) → re-parsed via `parse_file()`
- **`<script>` blocks** → re-parsed via `parse_file()`
- **`<script src="...">` refs** → emitted as `kind: function` symbols
- **`<style>` blocks** → emitted as `kind: constant` symbols (same as Razor)
- **File-level component** → emitted as `kind: class` symbol (named after the
stem)

Grammar reference:
[virchau13/tree-sitter-astro](https://github.com/virchau13/tree-sitter-astro)
is used as a structural guide for node types (`frontmatter`,
`script_element`, `style_element`), not as a runtime dependency.

**Forward-compat:** `ASTRO_SPEC.ts_language = "astro"` is already wired — if
`tree-sitter-language-pack` bundles the Astro grammar in a future release, the
generic spec-walk path activates automatically with no caller changes.

## Additional hardening included.

- Robust frontmatter splitting helper:
  - Handles UTF-8 BOM.
  - Handles CRLF.
  - Gracefully handles malformed or unclosed fences (fail-soft, no crash).
- Template `id="..."` extraction added for Astro:
  - IDs inside HTML comments are ignored.
  - Line/byte offsets map back to original source positions.
- Script parsing improvements:
  - `lang="ts|tsx|jsx"` inference for inline `<script>` blocks.
  - JSON/JSON-LD script blocks (e.g. `type="application/ld+json"`) are skipped.
- Import extraction improvements:
  - Frontmatter ESM imports remain extracted as real edges.
  - Synthetic template component usage edges are emitted when template components
    are used without explicit imports.
  - ~~Synthetic template component usage edges are emitted when template components
    are used without explicit imports, marked with:~~
    - ~~`synthetic: true`~~
    - ~~`reason: "template_component_usage"`~~
  - ~~Edge dedup now includes synthetic metadata in keying.~~
  - Edge shape is now consistent with Vue/Nuxt producers: `{specifier, names}`.
  - Dedup keys now use `(specifier, names)` only.
- `.astro` added to JS-like extension handling where behavior is generic:
  - Import specifier resolution extension set.
  - Package registry JS/TS-like language grouping.
  - Hook extension hints for code-navigation guidance.

## Files Changed.

| File | Change |
|------|--------|
| `src/jcodemunch_mcp/parser/languages.py` | `ASTRO_SPEC`, `.astro` extension mapping, `LANGUAGE_REGISTRY` entry |
| `src/jcodemunch_mcp/parser/extractor.py` | `_ASTRO_*` regex/constants, ~~hardened frontmatter splitter~~, template ID extraction, script language inference, JSON script skip, symbol dedup, `_parse_astro_symbols()`, dispatch case |
| `src/jcodemunch_mcp/parser/imports.py` | ~~Hardened Astro frontmatter extraction helper~~, template component synthetic edges, dedup logic, `.astro` in `_JS_EXTENSIONS`, `_extract_astro_imports()` registration retained |
| `src/jcodemunch_mcp/parser/astro_shared.py` | New shared Astro helpers: frontmatter split + HTML-comment masking with offset preservation |
| `src/jcodemunch_mcp/tools/plan_refactoring.py` | `"astro"` entries in `_IMPORT_PATTERNS` and `_DEF_PATTERNS` |
| `src/jcodemunch_mcp/tools/package_registry.py` | Includes `"astro"` in JS/TS-like package heuristics |
| `src/jcodemunch_mcp/cli/hooks.py` | Adds `.astro` to code-extension hint set |
| `tests/fixtures/astro/` | Existing 7 AstroWind-derived fixtures + new hardening fixtures (`sample_bom_crlf.astro`, `sample_malformed_frontmatter.astro`, `sample_multi_script.astro`) |
| `tests/test_parser.py` | Existing Astro coverage retained |
| `tests/test_astro.py` | New focused Astro hardening suite + shared-helper regression assertions |
| `tests/test_find_importers.py` | Added `.astro` specifier-resolution regression |
| `tests/test_languages.py` | Added direct `.astro` language mapping assertion |

## Symbols Extracted.

From any `.astro` file the parser surfaces:

- Component class (filename stem).
- TypeScript symbols from frontmatter: functions, interfaces, type aliases,
exported consts (`getStaticPaths`, `prerender`, `Props`, etc.).
- JavaScript/TypeScript functions and types from inline `<script>` blocks.
- `<script src="...">` external references.
- `<style>` block markers.
- Template `id` attributes as constant symbols (comments ignored).

## Real-world Patterns Tested (AstroWind).

Fixtures are modelled on actual files from the AstroWind starter:

| Fixture | Pattern |
|---------|---------|
| `DynamicRoute.astro` | `getStaticPaths satisfies GetStaticPaths` · `type Props = InferGetStaticPropsType<...>` |
| `WidgetComponent.astro` | `import type { Features as Props } from '~/types'` · `await Astro.slots.render()` |
| `BaseLayout.astro` | `export interface Props` · `import type { X as Y }` · `astrowind:config` |
| `ClientIsland.astro` | `<script define:vars={{ ... }}>` · `prerender = false` |

## Test Results.

4397 passed, 12 skipped in 50.19s
Astro-focused regression checks after the latest refactor:
- `tests/test_astro.py tests/test_parser.py -k astro` → `33 passed, 10 deselected`
- `tests/test_find_importers.py -k "astro_component or vue_template_components_in_index"` → `2 passed, 127 deselected`

No regressions observed in the updated Astro paths.

## Known Limitation.

The `satisfies` operator (TypeScript 4.9+) causes the current version of
`tree-sitter-language-pack`'s TypeScript grammar to merge two adjacent
`export const` declarations into a single AST node. When `export const
prerender`
immediately precedes `export const getStaticPaths = (...) satisfies T`, only
`getStaticPaths` is extracted. Documented in the relevant test docstring.
This is a grammar version constraint, not a parser logic issue.


jgravelle (changes requested):
Thanks for this, Gökhan. The careful work shows: the BOM/CRLF/malformed-fence splitter, the comment-masking that preserves byte offsets, and the `lang=` inference plus JSON-script skip are exactly the real-world edge cases that bite, and your test coverage for them (the dedicated `test_astro.py` plus the line-offset assertions) is excellent. Mirroring the Razor mixed-language strategy was the right structural call, and the `ts_language="astro"` forward-compat marker is a nice touch.

A few things need to change before I can merge, mostly repo hygiene rather than the parser itself.

**Blocking:**

1. **`.DS_Store` is committed and survives into the final tree.** It's added in the first commit and modified in commits 2 and 4, but never removed (commit 4 only deletes `.codex/config.toml`). It isn't tracked on `main` today. Could you `git rm --cached .DS_Store` and add `.DS_Store` to `.gitignore`?

2. **`uv.lock` got fully rewritten.** Your local uv bumped it from `revision = 1` to `revision = 3` (adding `upload-time` fields throughout, plus a `backports-tarfile` entry). That accounts for nearly the entire diff size, and none of it relates to Astro. Please revert `uv.lock` to match `main`.

3. **The health-radar action bump (commit 3) is unrelated.** Bumping the action ref and adding the in-repo `pip install` fallback is a reasonable idea on its own, but it touches release/CI plumbing and is out of scope for language support. Could you split it into a separate PR so we can review it independently?

**Worth tightening:**

4. `_split_astro_frontmatter` and `_mask_html_comments_keep_offsets` are duplicated verbatim in both `extractor.py` and `imports.py`. Two copies of offset-sensitive logic will drift over time. Mind factoring them into one shared spot?

5. The synthetic template-usage edges carry new `synthetic: true` / `reason` keys, but nothing downstream reads them yet, and our existing producers of the same kind of edge (Vue template components, Nuxt auto-imports) emit plain `{specifier, names}`. To keep the edge shape consistent, either drop the extra keys here or let's discuss wiring them through Vue/Nuxt too in a follow-up. Happy either way, just want one convention.

The `satisfies` limitation is a genuine tree-sitter-typescript grammar constraint and your docstring documents it well, so no concern there.

One more ask: the branch predates several releases (`main` is on 1.108.22 now). A rebase on current `main` will clear the lockfile noise and keep the diff focused on the actual feature.

Once the `.DS_Store`, `uv.lock`, and action-bump items are sorted, this is in good shape. Appreciate you picking up a whole new language.

 -jjg

gokhanozdemir:
Thanks so much for the careful review, jjg, and for the clear guidance. I really appreciate the detailed feedback.

I’ve now addressed all requested items:

1. Removed `.DS_Store` from tracking and added `.DS_Store` to `.gitignore`.
2. Reverted `uv.lock` to match current `main` (no Astro-unrelated lockfile churn).
3. Removed the health-radar action bump and fallback changes from this PR (will send separately if still useful).
4. Factored `_split_astro_frontmatter` and `_mask_html_comments_keep_offsets` into a shared helper to avoid drift.
5. Normalized Astro template-usage edges to the existing `{specifier, names}` shape (dropped `synthetic`/`reason` keys for consistency).

I actually  soft-reset/recomposed the commit set to clean up incidental noise and rebased on current `main` so the diff is now focused on Astro support only. 

Thanks again for the thoughtful review and encouragement. Soon I'll be coming with more updates I hope. Thank you for this great tool. 

Warm regards from Türkiye,
Gökhan

jgravelle (approved):
Gökhan, this is exactly the turnaround I hoped for. Thank you.

I verified every item against current `main` rather than taking the checklist on faith, and it all holds:

- `.DS_Store` untracked and now ignored.
- `uv.lock` back to match `main`, zero lockfile churn.
- Health-radar action bump pulled out of scope.
- The offset-sensitive helpers (`split_astro_frontmatter`, `mask_html_comments_keep_offsets`) now live in one place, `astro_shared.py`, and both `extractor.py` and `imports.py` import them. No more drift risk.
- Template edges normalized to `{specifier, names}`, consistent with the Vue/Nuxt producers.

The rebase is clean (diff is Astro-only now), and the full matrix is green on Linux and Windows across 3.10 through 3.13. The radar's B to C is just the large-diff churn penalty for landing a whole new language, which is expected and fine.

One harmless leftover for whenever you're next in here, no action needed for this merge: `_HTML_COMMENT_RE` in `imports.py` is now unused, since the masking moved into `astro_shared`. `find_dead_code` will point right at it.

Merging now. Astro is language 74. Thanks for picking up a whole new grammar and for the careful hardening on BOM, CRLF, malformed fences, and comment masking. If jCodeMunch has earned its keep, a word to other Astro and TypeScript folks who'd benefit goes a long way.

 -jjg

jgravelle:
Shipped in **v1.108.23**, Gökhan. Astro is officially language 74.

PyPI is still mid-review on an unrelated quarantine flag, so for now grab it from the GitHub-release wheel rather than a bare `pip install`:

```
pip install https://github.com/jgravelle/jcodemunch-mcp/releases/download/v1.108.23/jcodemunch_mcp-1.108.23-py3-none-any.whl
```

Release: https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.108.23

Thanks again for the clean rebase and the careful hardening work. If you know other Astro or TypeScript folks who'd get use out of jCodeMunch, a mention goes a long way.

 -jjg

gokhanozdemir:
I will, I definitely will. Cloudflare pushes Astro a lot and I think it's a Next.Js killer. Now this one is the only retrieval layer who supports Astro. 

I hope this quarantine flag goes away soon. 🍀

PR #317, opened 2026-05-26
fix(security): is_secret_file should match basename only, not full path

Myonsite:
## Problem

`is_secret_file()` runs `fnmatch.fnmatch(path_lower, pattern)` against the **full file path** as a fallback after the basename match. Because `SECRET_PATTERNS` contains broad globs like `*secret*` (plus `id_rsa.*`, `*.env`, etc.), any file path containing the substring `secret` anywhere in its directory chain is treated as a secret file and excluded from indexing.

Concrete example — a perfectly normal service file:

```
services/secrets-manager/app/main.py
```

The basename `main.py` does not match any secret pattern, but the full path `services/secrets-manager/app/main.py` matches the `*secret*` glob. Result: the **entire `services/secrets-manager/` subtree is silently dropped** from the index, taking every file under it with it.

## Where it came from (real impact)

Discovered 2026-05-26 by a downstream user of `jcodemunch-mcp` (`pip install jcodemunch-mcp`, currently locally patched while waiting for upstream merge) while running `index_folder` on a microservices platform repo:

- Before fix: **354 secret-skips** reported during indexing — all false positives caused by this substring match. An entire `services/secrets-manager/` and `services/secrets-bridge/` got dropped.
- After fix: same repo indexed **432 → 497 files (+65 source)**, secret-skips collapsed to **68 legitimate ones**, surfacing **+5,133 symbols** that had been invisible to `search_symbols` / `get_symbol`.

## Fix

Drop the full-path `fnmatch` fallback. Every pattern in `SECRET_PATTERNS` is a basename pattern (`.env`, `.env.*`, `*.pem`, `id_rsa`, `id_rsa.*`, `credentials.json`, `service-account*.json`, `*secret*`, …); applying them to the full path contributed **zero true positives** and produced the false positives above.

The maintainer's existing `_SECRET_DOC_EXEMPT_PATTERNS` doc-extension safe-list for `*secret*` (so `docs/secrets-handling.md` is not flagged) is preserved unchanged.

```diff
-    name = os.path.basename(file_path).lower()
-    path_lower = file_path.lower()
-    _, ext = os.path.splitext(name)
+    name = os.path.basename(file_path).lower()
+    _, ext = os.path.splitext(name)
 
     excluded = set(_config.get("exclude_secret_patterns", []))
     for pattern in SECRET_PATTERNS:
         if pattern in excluded:
             continue
         if pattern in _SECRET_DOC_EXEMPT_PATTERNS and ext in _SECRET_GLOB_SAFE_EXTENSIONS:
             continue
         if fnmatch.fnmatch(name, pattern):
             return True
-        # Also check full path for patterns like .env.*
-        if fnmatch.fnmatch(path_lower, pattern):
-            return True
     return False
```

Docstring updated to record the basename-only contract and the motivating bug.

## Tests

Added 17 new parameterised cases to `tests/test_security.py::TestSecretDetection`, covering both directions of the fix:

- `test_non_secret_files_under_secret_named_dirs_not_flagged` — paths like `services/secrets-manager/app/main.py`, `services/secret-config/foo.py`, `app/secret_helpers/utils.py`, `internal/secrets/router.go`, `pkg/secret/loader.rs`, `id_rsa_helpers/main.py` must **not** be flagged.
- `test_real_secrets_inside_secret_named_dirs_still_flagged` — real credential files inside a `secret`-named directory (`services/secrets-manager/.env`, `.../keys/id_rsa`, `.../certs/server.pem`, `.../credentials.json`, `deploy/secret-config/private.key`, `mysecretstuff.py`) must **still** be flagged.

### Test suite results

- **Without the fix** (new tests applied, `security.py` reverted): 7 of the new regression cases fail — confirms the bug is real and the tests catch it.
- **With the fix**: `tests/test_security.py` passes **133/133**. Full repo suite (excluding the unrelated `test_schema_budget` token-budget gate) passes **4418 passed, 11 skipped, 0 failed** in ~80s.

## CHANGELOG

Entry added under `[Unreleased] / Fixed`.

## Notes for the maintainer

- The downstream user is currently running `jcodemunch-mcp 0.2.30` with a local patch file (`security.py.bak.20260526` kept as backup); the patch is identical in spirit to this PR but the upstream code is a few hundred commits ahead, so this PR applies the same idea cleanly on top of `main` at `b8cfe16` (v1.108.24).
- CLA signature will be handled via the CLA Assistant prompt on this PR per `CONTRIBUTING.md`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to jcodemunch-mcp are documented here.
 
 ## [Unreleased]
 
+### Fixed
+
+- `is_secret_file()` no longer matches `SECRET_PATTERNS` against the full
+  file path — only the basename. Previously the loop fell through to
+  `fnmatch.fnmatch(path_lower, pattern)`, which caused any path containing
+  the substring `secret` (e.g. a `services/secrets-manager/` directory) to
+  match the `*secret*` glob, silently excluding every file underneath from
+  the index. All shipped `SECRET_PATTERNS` entries are basename patterns,
+  so the full-path branch contributed zero true positives. Documentation-
+  extension safe-list logic for `*secret*` is preserved unchanged.
+
 ## [1.108.24] - 2026-05-26 - check_edit_safe edit-safety preflight
 
 New tool `check_edit_safe`: the edit-safety companion to `check_delete_safe`.
--- a/src/jcodemunch_mcp/security.py
+++ b/src/jcodemunch_mcp/security.py
@@ -140,21 +140,28 @@ _SECRET_DOC_EXEMPT_PATTERNS: frozenset[str] = frozenset({"*secret*"})
 def is_secret_file(file_path: str) -> bool:
     """Check if a file path matches known secret file patterns.
 
-    Uses filename/extension matching, not content inspection. The broad
-    ``*secret*`` glob is not applied to known documentation extensions
-    (.md, .rst, .txt, .adoc, .html, .ipynb, etc.) to avoid false positives
-    on files like ``docs/secrets-handling.md``.
+    Matches against the file's basename only — never the full path. Every
+    pattern in :data:`SECRET_PATTERNS` is a basename pattern (``.env``,
+    ``*.pem``, ``id_rsa``, ``*secret*``, etc.); applying them to the full
+    path produced false positives that excluded whole legitimate subtrees.
+    For example, ``services/secrets-manager/app/main.py`` would match
+    ``*secret*`` against the full path even though ``main.py`` is clearly
+    not a credential file, silently dropping the entire service from the
+    index.
+
+    The broad ``*secret*`` glob is also skipped for known documentation
+    extensions (.md, .rst, .txt, .adoc, .html, .ipynb, etc.) to avoid
+    false positives on files like ``docs/secrets-handling.md``.
 
     Args:
         file_path: Relative file path (forward slashes).
 
     Returns:
-        True if the file matches a secret pattern.
+        True if the file's basename matches a secret pattern.
     """
     import fnmatch
 
     name = os.path.basename(file_path).lower()
-    path_lower = file_path.lower()
     _, ext = os.path.splitext(name)
 
     excluded = set(_config.get("exclude_secret_patterns", []))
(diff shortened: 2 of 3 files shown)


jgravelle:
Thank you for this, Mayank. The diagnosis is excellent and the bug is real: running `*secret*` as a substring against the full path means any directory with "secret" in its name drops its whole subtree, and your `services/secrets-manager/` example with the 354 → 68 secret-skip and +5,133-symbol before/after is exactly the kind of concrete, reproducible report that makes a fix easy to reason about. That false positive needed fixing.

One thing gave me pause about going basename-only, though. The full-path `*secret*` match was quietly doing a second job: catching real credentials stored as **data files inside a conventional secret-store directory** with an innocuous basename. The common convention here is Kubernetes / Terraform / Docker secrets, e.g.:

```
config/secrets/database.yaml
app/secrets/oauth.json
infra/secret/backend.tfvars
```

None of those basenames match any pattern, so under pure basename-only matching they'd silently start getting indexed. For a tool whose whole job is to make a repo searchable, indexing those would be a real regression in the unsafe direction.

So I've landed a fix on `main` that keeps your win and closes that gap. It's two-stage:

1. **Basename patterns** for everything (your false positive is gone, since `secrets-manager` is no longer matched).
2. A **secret-store directory** rule: a file is still treated as secret material when it sits under a *whole-segment* `secret`/`secrets` directory **and** has a non-source data/credential extension (`.yaml`, `.json`, `.env`, `.pem`, `.tfvars`, ...).

That carve-out is what threads the needle against your own test cases: `internal/secrets/router.go` and `pkg/secret/loader.rs` stay un-flagged (source files are code that *handles* secrets, not secret material), and `secrets-manager` stays un-flagged (it's a service name, not a store), while `config/secrets/database.yaml` stays excluded. I added regression tests in both directions, including the data-file-in-secret-dir cases.

Because this went in as a clean-room implementation on our side, there's no CLA needed from you here, and your report is credited in the CHANGELOG. I'm going to close this PR in favor of that change, which will go out in the next release (PyPI's mid-incident right now, so timing is a little fluid).

Genuinely appreciate the careful write-up. If jCodeMunch is saving you time on that microservices platform, the best thanks is telling another dev or dropping a star, it helps more people find it.

 -jjg

Issue #125, opened 2026-03-18
index_folder(".") traverses outside repo boundary in Codex full-access mode


oneart-dev:
I reproduced unexpected filesystem traversal while using jcodemunch_mcp.index_folder with path: "." from inside my repo workspace.

Important context: I was running Codex in full access mode (no filesystem sandbox), so the MCP had permission to read broadly. However, based on SECURITY.md, I still expected jCodemunch path-boundary protections to keep indexing scoped to the requested workspace/repo root.

**Expected behavior**
index_folder(".") should index only the current repo/workspace directory (for example /private/var/www/sites/my-site) and not escape that boundary. And probably have protection for traverse of root folder of entire system.

**Observed behavior**
Indexing appeared to continue beyond the repo root and traverse all files on my machine without any warnings.

Even in full-access environments, tool-level path traversal prevention should still enforce boundary checks for user-requested scopes.

Environment
Macos Tahoe 26.3
Latest codex mac app and lates codex cli

PR #322, opened 2026-05-30
fix: allow Windows UNC share child indexing

### LINKED ISSUE #321: Windows UNC share child paths trip broad-root guard
## Problem

On Windows, a project that lives directly under a UNC share root is rejected by the `index_folder()` broad-root guard even though the requested path is a concrete project directory, not the share root itself.

Example layout:

```text
\\server\share\my-project
```

or a mapped drive path that resolves to the same UNC path.

Running:

```powershell
jcodemunch-mcp index X:\my-project --no-ai-summaries
```

can fail with:

```text
Resolved path '\\server\share\my-project' is too broad to index safely (fewer than 3 path components).
```

## Root cause

The guard in `src/jcodemunch_mcp/tools/index_folder.py` currently checks `len(folder_path.parts) < 3`.

On Windows, `pathlib` represents a UNC path like this:

```python
Path(r"\\server\share\my-project").parts
# ('\\\\server\\share\\', 'my-project')
# len == 2
```

The UNC server/share root is stored as a single anchor component, so a real project directory directly under the share is treated the same as a dangerously broad shallow path.

The share root itself should still be rejected:

```python
Path(r"\\server\share").parts
# ('\\\\server\\share\\',)
# len == 1
```

## Related prior work

This is related to, but not the same as:

- #125: broad-root guard added after `index_folder(".")` could resolve against the MCP process CWD.
- #175: `trusted_folders` allowlist for intentionally broad roots.

This issue is narrower: a normal Windows UNC project path under a named share is counted too shallow because of `pathlib` UNC anchor semantics.

## Expected behavior

- `\\server\share` remains rejected as too broad.
- `\\server\share\my-project` is allowed to proceed through normal indexing, subject to the existing filters and file limits.
- Local drive behavior remains unchanged, e.g. `C:\repo` still has its existing broad-root behavior.

## Proposed fix

Use a small path-depth helper for the broad-root guard that counts a Windows UNC drive/anchor as server + share for safety-depth purposes. That makes `\\server\share\my-project` logically depth 3 while keeping `\\server\share` logically depth 2 and still blocked.

I have a focused patch with Windows UNC tests and will open a PR.

JayceeB1:
## Summary

Fixes #321.

Windows `pathlib` treats a UNC share root like `\\server\share\` as one anchor component, so `\\server\share\repo` has only two `Path.parts`. The broad-root guard currently rejects any non-container path with fewer than three parts, which incorrectly blocks projects mounted directly under a network share or via a mapped drive that resolves to UNC.

This PR adds a small helper for the broad-root safety depth calculation:

- UNC share roots count as server + share for guard-depth purposes.
- `\\server\share\repo` is allowed to proceed through normal indexing.
- `\\server\share` remains rejected as too broad.
- Local drive behavior is unchanged.

## Existing issues / PRs checked

I checked for existing open and closed issues/PRs around `UNC`, `network share`, and `too broad to index safely`; I did not find a duplicate. Related prior work:

- #125 added the broad-root guard after `index_folder(".")` could resolve against the MCP process CWD.
- #175 added `trusted_folders` for intentionally broad roots.

This is narrower than both: a normal Windows UNC project path is misclassified because of `pathlib` UNC anchor semantics.

## Testing

- `PYTHONPATH=src python -m pytest tests/test_tools.py::TestTrustedFolders tests/test_tools.py::TestContainerDetection tests/test_tools.py::TestWindowsUNCPathSafety -q`
- `PYTHONPATH=src python -m pytest tests/test_tools.py -q`
- `uv run pytest tests/ -q` -> `4431 passed, 14 skipped`
- Real Windows repro check with local patch: indexing a mapped drive project path that resolves to `\\server\share\repo` succeeds instead of failing the broad-root guard.

### DIFF
--- a/src/jcodemunch_mcp/tools/index_folder.py
+++ b/src/jcodemunch_mcp/tools/index_folder.py
@@ -128,6 +128,20 @@ def _is_container() -> bool:
     return False
 
 
+def _path_safety_part_count(path: Path) -> int:
+    r"""Count path components for the broad-root guard.
+
+    On Windows, pathlib stores a UNC share root such as ``\\server\share\`` as
+    one anchor component. Treat that anchor as server + share so
+    ``\\server\share\repo`` has the same logical depth as ``C:\Users\repo``,
+    while the share root itself remains too broad.
+    """
+    count = len(path.parts)
+    if os.name == "nt" and str(path.drive).startswith("\\\\"):
+        count += 1
+    return count
+
+
 @lru_cache(maxsize=512)
 def _is_trusted(
     folder_path: Path, trusted_folders: tuple, whitelist_mode: bool = True
@@ -952,7 +966,8 @@ def index_folder(
     # /workspace works out of the box while bare "/" is still rejected.
     container = _is_container()
     _MIN_PATH_PARTS = 2 if container else 3
-    if len(folder_path.parts) < _MIN_PATH_PARTS:
+    path_part_count = _path_safety_part_count(folder_path)
+    if path_part_count < _MIN_PATH_PARTS:
         if not is_trusted:
             error_msg = (
                 f"Resolved path '{folder_path}' is too broad to index safely "
@@ -971,7 +986,7 @@ def index_folder(
         logger.warning(warning_msg)
         warnings.append(warning_msg)
 
-    if container and len(folder_path.parts) < 3:
+    if container and path_part_count < 3:
         warning_msg = (
             f"Container environment detected — allowing shallow path '{folder_path}'. "
             "The minimum path depth has been relaxed from 3 to 2 components."
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1089,6 +1089,78 @@ class TestContainerDetection:
         assert "too broad to index safely" in result["error"]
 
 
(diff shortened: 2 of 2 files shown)


jgravelle (approved):
Thanks for turning this around so fast, and for the airtight root-cause work on both the issue and the PR.

I verified the fix against real Windows pathlib (Python 3.12): `\\server\share\repo` resolves to drive `\\server\share` and your helper scores it depth 3 (allowed), `\\server\share` scores 2 (still rejected), and `C:\repo` stays at 2 (local behavior unchanged). Exactly the matrix you specified.

What I checked beyond that:

- Scope is clean: only `index_folder.py` and `tests/test_tools.py`, nothing hidden.
- Both guard sites route through the new helper (the hard reject and the container warning), so a UNC mount inside a container won't emit a misleading "relaxed from 3 to 2" message.
- `os` was already imported, and the `os.name == "nt"` gate keeps POSIX paths untouched.
- Tests cover share-root-rejected and project-under-share-allowed, correctly Windows-gated.

Clean, focused, well-tested. Merging. If jCodeMunch is earning its keep for you, a word to your team or a star helps more people find it.

PR #337, opened 2026-06-18
feat: index templating-language files over a supported underlying language (#336)

### LINKED ISSUE #336: Index templating-language files (Jinja2/Twig/Liquid/Handlebars) over a supported underlying source language
Severity: Enhancement / Medium (templated source files — e.g. Jinja2 templates *of* TypeScript — are currently invisible to the index, so every real symbol inside them is dropped; this is a common shape in codegen, scaffolding, and infra-as-code repos)

## Summary

jCodeMunch should index **templating-language files whose body is a supported source language** — e.g. a Jinja2 template of a TypeScript file (`foo.ts.j2`), a Jinja template of Python (`service.py.jinja`), or a Twig template of TS (`widget.ts.twig`). The real source symbols (functions, classes, interfaces, imports) live inside these files but are currently never extracted.

The proposed convention is `name.<underlying-ext>.<engine-ext>` — the **middle extension carries the underlying language**, which lets one mechanism generalize to *every* language jCodeMunch already supports as the template body, with no per-language work.

## Motivating case

A repo of Jinja2 templates that render TypeScript:

```jinja
import { Helper } from "./helper";

{% macro greeting(name) %}{{ name }}{% endmacro %}

export interface User {
  id: number;
  name: string;
}

export function buildUser(id: number): User {
  const label = "{{ default_label }}";
  return { id, name: label };
}
```

Indexing this `user.ts.j2` should yield: the `User` interface, the `buildUser` function, the `./helper` import — and (nice-to-have) the `{% macro greeting %}` definition.

## Current behavior (dropped at discovery)

The file is silently skipped:

- `get_language_for_path()` (`src/jcodemunch_mcp/parser/languages.py`) resolves compound extensions from the **first dot** in the basename, so `foo.ts.j2` looks up `.ts.j2`, which isn't registered → returns `None`.
- `_should_index_file()` (`src/jcodemunch_mcp/tools/index_folder.py`) then skips the file as `wrong_extension`. None of the inner TypeScript symbols are indexed.

## Proposed solution

The codebase already contains every primitive needed; this is an additive composition of existing patterns rather than new infrastructure:

1. **Recognition** — strip the engine extension and recursively resolve the remaining name's language (`foo.ts.j2` → `.j2` stripped → resolve `foo.ts` → `typescript`). A bare template with no inner language extension (`report.j2`) has nothing to parse as source and is skipped.
2. **Offset-preserving masking** — mask the engine constructs (`{{ }}`, `{% %}`, `{# #}`) by replacing each character with a space and preserving newlines, mirroring `mask_html_comments_keep_offsets()` in `parser/astro_shared.py`. Because offsets and line numbers are preserved, the re-parse positions point at the real lines of the template file — **no block-offset rewrapping is needed** (simpler than the Astro/Razor mixed-language parsers).
3. **Recursive delegation** — re-parse the masked text as the underlying language via `parse_file()`, the same two-tier dispatch the Razor/Astro parsers already use. Underlying symbols keep their sub-language tag (`typescript`); the file's positions are correct.
4. **Engine-pluggable** — a small `TemplateEngine` registry keyed by engine so adding another engine is a few lines (delimiter pattern + extensions).

Value-position holes (`const x = {{ v }}` → `const x =      `) are recovered by tree-sitter's error tolerance; a hole at a *name* position (`function {{ name }}()`) erases that one symbol — the same best-effort contract dbt-Jinja SQL parsing already accepts.

## Scope (proposed)

- **Engines:** Jinja2 (`.j2`/`.jinja`/`.jinja2`), Twig (`.twig`), Liquid (`.liquid`), Handlebars (`.hbs`/`.handlebars`/`.mustache`). EJS already has its own parser. Registry is pluggable for more.
- **Bare templates** (`report.j2`, no underlying-language extension): skipped — no inferable body language.
- **Engine directives:** Jinja/Twig `{% macro %}` / `{% block %}` definitions are also surfaced as symbols (reusing the dbt directive-extraction approach). Liquid/Handlebars get underlying-language extraction now; their directive extraction is a follow-up.
- **Purely additive:** these files were skipped before, so indexing them only adds symbols — no regression to existing output, and no `INDEX_VERSION` bump (the stored schema is unchanged; a re-index picks up the now-recognized files).

## Relevant prior art

- **Astro support (PR #311, v1.108.23)** — the canonical "new mixed-language file type" checklist: extension mapping in `languages.py`, a custom `_parse_*_symbols` in `extractor.py` that recursively calls `parse_file()` on embedded code, an import extractor in `imports.py`, a `*_shared.py` offset helper, and fixtures + regression tests. This proposal follows the same checklist.
- **Razor/Blazor support** — same recursive-delegation strategy for `.cshtml`/`.razor` (frontmatter/`@functions` → C#, inline `<script>` → JS), with offset-corrected re-wrapping.
- **dbt Jinja handling** — `parser/sql_preprocessor.py` already strips `{{ }}`/`{% %}`/`{# %}` and extracts `{% macro %}`/`{% test %}` directives as symbols for dbt SQL. The directive-extraction structure generalizes directly to generic Jinja/Twig macros/blocks.
- **`mask_html_comments_keep_offsets()`** (`parser/astro_shared.py`) — the exact offset-preserving masking primitive to mirror.
- **`extra_extensions`** config plumbing (`parser/languages.py::_apply_extra_extensions`) — precedent for additive extension→language registration.

## Environment

- jcodemunch-mcp 1.108.59.
- Use case: a repository of Jinja2 templates that render TypeScript source (`*.ts.j2`); generalizes to any supported underlying language and to additional engines.
- No new runtime dependencies; tree-sitter handles the masked underlying source as it does any other file.

---

_A working implementation following this design is on a feature branch (offset-preserving mask + recursive `parse_file` delegation + a pluggable engine registry), with line-number-accurate extraction verified end-to-end on the motivating `*.ts.j2` case._

horknfbr:
Closes #336.

## Summary

Files of the form `name.<underlying-ext>.<engine-ext>` — e.g. a Jinja2 template of TypeScript (`foo.ts.j2`), Jinja of Python (`service.py.jinja`), or a Twig template of TS (`widget.ts.twig`) — were previously dropped at discovery as `wrong_extension`, so every real source symbol inside them went unindexed. They are now recognized: the engine constructs are **masked offset-preserving** and the body is re-parsed as its underlying language, so symbols, signatures, imports, and line/byte positions all resolve correctly.

Because the mask is offset-preserving (each engine hole → same-length filler, newlines kept), the underlying-language symbols come back with positions that already point at the real lines of the template file — no block re-wrapping. That makes this simpler than the Astro/Razor sub-block extractors. Purely additive (these files are skipped today), with **no `INDEX_VERSION` bump**.

## Requested changes from the #336 review

All four review asks are incorporated:

1. **Scoped the first cut to Jinja2 + Twig.** Engines: Jinja2 (`.j2`/`.jinja`/`.jinja2`) and Twig (`.twig`) — the engines whose `name.<lang>.<engine>` double-extension convention this targets. Liquid/Handlebars/Mustache (single-extension, HTML-bodied, no underlying extension to resolve) are dropped from the first cut; the `TemplateEngine` registry (`parser/template_shared.py`) stays pluggable so they can be added on demand.
2. **Guarded the discovery-layer change.** The "strip a known engine extension, recurse on the remainder" step now sits **after** the existing compound and last-extension checks in `get_language_for_path`, and fires only for registered engine extensions — so it can only ever turn `None` into a language and never re-resolve an existing compound extension. Includes non-collision tests for `.d.ts`, `.test.ts`, `.blade.php`, and a bare `report.j2` that correctly skips.
3. **Reused the dbt directive path.** `sql_preprocessor.extract_dbt_directives` is generalized to accept a `directive_keywords` set (default = the dbt macro/test/snapshot/materialization set, so that path is byte-identical); the template parser delegates to it with `("macro", "block")` and maps the returned rows to Symbols, rather than a parallel regex implementation.
4. **Followed the Astro checklist and kept scope tight.** Extension mapping, `_parse_*` recursive delegation, import extractor, `template_shared.py` offset helper, fixtures + regression tests. **No CI or workflow files are touched.**

## Implementation notes

- **Underlying language is inferred from the middle extension** (`foo.ts.j2` → `typescript`) by stripping the engine suffix and resolving the remainder recursively — so every language jCodeMunch already supports works as the body with no per-language work. EJS (`.ejs`) keeps its existing parser.
- **Offset-preserving mask** mirrors `astro_shared.mask_html_comments_keep_offsets` but fills each hole with an identifier placeholder (`{{ x }}` → same-length `_`), keeping value-position holes (`DEBUG = {{ x }}`) parseable in strict grammars — the offset-preserving form of dbt's `__jinja__`. A hole at a *name* position, or free template text emitted inside a block body, is best-effort (same contract as dbt SQL parsing).
- **`{% macro %}` / `{% block %}`** definitions are additionally surfaced as symbols; underlying body symbols keep their own sub-language tag.
- **`repo` is forwarded into the recursive body parse** so the underlying language honors per-project `.jcodemunch.jsonc` enable/disable gating (caught in review of this branch).

## Tests

- New `tests/test_templates.py` — offset preservation, macro/block extraction, BOM/CRLF, imports, non-collision guards (`.d.ts`/`.test.ts`/`.blade.php`/bare `report.j2`), and repo propagation — plus fixtures under `tests/fixtures/templates/`.
- Parser-layer sweep (templates + sql/dbt + astro/razor/import/language/extractor + plan_refactoring) all green.
- `CHANGELOG.md` and `LANGUAGE_SUPPORT.md` updated.

https://claude.ai/code/session_01AZzyK49SHjZvBHf2sV975k

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,51 @@ All notable changes to jcodemunch-mcp are documented here.
 
 ## [Unreleased]
 
+### Added
+
+- **Index templating-language files over a supported underlying source
+  language** (issue #336). Files of the form `name.<underlying-ext>.<engine-ext>`
+  — e.g. a Jinja2 template of TypeScript (`foo.ts.j2`), Jinja of Python
+  (`service.py.jinja`), or a Twig template of TS (`widget.ts.twig`) — were
+  previously dropped at discovery as `wrong_extension`, so every real source
+  symbol inside them went unindexed. They are now recognized, the engine
+  constructs are masked **offset-preserving**, and the body is re-parsed as its
+  underlying language so symbols, signatures, imports, and line/byte positions
+  all resolve correctly.
+  - **Engines:** Jinja2 (`.j2`/`.jinja`/`.jinja2`) and Twig (`.twig`) — the
+    first cut, scoped to the engines whose `name.<lang>.<engine>` double-extension
+    convention this feature targets. The masking layer is a pluggable
+    `TemplateEngine` registry (`parser/template_shared.py`), so adding a
+    single-extension HTML-bodied engine (Handlebars/Liquid/Mustache) later is a
+    few lines. EJS (`.ejs`) keeps its existing parser.
+  - **Underlying language is inferred from the middle extension**, so *every*
+    language jCodeMunch already supports works as the template body with no
+    per-language work — the engine extension is stripped and the remaining name
+    is resolved recursively (`foo.ts.j2` → `typescript`). The resolution step
+    runs *after* the existing compound and last-extension checks in
+    `get_language_for_path`, so it can only ever turn an unresolved path into a
+    language — never re-resolve a `.d.ts` / `.blade.php` that already matched.
+  - **Jinja/Twig `{% macro %}` / `{% block %}` definitions** are additionally
+    surfaced as symbols, reusing the dbt directive path
+    (`sql_preprocessor.extract_dbt_directives`, generalized to accept a directive
+    keyword set) rather than a parallel implementation. The underlying body
+    symbols keep their own sub-language tag.
+  - **Offset-preserving mask** mirrors `astro_shared.mask_html_comments_keep_offsets`
+    but fills each hole with an identifier placeholder (`{{ x }}` →
+    same-length `_`), keeping value-position holes (`DEBUG = {{ x }}`) parseable
+    in strict grammars — the offset-preserving form of dbt's `__jinja__`. A hole
+    at a *name* position, or free template text emitted inside a block body, is
+    best-effort (same contract as dbt SQL parsing).
+  - **Bare templates** with no underlying-language extension (`report.j2`) are
+    skipped — there is nothing to parse as source.
+  - Purely additive: previously-skipped files only add symbols, with no
+    `INDEX_VERSION` bump (a re-index picks up the now-recognized files).
+  - New `tests/test_templates.py` (16 cases, incl. non-collision guards for
+    `.d.ts` / `.test.ts` / `.blade.php` / bare `report.j2`) + fixtures under
+    `tests/fixtures/templates/`. Implemented via offset-preserving masking +
+    recursive `parse_file()` delegation (the Razor/Astro pattern), and the
+    `get_language_for_path` resolution branch.
+
 ## [1.108.61] - 2026-06-18 - Every config key is now documented (no more "Other")
 
 ### Fixed
--- a/LANGUAGE_SUPPORT.md
+++ b/LANGUAGE_SUPPORT.md
@@ -81,6 +81,31 @@ These languages are fully indexed and searchable via `search_text`. Symbol extra
 | -------- | -------------- | ------------------------------------------------------------------ |
 | TOML     | `.toml`        | Tables indexed; key-as-symbol extractor planned                    |
 
+### Templating engines (over an underlying language)
+
+A template file named `name.<underlying-ext>.<engine-ext>` is indexed by masking
+the engine's constructs (offset-preserving) and re-parsing the body as its
+underlying language — so a Jinja2 template of TypeScript (`foo.ts.j2`) yields the
+real TypeScript symbols, with correct line/byte positions. The **underlying
+language is inferred from the middle extension**, so any language above works as
+the body. A bare template with no underlying extension (`report.j2`) is skipped.
+
+| Engine     | Extensions                          | Notes                                                                 |
+| ---------- | ----------------------------------- | --------------------------------------------------------------------- |
+| Jinja2     | `.j2`, `.jinja`, `.jinja2`          | `{% macro %}` / `{% block %}` also surfaced as symbols                 |
+| Twig       | `.twig`                             | Shares Jinja delimiters; macro/block extraction applies               |
+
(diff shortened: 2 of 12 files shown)


jgravelle:
Merged and shipped in **v1.108.62** (live on PyPI now).

Scott, this is exactly the implementation the issue promised. You hit all four review asks cleanly: scoped to Jinja2 + Twig with the registry left pluggable, guarded the `get_language_for_path` fallback so it can only ever upgrade an unresolved path (with the `.d.ts` / `.test.ts` / `.blade.php` / bare `report.j2` non-collision tests to prove it), reused `extract_dbt_directives` via a generalized keyword set rather than a parallel regex, and followed the Astro checklist with no CI or workflow files touched. I pulled the branch and ran it: 17/17 template tests plus the parser-layer sweep all green, and the dbt path stayed byte-identical. The `repo` forwarding into the recursive body parse was a nice catch on your own review.

The offset-preserving mask is the part I most wanted to keep, and you kept it: because the engine holes are filled in place, the re-parsed body carries correct line and byte positions with no block re-wrapping, which makes this genuinely simpler than the Astro and Razor parsers rather than another copy of them.

If this is solving a real pain in your template repos, a mention to others evaluating jCodeMunch is always welcome.

 -jjg

PR #361, opened 2026-07-06
feat: advertise MCP readOnlyHint on every tool so plan-mode clients run read tools silently

thomasmodeneis:
### Problem

MCP clients that gate tool execution (notably Claude Code's plan mode) prompt for
approval on every tool they cannot prove is read-only. jcodemunch's tools carry
no `annotations`, so `tools/list` returns `annotations: null` for all of them and
plan mode asks for approval on every jcm call (`search_symbols`,
`get_symbol_source`, `search_text`, ...). jcm is read-only by charter, so these
prompts are pure friction.

TLDR: no annotation -> client cannot prove read-only -> prompts on every call /
fix: set `readOnlyHint` per tool / upside: silent read tools in plan mode.

### Change

Annotate tools at ONE choke point instead of editing ~87 `Tool(...)` literals.
Every tool gets `ToolAnnotations(readOnlyHint=...)`: read tools `True`,
state-changing tools `False`. The write-set reuses the project's own authoritative
`counter.STATE_CHANGING_ACTIONS` so it cannot drift.

Files:

- `src/jcodemunch_mcp/server.py`
  - Add `ToolAnnotations` to the `mcp.types` import.
  - Add module-level `_NON_READONLY_TOOLS = counter.STATE_CHANGING_ACTIONS | {"index_dependency", "order", "route"}`.
  - Add `_apply_readonly_annotations(tools)` — for each tool without existing
    annotations, attach `ToolAnnotations(readOnlyHint = name not in _NON_READONLY_TOOLS)`
    via a non-mutating `model_copy`.
  - Apply it at both return points of `_build_tools_list()` (the main non-counter
    path and the `counter`-surface early return) so annotations are consistent on
    every surface.
- `tests/test_readonly_annotations.py` (new) — 4 tests:
  - the write-set covers `counter.STATE_CHANGING_ACTIONS | {index_dependency}` (drift guard),
  - every surfaced tool has a non-null `readOnlyHint`,
  - each tool's `readOnlyHint` equals `name not in _NON_READONLY_TOOLS`,
  - representative read tools (`get_symbol_source`, `search_symbols`, `search_text`, `plan_turn`) are `True` and write tools (`index_folder`, `register_edit`, `index_dependency`, `set_tool_tier`) are `False`.

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -16,7 +16,7 @@ from pathlib import Path
 from typing import IO, Any, Optional
 
 from mcp.server import Server
-from mcp.types import Tool, TextContent, Resource, Prompt, PromptMessage, GetPromptResult, CallToolResult
+from mcp.types import Tool, ToolAnnotations, TextContent, Resource, Prompt, PromptMessage, GetPromptResult, CallToolResult
 
 from . import __version__
 from . import config as config_module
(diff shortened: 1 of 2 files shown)


jgravelle (approved):
Thanks for this, and especially for deriving the write-set from `counter.STATE_CHANGING_ACTIONS` instead of hand-listing 87 tools, plus the drift-guard test that fails if that set grows. That is exactly the right instinct, and the single-choke-point approach in `_build_tools_list` keeps it clean.

Nice catch on `index_dependency` too. You found a real gap: that tool is also absent from the counter's `order` state-change gate (`counter.py`), so today `order("index_dependency", ...)` runs a write tool without the `allow_state_change=true` opt-in. Your annotation handles it correctly here; I will follow up by adding it to `STATE_CHANGING_ACTIONS` at the source so both the gate and the annotations derive from one list.

Merging. If jCodeMunch is useful to you, a word to others is always appreciated.

 -jjg

PR #362, opened 2026-07-07
feat: add Svelte (.svelte) language support

thomasmodeneis:
Svelte single-file components were dropped at discovery as wrong_extension, so a SvelteKit codebase was invisible to every jcodemunch tool. The infrastructure was already ~60% wired (.svelte in the JS-family extension set, the template/flow-edge sets, the hook nudge, $lib/.svelte-kit alias resolution); the gap was the parser layer.

Svelte SFCs share the Vue SFC shape (<script> + <style> + markup) and the bundled tree-sitter svelte grammar produces the same node structure as Vue, so extraction mirrors the tree-sitter Vue parser (_parse_vue_symbols) rather than the regex-based Astro path. New _parse_svelte_symbols re-parses each <script> block's raw_text with the JS/TS grammar and surfaces a synthetic component symbol (kind=class), top-level functions/classes/interfaces/types, Svelte 5 runes (let x = $state(...) / $derived(...), including the $derived.by member form and destructured let { a, b } = $props()), Svelte 4 props (export let / export const), and Svelte 4 reactive labels ($: doubled = ...) — all surfaced as kind=constant. Both an instance <script> and a module <script context="module"> block are parsed, each with its own line offset. Symbols carry language="svelte", matching Vue's "vue" convention.

Registration: ".svelte" -> "svelte" in LANGUAGE_EXTENSIONS, a new SVELTE_SPEC + LANGUAGE_REGISTRY entry (auto-flows into the MCP language enum and config validation). New _extract_svelte_imports mirrors _extract_astro_imports (whole-file ESM scan plus PascalCase component-usage edges over the HTML-comment-masked body, deduped).

Secondary parity: svelte added to the JS-family tuples in package_registry.py (cross-repo specifiers) and to the import/decl patterns and import-statement generator in plan_refactoring.py. No svelte_shared.py — reuses astro_shared.mask_html_comments_keep_offsets and keeps the rest inline, matching the Vue parser's shape.

Purely additive: no INDEX_VERSION bump, since previously-skipped .svelte files simply gain symbols on the next index, the same as the Astro and template-language releases.

New tests/test_svelte.py (13) + tests/fixtures/svelte/. Docs: LANGUAGE_SUPPORT.md Svelte row + valid-name entry


jgravelle (approved):
Thanks for this, Thomas. Genuinely one of the cleanest language-support PRs the project has taken: you found that the infrastructure was already ~60% wired and closed the exact remaining gap (the parser layer) instead of rebuilding around it, and you mirrored the tree-sitter Vue path (`_parse_vue_symbols`) rather than the regex Astro one because the bundled `svelte` grammar really does produce the same `document -> script_element -> raw_text` shape. That's the right call.

Reviewed and verified locally against current `main` (v1.108.108):

- `svelte` tree-sitter grammar confirmed bundled; `_parse_svelte_symbols` re-parses each `<script>` block with the JS/TS grammar and offsets lines correctly.
- Instance + module (`context="module"`) blocks both parsed, each with its own line offset.
- Svelte 5 runes (`$state`/`$derived`, incl. `$derived.by` member form and destructured `$props()`), Svelte 4 `export let`/`export const` props, and `$:` reactive labels all surface as `kind=constant`, parented to the synthetic component symbol.
- Import extraction correctly strips `<script>`/`<style>` bodies before the PascalCase tag scan, so TS generics (`identity<T>`, `Array<Item>`) don't get misread as component tags. Nice catch on that failure mode, and the comment-masking test covers the `<Ghost />`-in-a-comment case.
- Secondary parity (package_registry JS-family tuple, plan_refactoring patterns, LANGUAGE_SUPPORT.md) all consistent.
- No INDEX_VERSION bump: correct. Previously-skipped `.svelte` files just gain symbols on the next index, same contract as the Astro and template-language releases.

Test results: your `tests/test_svelte.py` passes 14/14; full suite green at 5137 passed, 7 skipped. Scope reconciled against the diff, no hidden edits.

Merging. If jCodeMunch is helping your SvelteKit work, a star or a word to other Svelte folks goes a long way for a small project. Thanks again.

-jjg

PR #363, opened 2026-07-10
docs: add Autohand Code setup

igorcosta:
## What changed

Added Autohand Code to the supported-client summary and the tested-configuration table with a direct stdio setup command.

## Why

jCodeMunch already supports any MCP-compatible client and documents many concrete agent/IDE configurations. This gives Autohand Code users a copy-pasteable entry that uses the existing `uvx jcodemunch-mcp` launch path.

## Validation

- Verified the command against the current Autohand Code MCP CLI implementation
- Ran `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 > **335B+ tokens saved** · **48,000+ developers** · **$1.69M+ in AI spend avoided** · **40,000+ kg CO₂ prevented**
 > Live telemetry at **[jcodemunch.com](https://jcodemunch.com/)** — benchmark: **95% average token reduction** (15 tasks / 3 repos, 99.8% peak).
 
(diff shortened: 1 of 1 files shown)


jgravelle (approved):
Thanks, Igor. This is exactly the kind of contribution that's easy to say yes to: you didn't just drop a name in the list, you gave Autohand Code users a copy-pasteable command with the right project-scope flag and a source link.

I verified it against the code-cli source before merging. 'autohand mcp add <name> <target> [args...]' is a real top-level command, and '-s, --scope <scope>' accepts user|project, so the placement before the server name is correct. Nice and accurate.

Merging now. If you're an Autohand Code user yourself, I'd love it if you spread the word about jCodeMunch to others in that community.

PR #378, opened 2026-07-24
feat: Add TOML symbol extraction support

zuoYu-zzz:
- Implement _parse_toml_symbols function in extractor.py
- Extract tables as type symbols, array tables as class symbols, and key-value pairs as constant symbols
- Add TOML language routing in parse_file
- Update TOML_SPEC comment in languages.py
- Add comprehensive tests for TOML parsing
- Update LANGUAGE_SUPPORT.md documentation

### DIFF
--- a/LANGUAGE_SUPPORT.md
+++ b/LANGUAGE_SUPPORT.md
@@ -50,6 +50,7 @@
 | SCSS              | `.scss`                                         | tree-sitter-scss + custom walker | function (`@mixin`/`@function`/`@include`), class (selectors/`%placeholder`), type (`@media`/`@supports`), constant (`$variable`) | — | `//` and `/* */` comments | Full SCSS extraction including variables and nested rules |
 | SASS              | `.sass`                                         | text search only (no grammar)   | — (files indexed for text search) | — | — | Indented SASS syntax; no tree-sitter-sass grammar in language-pack; falls back to CSS parser which cannot handle indented syntax → no symbols emitted |
 | YAML              | `.yaml`, `.yml`                                 | custom dict walker (pyyaml)     | function/type/constant (structural keys and containers extracted by depth/shape) | — | — | Generic YAML; Ansible-specific YAML detected via path heuristics and routed to the Ansible parser instead |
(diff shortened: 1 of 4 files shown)


jgravelle:
Merged, and thanks for this. The extractor is sound and the PR is well scoped.

A few things you got right that are worth calling out:

- The `[[array table]]` handling. We went looking for a bug there on purpose: three repeated `[[bin]]` entries should collide on symbol ID and get silently eaten by the store's `INSERT OR REPLACE`. The existing dedupe pass already handles it. All three survive with correct line numbers, and the PR needed nothing extra.
- `LANGUAGE_SUPPORT.md` updated in both places, added to the supported table and removed from the text-search-only table. Contributors miss that second edit routinely.
- `TOML_SPEC` correctly keeps `symbol_node_types={}`, so nothing double-extracts.

One follow-up we pushed ourselves in commit `f0eda7b`, which is why we merged instead of sending it back. It was small enough that fixing it directly beat a review round trip. `_extract_key` scanned a `dotted_key` node's direct children for `bare_key`/`quoted_key`, but tree-sitter-toml nests `dotted_key` left-recursively, so `[tool.ruff.lint]` parses as `dotted_key(dotted_key(tool, ruff), lint)`. The inner node was not in the filter tuple, so every segment except the last got dropped. Two-level paths worked, which is exactly why it looked right at a glance. Three-plus levels broke. On this repo's own `pyproject.toml`, `[tool.hatch.build.targets.wheel]` came back as `wheel` with signature `[wheel]`, a header that appears nowhere in the file. The fix returns path segments and recurses on the nested node. Building from segments also let us set `name` to the leaf and `qualified_name` to the full path, matching every other extractor in that module.

The new test asserts three-deep and five-deep tables and is proven non-vacuous (restoring the one-token bug flips it to FAIL). Worth saying plainly: this was invisible to the PR's own tests because `TOML_SOURCE` uses only single-segment headers, so it isn't something review would obviously have caught either.

It rides the next release. If jCodeMunch is useful to you, we'd appreciate you spreading the word.

 -jjg

Issue #366, opened 2026-07-12
Index max files cap seems to be silent (and a bit low for mono repos)


oderwat:
While verifying the .mjs fix from [#365](https://github.com/jgravelle/jcodemunch-mcp/issues/365) on a repo with ~2,150 indexable files, we noticed the index contained exactly 2,000 files. The default max_folder_files=2000 had silently dropped ~150 files - entire subdirectories were missing from search_text / search_symbols results, and nothing in the index result or in query responses indicated truncation. The index looked healthy while quietly missing files; we only caught it because a file we knew existed didn't show up.

Raising max_folder_files in config.jsonc fixed it, but the failure mode is nasty for anyone who doesn't know their exact file count.

Suggestion: when the file walk hits max_folder_files, surface it explicitly - e.g. "truncated": true plus a skipped-file count and a hint to raise max_folder_files in the index result, and ideally a warning in the _meta envelope of queries against a truncated index.

P.S.: Above was AI generated :) We wondered for a while why certain packages were often missed, and we had to manually point the agent there. It seems that they were just not in the index. The repo is a monorepo and the files are actual code, no "import / build directory" bloat.

jgravelle:
Thanks for this, and especially for the P.S. context. "The index looked healthy while quietly missing files" is exactly the failure mode worth fixing, and knowing you'd spent time wondering why certain packages kept getting missed made the priority obvious.

Fixed in v1.108.126 (just published). The cap now announces itself instead of dropping files in silence:

- The index result carries `truncated: true` plus `files_discovered` / `files_indexed` / `files_skipped_cap` and a warning naming `max_folder_files`. This now fires on the incremental re-index path too, not just a fresh index (the re-index path is the one you'd have hit verifying the .mjs fix, and it was silent before).
- The status is persisted in the index, so `resolve_repo` reports it up front, and `search_text` / `search_symbols` attach `_meta.index_truncated` with a note that a missing or thin result may be truncation rather than genuine absence. It self-heals: re-index under a raised cap and the flag clears.

On "a bit low for mono repos": I left the default at 2000 on purpose. Raising it silently would just trade your surprise for a different one (a much larger, slower index someone didn't ask for). Now that truncation is loud, the fix is a one-liner once you know you need it, in `config.jsonc`:

    "max_folder_files": 20000,

[...]

PR #379, opened 2026-07-24
Gleam import extraction and module-path resolution

oderwat:
# feat(parser): Gleam import extraction and module-path resolution

## Problem

For Gleam codebases, every import-graph-based tool silently returns wrong
answers instead of "unsupported":

- `get_call_hierarchy` reports **0 callers for every Gleam symbol**, with
  `methodology: "ast_call_references"` and `confidence_level: "medium"`
- `get_impact_preview` reports "nothing affected" for every Gleam symbol
- `find_importers` returns empty for every `.gleam` file
- dead-code analysis marks virtually every Gleam module as unused

For a "can I delete this?" question the answer is a confident, wrong "yes".

Root cause: `_LANGUAGE_EXTRACTORS` in `parser/imports.py` has no `"gleam"`
entry, so `extract_imports()` returns `[]` for every Gleam file and the
import graph stays empty. (`index_folder` does surface this as
`missing_extractors: ["gleam", ...]`, but the query tools do not.)

## Fix

Follows the existing per-language extractor pattern (regex over content,
same shape as the Haskell/Dart/Swift extractors):

- **`_extract_gleam_imports`**: handles plain imports, `.{...}` unqualified
  lists (including `type X` entries and `X as Y` aliases, reusing
  `_clean_names`), module aliases (`import holmes_msg as msg`), deep module
  paths, and multi-line `.{...}` lists as emitted by `gleam format`.
- **`_gleam_source_roots`**: Gleam module paths are relative to a package's
  `src/`, `test/` or `dev/` directory. Roots are derived structurally from
  indexed `.gleam` paths (every prefix ending in a `src`/`test`/`dev`
  segment) plus `<dir>/src|test|dev` for every indexed `gleam.toml`.
  Cached by `source_files` identity, mirroring `_python_source_roots`.
- **`resolve_specifier`**: for `.gleam` importers, slash-style specifiers
  resolve as `<root>/<module_path>.gleam` against those roots, preferring
  the importer's own package (module paths are only unique per package).
  Stdlib and hex-dependency imports (`gleam/io`, `lustre/...`) resolve to
  nothing, so no edge is created.

No API changes; other languages are untouched. `gleam` disappears from
`missing_extractors`.

## Tests

`tests/test_gleam_imports.py` (14 tests): extraction of all syntax forms,
source-root detection (structural, gleam.toml-derived, root-level package),
resolution (cross-package, same-package, test-to-src, test-only helpers,
dev-to-src, stdlib negatives), non-gleam importers unaffected, and the
`index_folder` / `missing_extractors` integration.

## Verification against a real monorepo

Measured on a Gleam monorepo with 1412 `.gleam` files across ~90 packages
(8756 raw import lines):

- extraction: 8741 edges, **zero real imports missed** (the small gap vs
  raw line count is per-file dedup, plus import lines inside multi-line
  string literals of one code-generator file)
- resolution: every unresolved specifier is a genuine external dependency
  (`gleam/*` stdlib, hex packages); **zero internal imports unresolved**
- **1 wrong edge out of 8741**, inside vendored compiler-test fixtures
  where two unrelated fixture packages define the same module name
- end to end: a widely used helper (`get_list_block`) went from 0 reported
  callers to 24 callers across 9 files; `find_importers` on its file
  matches full-text-search ground truth exactly (31/31 files)

## Known limitations (shared with the other regex extractors)

- An `import ...` at line start inside a multi-line string literal is
  extracted as an edge (regex extractors do not see string boundaries).
- Cross-package module-name collisions are disambiguated by preferring the
  importer's own package, not by parsing `gleam.toml` dependencies. In the
  1412-file corpus this produced exactly one wrong edge, in compiler-test
  fixtures.

Erlang (no import statements; dependencies are `mod:fun()` call sites) is
deliberately out of scope for this PR.

### DIFF
--- a/src/jcodemunch_mcp/parser/imports.py
+++ b/src/jcodemunch_mcp/parser/imports.py
@@ -937,14 +937,15 @@ def _gleam_source_roots(source_files) -> tuple[str, ...]:
     """Detect Gleam source roots from the indexed file set.
 
     A Gleam module path like ``webse/config_types`` is relative to a
-    package's ``src/`` or ``test/`` directory, so those directories are the
-    source roots. They are derived structurally: every path prefix of an
-    indexed ``.gleam`` file that ends in a ``src`` or ``test`` segment
-    (``cells/webse/src/webse/config.gleam`` -> ``cells/webse/src``), plus
-    ``<dir>/src`` and ``<dir>/test`` for every indexed ``gleam.toml``. The
-    two signals overlap for normal packages; the gleam.toml one also covers
-    packages whose files did not make it into the index, and the structural
-    one covers indexes that exclude toml files.
+    package's ``src/``, ``test/`` or ``dev/`` directory, so those
+    directories are the source roots. They are derived structurally: every
+    path prefix of an indexed ``.gleam`` file that ends in a ``src``,
+    ``test`` or ``dev`` segment (``cells/webse/src/webse/config.gleam`` ->
+    ``cells/webse/src``), plus ``<dir>/src``, ``<dir>/test`` and
+    ``<dir>/dev`` for every indexed ``gleam.toml``. The two signals overlap
+    for normal packages; the gleam.toml one also covers packages whose
+    files did not make it into the index, and the structural one covers
+    indexes that exclude toml files.
     """
     cache_key = source_files if isinstance(source_files, frozenset) else frozenset(source_files)
     cached = _gleam_roots_cache.get(cache_key)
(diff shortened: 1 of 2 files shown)


jgravelle (approved):
Thank you for the negative cases. `test_no_false_positives_from_non_import_lines` and `test_non_gleam_importer_unaffected` are the two tests that make this reviewable quickly: an import-extraction regex is easy to write and hard to trust, and you proved the boundary in both directions rather than only the happy path. The multiline unqualified list and the `type` prefix stripping are the shapes that would have bitten later.

Verified before merging:

- 22 tests pass on the branch, and again against a trial merge onto current `main` (which is four commits ahead at v1.108.170).
- No conflicts. `main` has not touched `imports.py` since well before this branched, so no rebase was needed.
- 210 tests green across the neighbouring import and language suites (`test_dart_imports`, `test_find_importers`, `test_dependency_graph_imports_alias`, `test_languages`) on the merged state.
- ruff clean.
- Diff is two files and adds nothing beyond regex plus pure functions -- no subprocess, filesystem, or dynamic execution.

Gleam was already registered in `LANGUAGE_REGISTRY`, so symbols extracted but the import graph stayed empty for it. That gap is what made `find_importers`, `get_blast_radius`, and `get_dependency_graph` blind on Gleam projects, and it is now closed. The `dev/` source-root follow-up commit was the right catch.

Merged. Thanks again -- this is the third useful thing you have sent this repo, and #366 (the silent `max_folder_files` truncation) is still one of the better bug reports it has received.

If jCodeMunch is useful to you, mentioning it to someone else is the best thing you can do for the project.

 -jjg

PR #380, opened 2026-07-25
Add Atlas Cloud summarizer provider

binyangzhu000-sudo:
## Summary
- add an `atlascloud` AI summarizer provider backed by the existing OpenAI-compatible request path
- support `atlascloud`, `atlas-cloud`, and `atlas` provider aliases plus `ATLASCLOUD_*` / `ATLAS_CLOUD_*` key, base URL, and model env fallbacks
- keep hosted-provider auto-detection behind the existing paid summaries opt-in, and include Atlas Cloud in config diagnostics, keyring env resolution, and test coverage

## Validation
- `uv run pytest tests/test_summarizer.py tests/test_v1_108_128.py tests/test_credentials.py -q` (97 passed)
- `uv run ruff check src/jcodemunch_mcp/summarizer/batch_summarize.py src/jcodemunch_mcp/server.py src/jcodemunch_mcp/config.py src/jcodemunch_mcp/credentials.py src/jcodemunch_mcp/tools/test_summarizer.py tests/test_summarizer.py tests/test_v1_108_128.py tests/test_credentials.py`
- `uv run python -m compileall -q src/jcodemunch_mcp/summarizer/batch_summarize.py src/jcodemunch_mcp/server.py src/jcodemunch_mcp/config.py src/jcodemunch_mcp/credentials.py src/jcodemunch_mcp/tools/test_summarizer.py`
- `git diff --check`
- Atlas live model catalog check returned 121 models and included `qwen/qwen3.5-flash` plus `deepseek-ai/deepseek-v4-pro`

No README, logo, sponsor, credits, or partner copy changes are included.


jgravelle:
Thanks for the careful work on this, @binyangzhu000-sudo. Two things stand out: you added `atlascloud` to `_PAID_CLOUD_PROVIDERS` rather than routing around it, so the provider stays behind the paid-summaries opt-in and can't silently bill from a bare env key. That guard exists because of a real report, and honoring it was the right call. Tests are in the PR too, not bolted on later.

One procedural note first: the CLA is unsigned, so this branch can't be merged as-is regardless of anything below.

On substance, we're closing this because the capability already works today with no new code. Atlas Cloud exposes an OpenAI-compatible endpoint, and jCodeMunch already speaks that. For anyone who lands here wanting to use Atlas Cloud right now:

```
JCODEMUNCH_SUMMARIZER_PROVIDER=openai
OPENAI_API_BASE=https://api.atlascloud.ai/v1
OPENAI_API_KEY=<key>
OPENAI_MODEL=qwen/qwen3.5-flash
```

What a merged branded provider adds is permanent configuration surface: 8 environment variable spellings (`ATLASCLOUD_` and `ATLAS_CLOUD_` prefixes, each across `API_KEY` / `API_BASE` / `BASE_URL` / `MODEL`) plus 3 provider aliases (`atlas`, `atlas-cloud`, `atlascloud`). We're on a 1.x line with a no-removal contract for configuration, so every one of those names is ours forever once it ships.

To be fair about the bar we're applying: we have merged branded aliases of exactly this shape before, MiniMax, GLM, and OpenRouter among them. So this isn't "we don't accept branded providers." The difference is demand. Those went in against actual user requests, and we gate this class of permanent surface on someone asking for it, the same way we gate platform installers.

So if a jCodeMunch user wants Atlas Cloud specifically, reopen this PR or open an issue pointing at that and it goes in. Your work is done and wouldn't need redoing.

If jCodeMunch is useful to you, we'd appreciate you telling people about it.

 -jjg

PR #387, opened 2026-07-26
docs: quote pip extras install examples

nyxst4ck:
## Summary
- Quote `pip install` examples that include extras.

## Why
- Shells such as zsh can expand unquoted bracket expressions before `pip` receives them.

## Scope
- Files changed:
- `CONTEXT_PROVIDERS.md`
- `README.md`
- This does not change dependencies or runtime behavior.

## Testing
- `git diff --check`

## Maintainer Fit
- Small install-command correctness fix.
- Duplicate PR search terms checked: `quote pip extras`, `quote optional dependency`, `zsh pip install extras`, `pip extras install examples`

### DIFF
--- a/CONTEXT_PROVIDERS.md
+++ b/CONTEXT_PROVIDERS.md
@@ -124,7 +124,7 @@ When the dbt provider is active, `index_folder` returns enrichment stats:
 The dbt provider requires `pyyaml` for schema.yml parsing:
 
 ```bash
-pip install jcodemunch-mcp[dbt]
+pip install 'jcodemunch-mcp[dbt]'
 ```
 
 Without PyYAML, doc blocks are still parsed but model/column metadata from YAML files is skipped.
--- a/README.md
+++ b/README.md
@@ -662,7 +662,7 @@ See **[speedreview/README.md](speedreview/README.md)** for full setup and config
 Ask any question about any codebase. Get an answer in under 3 seconds.
 
 ```bash
-pip install jcodemunch-mcp[groq]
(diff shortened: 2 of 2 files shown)


jgravelle:
Thanks — you were right, and it's fixed in `cc8a313` on `main`.

Closing this rather than asking you to sign a CLA for four lines of quoting. That's a real hurdle for a fix this size, and it would be a poor trade to make you clear it.

**Your diagnosis was correct and the scope was wider than the PR.** The same defect was in seven more places, and five of them are `ImportError` messages — the string a user is told to copy at the exact moment they're already blocked:

- `groq/explainer.py`, `groq/inference.py`, `groq/voice.py`, `summarizer/batch_summarize.py` (x2)
- `CONFIGURATION.md` (x2), `USER_GUIDE.md`, `CLAUDE.md` (x2)

Those five were the ones worth having. A broken command in the README costs someone a search; a broken command in the error message that fired *because* their install was incomplete costs them their patience.

**One deliberate difference from your patch: double quotes rather than single.** `QUICKSTART.md`, `TROUBLESHOOTING.md` and most of `README.md` were already using `"..."`, so single quotes would have introduced a third style across the docs. Double quotes also survive `cmd.exe`, where `'...'` is passed through literally — and a meaningful share of this project's users are on Windows. Same fix, one shell wider.

`CHANGELOG.md` is left as-is on purpose. It's a record of what was written at the time, not a surface anyone copies from.

You're credited in the commit. If you'd like the CLA on file for future work it's a one-click thing at the link the bot posts, but nothing here required it — and please do send the next one.

 -jjg

PR #388, opened 2026-07-27
fix: avoid duplicate indexing when auto-watch starts

### LINKED ISSUE #384: index_folder is missing from _AUTO_WATCH_EXCLUDED, so a watch-enabled call indexes the tree twice
Found while investigating #375; split out so that issue can close on the stall fix. Not caused by, and does not cause, the stall.

## What happens

`_AUTO_WATCH_EXCLUDED` (`server.py:4522`) lists tools that must not trigger auto-watch. `index_file` is in it, with the comment "path arg is a file path, not a folder; requires repo already indexed". **`index_folder` is not.**

So when `watch` is enabled, a call to `index_folder(path=X)` hits the guard at `server.py:4569`, falls through, and `_auto_watch_if_needed` awaits a full `ensure_indexed(X)` **before the tool dispatches**. The tool then indexes X again. One user-visible call, two full indexing passes over the same tree.

This is independent of `progressToken` and of the provider-regex stall. On a large repo it roughly doubles the cost of the first `index_folder` call.

## Why it has not been patched

Both obvious fixes have a real downside, which is why this is a filed issue rather than a commit:

1. **Drop the redundant `ensure_indexed` but keep `add_folder`.** The watch task's own initial index would then race the tool's index on the same `indexwrite` lock. Every acquire passes `wait_seconds=60.0`, so the loser waits up to a minute. That is plausibly worse than the duplicate work it replaces.
2. **Add `index_folder` to `_AUTO_WATCH_EXCLUDED`.** Simple and safe against the race, but it silently removes auto-start-watching for the single most natural way a user would expect to start watching a folder. That is a behaviour removal under the 1.x zero-surprise contract.

A third option, not yet explored: have `_auto_watch_if_needed` register the folder for watching without performing an initial index when the tool about to run will index it anyway.

## Repro sketch

Enable `watch`, call `index_folder` on a cold repo, and observe two full passes. Anyone confirming this on a large tree with timings would be helping; the cost is most obvious there.

Bortlesboat:
Closes #384.

## Summary

- register `index_folder`'s auto-watch task without a second initial index
- preserve the existing index-before-watch behavior for query tools and standby takeover
- hydrate the watcher's hash cache lazily after the delegated index completes

## Testing

- `python -m pytest tests/ -q` — 6164 passed, 10 skipped
- `python -m ruff check src/jcodemunch_mcp/server.py src/jcodemunch_mcp/watcher.py tests/test_watcher_dynamic.py`

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -4673,16 +4673,24 @@ async def _auto_watch_if_needed(name: str, arguments: dict, storage_path: Option
     # Opportunistic standby takeover before indexing
     maybe_takeover = getattr(_watcher_manager, "maybe_takeover", None)
     if maybe_takeover is not None:
-        result = await maybe_takeover(folder)
+        if name == "index_folder":
+            result = await maybe_takeover(folder, initial_index=False)
+        else:
+            result = await maybe_takeover(folder)
         if result.get("status") in {"started", "already_watched"}:
-            await _watcher_manager.ensure_indexed(folder)
+            if name != "index_folder":
+                await _watcher_manager.ensure_indexed(folder)
             return
 
     # Race-safe reindex, then start watching
     try:
-        await _watcher_manager.ensure_indexed(folder)
-        await _watcher_manager.add_folder(folder)
-        logger.debug("Auto-watch: indexed and watching %s", folder)
+        if name == "index_folder":
+            await _watcher_manager.add_folder(folder, initial_index=False)
+            logger.debug("Auto-watch: watching %s; index delegated to index_folder", folder)
+        else:
+            await _watcher_manager.ensure_indexed(folder)
+            await _watcher_manager.add_folder(folder)
+            logger.debug("Auto-watch: indexed and watching %s", folder)
     except Exception:
         logger.debug("Auto-watch failed for %s", folder, exc_info=True)
 
--- a/src/jcodemunch_mcp/watcher.py
+++ b/src/jcodemunch_mcp/watcher.py
@@ -244,8 +244,13 @@ async def _watch_single(
     on_reindex: Optional[Callable[[], None]] = None,
     quiet: bool = False,
     log_file_handle: Optional[IO] = None,
+    initial_index: bool = True,
 ) -> None:
-    """Watch a single folder and re-index on changes."""
+    """Watch a single folder and re-index on changes.
+
+    ``initial_index=False`` is for callers that register the watcher immediately
+    before performing the initial index themselves.
+    """
     _watcher_output(f"Watching {folder_path} (debounce={debounce_ms}ms)", quiet=quiet, log_file_handle=log_file_handle)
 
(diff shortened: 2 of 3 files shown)


jgravelle:
Closing this, and you are owed a straight account of why rather than a one-liner.

**We pre-empted you, and it was our fault.**

You opened this at 06:51 UTC on 2026-07-27. At 12:56 UTC the same day we shipped v1.108.189 with our own fix for #384 and closed the issue, having never looked at the open pull requests on a repo where an issue had one attached. The cross-reference was sitting on #384's timeline the whole time. That is a process failure on our side, not a judgment about your work, and it is the reason this PR is being closed instead of merged.

Because .189 rewrote the same three functions this PR touches, GitHub now reports it `CONFLICTING`. So the only path to merging it would be asking you to rebase onto a shipped version of your own fix and sign a CLA to do it. That is a lot to ask of someone we went around, so we did not ask.

**What we did instead: shipped the part you got right and we missed, with credit, in v1.108.190.**

You covered `maybe_takeover`. We did not. Our .189 fix threaded `skip_initial_index` through `add_folder` and stopped there, so when another process had been watching the folder and this one adopted it, the watch task still ran a full pass over the tree `index_folder` had just finished. The bug #384 reports survived for exactly that case, and it survived because we fixed one of the two paths you fixed. That is now in v1.108.190, `maybe_takeover` takes the flag, and the default is pinned to `False` by a test because the standby signal loop and the fallback retry loop both depend on it staying that way.

Your lazy hash-cache hydration is in as well, and it turned out to be the more interesting of the two. It fixes something older than either of our #384 fixes: the cache is only built on a *successful* initial index, so a failed one leaves it empty for the life of the watch task, and an empty cache does not raise. It silently drops every `old_hash`. Neither the issue nor our fix had noticed that. One change from your version, and I would rather explain it than let you find it: we guard on an explicit `_hash_cache_built` flag rather than `if not _hash_cache`, so a legitimately empty index does not re-read the store on every change.

**Where our approaches genuinely differed, for what it is worth.**

You registered the watch before dispatch and hydrated lazily. We deferred registration to after dispatch, in the dispatcher's `finally`. The deferral also closes a window your version leaves narrow but open: with the watcher live while `index_folder` is still running, a file edited mid-index can fire a debounced reindex into the tool's in-flight index on the same `indexwrite` lock. Your lazy hydration handles the cache but not that. Conversely you handled the takeover path and we did not. Both halves were right about different things, which is the useful outcome even though the process getting there was bad.

Credit is in the CHANGELOG, the v1.108.190 release notes, and on #384.

**Two practical notes.**

Your CLA is still unsigned, and it stayed a hard gate throughout, so nothing here turned on it. If you send anything else our way it is thirty seconds at the link CLA Assistant posts, and worth doing in advance.

If you disagree with any of the above, particularly the read that our deferral is safer than pre-dispatch registration, say so and reopen the argument. You found a real defect in a shipped release on your first contribution here, and independently reached the same design we did. That is not something we want to discourage on the strength of a scheduling accident.

 -jjg

Issue #241, opened 2026-04-14
Claude bypassing the jcodemunch rules


Bamieh:
I noticed this behavior happening inside claude, i am just using a plain installment of jcodemunch nothing custom with the installed hooks:

```
⏺ Update(file name)
  ⎿  File must be read first

  Read 1 file (ctrl+o to expand)
  ⎿  PreToolUse:Read hook returned blocking error                                                           ⎿  This is a 5,858-byte code file. Use jCodemunch for efficient navigation: get_file_outline to see the
                                                                                                                file structure, then get_symbol_source for specific symbols you need. Only fall back to Read when
                                                                                                               you need exact line numbers for Edit, or for non-code files.

⏺ I need to Read this file to Edit it. The hook says to fall back to Read for exact line numbers for Edit — that's what I'm doing.

  Read 1 file (ctrl+o to expand)
  ⎿  PreToolUse:Read hook returned blocking error                                                           ⎿  This is a 5,858-byte code file. Use jCodemunch for efficient navigation: get_file_outline to see the
                                                                                                                file structure, then get_symbol_source for specific symbols you need. Only fall back to Read when
                                                                                                               you need exact line numbers for Edit, or for non-code files.

⏺ The hook blocks Read on this code file entirely. I'll use Bash to insert the new function since Edit requires Read as a prerequisite.


⏺ Bash(sed -i '' '48a\
      \…)
  ⎿  (No output)

  Read 1 file (ctrl+o to expand)

⏺ Good. Now add the export to index.ts.

  Read 1 file (ctrl+o to expand)

⏺ Update(xx/index.ts)
```

jgravelle:
This is a version problem.

v1.21.27 shipped a **PreToolUse hook** that got a little too trigger-happy. It hard-blocked `Read` on code files over 4KB. Sounds fine in theory… except **Claude Code needs Read before Edit**, so you basically kneecapped the workflow.

Result:
Claude tries to Read → blocked
Retries → blocked again
Eventually says "fine, I'll do it myself" and starts hacking around it with `sed` via Bash

That's the circus you're watching.

We fixed it in **v1.22.5 (commit 4516640)**. The hook now does what it *should've* done all along:

* Prints a polite nudge to use jCodeMunch tools
* **Does NOT block Read anymore**
* Targeted reads (offset/limit) go through quietly like a professional

**Fix:**

```
pip install --upgrade jcodemunch-mcp
```

After that:
[...]

Issue #411, opened 2026-08-03
tests/test_v1_108_194.py: max_file_size tests read the developer's real config


rknighton:
## The one thing that is wrong

`_run_config()` scrubs the `JCODEMUNCH_MAX_FILE_SIZE` env var but not the config file, so
the subprocess reads the developer's real `~/.code-index/config.jsonc` and both
`max_file_size` tests fail on any machine where that key is set.

You identified this while reviewing
[#404](https://github.com/jgravelle/jcodemunch-mcp/issues/404#issuecomment-5165467060):

> Your two failing config tests are unrelated — they read the real
> `~/.code-index/config.jsonc` and fail on any box with a non-default `max_file_size`,
> which is worth its own issue if it keeps costing you a clean run.

This issue records it formally and adds a tested isolation patch.

## Reproduction

Put the documented escape hatch in `~/.code-index/config.jsonc`:

```jsonc
{
  "max_file_size": 2147483647
}
```

Then run the file:

```powershell
$env:PYTHONPATH = "src"
py -3 -m pytest tests/test_v1_108_194.py -q
```

```text
FAILED tests/test_v1_108_194.py::test_config_reports_max_file_size_default
FAILED tests/test_v1_108_194.py::test_config_attributes_an_env_override_to_env
2 failed, 9 passed
```

## Expected vs actual

Expected: the suite passes regardless of the developer's local configuration.

Actual: it passes only when the developer's global configuration does not currently contain
an explicit `max_file_size`.

## Environment

- jCodeMunch version: 1.108.234, at tag `v1.108.234`. Reproduced and patched there.
  Present since the file was added in v1.108.194.
- Python version: 3.13.7
- OS: Windows 11 26200. Not material; the trigger is a config file, not a platform.

### Patch

`CODE_INDEX_PATH` fully isolates config resolution. `tempfile` is already imported.
Verified with `git apply --check` at `v1.108.234`:

```diff
diff --git a/tests/test_v1_108_194.py b/tests/test_v1_108_194.py
index bb38485..f243b87 100644
--- a/tests/test_v1_108_194.py
+++ b/tests/test_v1_108_194.py
@@ -40,13 +40,17 @@ def _run_config(env_extra: dict[str, str] | None = None) -> str:
     env.pop("JCODEMUNCH_MAX_FILE_SIZE", None)
     if env_extra:
         env.update(env_extra)
-    proc = subprocess.run(
[...]

PR #420, opened 2026-08-07
fix(hooks): deliver agent-facing output on channels that reach the model

georgebashi:
## Summary

Several hook messages meant to steer the model were emitted on channels only the user can see. For a non-blocking hook that exits 0:

| Output | Recipient |
|---|---|
| stderr | transcript / debug log — not model context |
| top-level `systemMessage` | user |
| `hookSpecificOutput.additionalContext` | model |

(Exit 2 does feed stderr to the model, but it also blocks the call — not what an advisory nudge wants. `permissionDecisionReason` likewise reaches the model, which is why strict mode was never affected. See the [hooks reference](https://code.claude.com/docs/en/hooks).)

So the advisory Read/Grep nudges and the SubagentStart briefing were inert: the hooks fired, gated correctly, computed the right text, and none of it landed. The Read nudge has been in this state since v1.22.5 (4516640), which correctly replaced a hard `Read` deny with an exit-0 stderr warning to unbreak Read-before-Edit (#241) — and in moving off the deny channel, moved off the only channel that reached the model. The comment there read:

```python
# Stderr text is surfaced to the agent as guidance.
```

## Changes

- Route the advisory Read and Grep nudges through `additionalContext`.
- Route the SubagentStart repo briefing through `additionalContext` — it exists to orient the subagent, so the user-facing channel defeated its purpose.
- Keep TaskCompleted diagnostics on `systemMessage`: that event has no non-blocking model channel, and its only alternative — exit 2 — would refuse task completion over advisory findings.
- Add a SessionStart hook for `compact|resume|fork`. PreCompact can report its snapshot to the user but cannot inject it into model context, so the snapshot was computed at the moment state was about to be lost and then discarded. SessionStart restores it afterwards; both hooks share snapshot construction.
- Stay silent on `startup|clear`, empty snapshots, and missing journals — an unrelated session's journal would present stale files as current focus.
- Register SessionStart idempotently, including on existing installs. Duplicate detection is per-subcommand, so the legacy `"hook-p"` marker does not block the new event.

## Behavior preserved

Advisory output carries no `permissionDecision`, so calls proceed and Read-before-Edit keeps working; a regression test asserts this. Unchanged: the 4KB threshold, extension allowlist, indexed-repo checks, the `offset`/`limit` escape hatch, `JCODEMUNCH_HOOK_MIN_SIZE`, `JCODEMUNCH_HOOK_GREP_NUDGE`, `JCODEMUNCH_ENFORCE`, and the strict-mode deny.

## Tests

New `tests/test_hook_output_channels.py` pins delivery channels rather than message content, covering Read, Grep, SubagentStart, and SessionStart: silent paths silent on every channel, source scoping, error containment, shared snapshot construction, and install idempotency. Nine tests in `test_hooks.py` asserted `"search_text" in err` — they passed *because* the message went nowhere — and now assert the model-facing channel.

Full suite: 7,023 passed. Eight failures in `test_count_cap_end_to_end.py`, `test_install_pack.py`, `test_suggest_corrections.py`, and `test_v1_108_163.py` reproduce identically on unmodified `main` (network/env-dependent).

Manually verified in a live Claude Code session: the Read nudge now arrives as a system reminder, targeted reads stay silent, and `hook-sessionstart` injects only for continuing-session sources.

### DIFF
--- a/src/jcodemunch_mcp/cli/hooks.py
+++ b/src/jcodemunch_mcp/cli/hooks.py
@@ -8,6 +8,11 @@ PostToolUse — auto-reindex after Edit/Write to keep the index fresh.
 
 Both read JSON from stdin and write JSON to stdout per the Claude Code
 hooks specification.
+
+Output channel discipline: anything meant for the *model* goes out as
+``hookSpecificOutput.additionalContext`` (see ``_emit_additional_context``).
+Neither stderr-on-exit-0 nor top-level ``systemMessage`` reaches the model —
+both surface to the user — so a nudge written to either is silently inert.
 """
 
 import json
@@ -83,8 +88,8 @@ _MIN_SIZE_BYTES = int(os.environ.get("JCODEMUNCH_HOOK_MIN_SIZE", "4096"))
 def _enforce_mode() -> str:
     """jCodemunch enforcement tier for native file tools (``JCODEMUNCH_ENFORCE``).
 
-    * ``"advisory"`` (default): warn on stderr but **allow** — the v1.108.47
-      behavior. A hard deny here would break Read-before-Edit.
+    * ``"advisory"`` (default): nudge via ``additionalContext`` but **allow** —
+      the v1.108.47 behavior. A hard deny here would break Read-before-Edit.
     * ``"strict"``: **deny** a native Read/Grep that an indexed-repo jcm route
       can already serve. Targeted reads (``offset``/``limit``), tiny files, and
       paths outside every indexed repo still pass, so the escape hatch is always
@@ -106,14 +111,43 @@ def _enforce_mode() -> str:
 def _emit_pretooluse_deny(reason: str) -> int:
     """Emit a Claude Code PreToolUse ``deny`` decision (stdout JSON) and exit 0.
 
-    The deny lives in the JSON decision channel; stderr stays free for advisory
-    hints, and exit code stays 0 so the harness reads the decision, not a crash.
+    The deny lives in the JSON decision channel, and exit code stays 0 so the
+    harness reads the decision, not a crash.
     """
     print(json.dumps({
         "hookSpecificOutput": {
             "hookEventName": "PreToolUse",
             "permissionDecision": "deny",
             "permissionDecisionReason": reason,
+        }
+    }))
+    return 0
+
+
+def _emit_additional_context(event_name: str, text: str) -> int:
+    """Deliver ``text`` into the *model's* context for ``event_name``, and exit 0.
+
+    ``hookSpecificOutput.additionalContext`` is the only channel that reaches the
+    model from a hook that exits 0. Claude Code wraps the string in a system
+    reminder and inserts it where the hook fired. Two channels that look
+    equivalent are not:
+
+    * **stderr on exit 0** — routed to the transcript/debug log. On most events
+      it is shown to the user, never to the model. Only exit code 2 feeds stderr
+      back to the model, and on PreToolUse that also blocks the call.
+    * **top-level ``systemMessage``** — a warning shown to *the user*, not the
+      model ("unlike on every other event, where you see the systemMessage and
+      Claude doesn't"). Fine for user-facing notices, useless for steering.
+
+    Supported on SessionStart, Setup, SubagentStart, UserPromptSubmit,
+    UserPromptExpansion, PreToolUse, PostToolUse, PostToolUseFailure,
+    PostToolBatch, Stop, and SubagentStop. NOT on PreCompact — see
+    ``run_precompact``. Strings are capped at 10,000 characters by the harness.
+    """
+    print(json.dumps({
+        "hookSpecificOutput": {
+            "hookEventName": event_name,
+            "additionalContext": text,
         }
     }))
     return 0
@@ -130,7 +164,7 @@ def run_pretooluse() -> int:
         + ``get_symbol_source`` instead.
 
     Strength is set by ``_enforce_mode()`` (``JCODEMUNCH_ENFORCE``): the default
-    ``advisory`` tier warns on stderr but allows (so Read-before-Edit and the
+    ``advisory`` tier nudges the model but allows (so Read-before-Edit and the
     Grep fallback keep working); ``strict`` denies the same calls but only when
     an indexed-repo jcm route can serve them — targeted reads (``offset`` /
     ``limit``), tiny files, non-code files, and paths outside every indexed repo
@@ -206,14 +240,12 @@ def run_pretooluse() -> int:
 
     # Advisory: full-file exploratory read on a large code file — warn but allow.
     # Hard deny breaks the Edit workflow (Claude Code requires Read before Edit).
-    # Stderr text is surfaced to the agent as guidance.
-    print(
+    return _emit_additional_context(
+        "PreToolUse",
         f"jCodemunch hint: this is a {size:,}-byte code file. "
         "Prefer get_file_outline + get_symbol_source for exploration. "
         "Use Read only when you need exact line numbers for Edit.",
-        file=sys.stderr,
     )
-    return 0
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


jgravelle (commented):
Thank you for this — the diagnosis is right, and the writeup did the hard part: separating "the hook didn't fire" from "the hook fired, computed the right text, and shipped it down a channel nobody reads." That distinction is why this sat undetected since v1.22.5.

I verified the central claim independently against the hooks reference rather than taking it on the PR description, since the whole change rests on it. It holds, verbatim:

> Stderr from a hook that exits 0 goes to the debug log only, never the transcript, and Claude never sees it.

and `systemMessage` is documented as "Warning message shown to the user." So the advisory Read/Grep nudges and the SubagentStart briefing were inert exactly as you describe.

Both judgment calls check out against the schemas too:

- `SessionStart` documents `source` as `startup | resume | clear | compact | fork`, so your `compact|resume|fork` matcher is the correct subset and the `startup|clear` exclusion is right for the reason you give.
- `PreCompact` and `TaskCompleted` really do document a `hookSpecificOutput` with `hookEventName` only. Keeping TaskCompleted diagnostics on `systemMessage` rather than blocking completion over advisory findings is the right call.

I also confirmed `_extract_jcm_subcommand` (init.py:773) does the per-subcommand duplicate detection you rely on, so the "adds to an existing install on re-run" claim holds.

## One correction to make before merge

The `_emit_additional_context` docstring says the text is "capped at 10,000 characters by the harness." That is not what happens. Per the docs, past 10,000 characters Claude Code writes the full text to a file and hands the model a path plus a short preview. That is not truncation, it is indirection with a re-read cost — the opposite failure mode from the one you are fixing, and a future maintainer reading "capped" would assume silent truncation and stop worrying about size.

Not a live problem, and I measured rather than guessing: on this repo's index the SubagentStart briefing is **866 characters** and snapshot plus landmarks is **91**. Nowhere near the threshold. Just fix the wording.

## The pre-existing-failure baseline is gone

Worth knowing before you rebase: of the eight failures you attributed to network/env dependence on unmodified main, **four were neither** — `test_install_pack.py` and `test_v1_108_163.py` were caused by an unreleased commit that added a starter-pack CDN cache-key pin. Its new catalog probe consumed the first scripted fixture response, so the download then failed with an unscripted-GET transport error. Three of those tests still returned the exit code they asserted while driving a completely different branch, which is the same shape as the bug you are fixing here.

That is repaired in v1.108.252, and the full suite on current main is **7139 passed, 7 skipped, 0 failed**. So there is no failure allowance to carry forward: anything red after your rebase is real signal.

## Rebase needed

The PR is currently CONFLICTING against main. v1.108.252 also touched `cli/hooks.py` — it added a `_note_transcript_root(data)` call at every hook's stdin load, to fix #421 (`receipt` counted 12 of 348 calls because it only scanned the default `CLAUDE_CONFIG_DIR` profile). That call needs to survive in the functions you refactor, `run_precompact`, `run_taskcomplete` and `run_subagentstart` in particular, where the `json.load(sys.stdin)` moved.

Could you rebase onto main? You know the intent of this change better than a mechanical resolution would.

## While you are in there

Your new `run_sessionstart` is arguably the **best** registration point for #421 that exists. It is the earliest hook to fire on a resumed session, and its payload carries `transcript_path` like every other hook event. Adding `_note_transcript_root(data)` right after the `json.load` would mean a profile gets registered at session start rather than waiting for the first Read or Edit. Entirely optional, and a natural follow-up if you would rather keep this PR to one concern.

## Nit

In `test_briefing_uses_additional_context`, the `if out:` guard lets the test pass vacuously if the mock ever stops taking effect — and the mocked store is patched at `jcodemunch_mcp.storage.IndexStore` while the function imports it locally, so that coupling is worth pinning. An assertion that `out` is non-empty would keep the test honest.

Nothing here is structural. The channel analysis is correct, the tests assert delivery rather than content, and the negative assertions are the load-bearing ones. Rebase and I will merge.

If jCodeMunch is saving you time, tell your friends and colleagues about it!

georgebashi:
Rebased onto `main` (now v1.108.253) and pushed — `MERGEABLE`. All four points addressed, plus the optional one.

## The `_note_transcript_root` conflict

Resolved by hand, as you suggested. The conflict was exactly where you predicted: `.252` wrapped `json.load(sys.stdin)` in the three functions I had refactored, and in `run_precompact` that line no longer exists in the same place — the stdin read stayed in the hook while the snapshot body moved to `_build_session_snapshot`. Git's auto-merge put the call inside the shared builder, which would have been wrong twice over: `_build_session_snapshot` reads no stdin, and `run_sessionstart` calls it too, so the root registration would have moved out of the hooks entirely.

Verified by enumerating call sites rather than reading the diff — all five of yours survived, in the right functions:

| line | function |
|---|---|
| 173 | `run_pretooluse` |
| 358 | `run_posttooluse` |
| 520 | `run_precompact` |
| 550 | `run_sessionstart` ← new |
| 747 | `run_taskcomplete` |
| 892 | `run_subagentstart` |

## #421 in `run_sessionstart`

Took it — the argument that it is the earliest hook on a resumed session is the whole point, and it is two lines.

One deliberate choice: the call sits **before** the `compact|resume|fork` gate, not after. The transcript root is a property of the session, not of whether we inject anything, so a `startup` session — silent on every output channel — still registers its profile. Putting it after the gate would have meant fresh sessions never register, which is most of them. Covered by `test_registers_even_when_no_snapshot_is_injected`, which asserts registration happens on `startup`/`clear` while the existing tests still assert those emit nothing.

## The docstring

Fixed, and thank you for insisting on the distinction — "capped" and "written to a file with a preview" fail in opposite directions, and the one I wrote is the reassuring one. It now says the text is *not* truncated, names the indirection and its re-read cost, and records your two measurements (866 / 91) so the next reader has a sense of scale rather than a rule.

## The nit

You were right, and it was worse than vacuous-on-paper. I removed the `if out:` guard and made the assertion unconditional. Proving the guard was load-bearing: with the mocked `list_repos` returning `[]`, the test now fails on `assert ''` where before it passed silently. The comment pins the coupling you flagged — `run_subagentstart` does `from ..storage import IndexStore` at call time, so patching the module attribute is what works, and rebinding it to a module-scope from-import would break it.

## Baseline

Thanks for the correction on my failure allowance — and the note that three of those tests returned the exit code they asserted while driving a different branch is the sharper half. I had reported eight failures as "network/env-dependent" from a single run without checking any of them, which is the same mistake in miniature: asserting a cause from a signal that was consistent with several.

So I did not carry the allowance forward. Ran the full suite on my branch and on unmodified `main` in a throwaway `git worktree --detach`, same box, same isolation, and diffed the failure **names**:

- **identical sets, 7 names, both sides** — pre-existing and untouched by this change
- 7063 passed on mine vs 7031 on baseline: delta of exactly **+32**, the new `test_hook_output_channels.py` count, nothing else moved

Both are local-environment leakage of the #411 class, and I checked rather than assuming: `test_count_cap_end_to_end.py[env_var]` fails because my real `~/.code-index/config.jsonc` sets `max_folder_files: 100000` and config correctly beats env per .250 — the `[config]` params pass, and all 10 pass under an isolated `CODE_INDEX_PATH`. `test_suggest_corrections.py::test_freshness_correction_no_patch` is *not* config: I have the `watch-all` launchd service actually running, so the code takes its "service running" branch and returns different advice than the test expects. That one survives `CODE_INDEX_PATH` isolation because the service is detected outside the store. Neither is worth filing off the back of this PR, but say the word and I will open one for the watch-service test — it will fail for anyone who followed `watch-install`, which is the documented path.

jgravelle (approved):
Verified the rebase. Approving.

Re-checked the two things a rebase can quietly undo, since the #421 transcript-root work landed in the same functions after you opened this:

- `_note_transcript_root` survives in all five prior hook entrypoints. The diff only moves the `run_precompact` call into the new thin wrapper, which still runs it before anything else.
- The `SessionStart` registration is placed BEFORE the source gate, with the reason stated in the comment. That is the right side of that gate: on `startup` the root is still a real profile whose history `receipt` should count, and it is now learned a whole session earlier than the first Read or Edit would have managed. `test_registers_even_when_no_snapshot_is_injected` pins it.

On the four review points, all addressed. The one I want to record for the next person reading this file:

- `_build_session_snapshot` returning `(text, used_fallback)` is what lets `run_sessionstart` refuse to inject the no-journal message. That message is written for a human ("no live session was found"); injected into model context it is a distractor that reads like a finding. `test_silent_when_no_journal` asserts the refusal, which is the test I would have asked for if it were missing.
- Uninstall needed no change and I checked rather than assumed: `_strip_jcm_hooks` matches any command containing `jcodemunch-mcp` across every event, so `SessionStart` is removed by the existing path.
- `_emit_additional_context` documents the 10,000-character behaviour as a re-read cost rather than truncation, with a measured number beside it (~866 chars for the briefing). Measuring instead of guessing at a threshold is the part worth copying.

One gap I am NOT sending back to you, because it is one line and you have already done two rounds: `config --check` carries its own hardcoded `_expected_hooks` map in `server.py`, so it will report the other five hooks and stay silent about `hook-sessionstart`. `install_status` is generic and needs nothing. I am merging and adding that entry on top so the diagnostic agrees with what `init` actually installs.

Thanks for chasing the channel semantics down to the commit that introduced it. A hook that fires, gates correctly, computes the right text and delivers it nowhere is the worst failure shape there is, because every observable signal says it works.

 -jjg

PR #433, opened 2026-08-09
fix(nextjs): detect src/app App Router layouts, not just root app/

### LINKED ISSUE #432: Next.js plugin silently indexes zero routes for src/app layouts: detect() passes on root next.config.* but _parse_app_router only checks root app/
**Version:** jcodemunch-mcp 1.108.268 (uvx, `[local-embed]`), Windows 11, Python via uv.

## Summary

The Next.js context plugin activates for any repo with a root `next.config.*`, but `_parse_app_router` only looks for `app/` at the repo root. Repos using the officially supported [`src/` directory layout](https://nextjs.org/docs/app/building-your-application/configuring/src-directory) (`src/app/`) index with **zero** Next.js routes — no `nextjs-page` / `nextjs-api` / `nextjs-layout` context entries, empty `nextjs_routes` / `nextjs_api` metadata, and zeroed route stats — even though the repo has hundreds of route files.

Symbol retrieval itself is unaffected (handlers inside `src/app/**/route.ts` are still indexed and queryable), so the failure is silent: the plugin reports "detected: nextjs" and then contributes nothing.

## Repro

Real-world hit: a private Next.js repo (App Router, `src/` layout) with `next.config.mjs` at the root and, under `src/app/`: 6 `page.tsx` + 171 `route.ts`. After a clean `index_folder`, Next.js enrichment reported 0 pages / 0 API routes.

Minimal repro:

```
repo/
  next.config.mjs
  src/app/page.tsx
  src/app/api/users/route.ts   # exports GET
```

Index the folder → `stats()` shows no `nextjs_routes` / `nextjs_api`; no `app-router` tagged context entries.

Expected: 1 page route (`/`), 1 API route (`/api/users`, GET) — identical to the same tree with `app/` at the root.

## Root cause (from the installed 1.108.268 source)

`jcodemunch_mcp/context/nextjs.py`:

```python
def detect(self, folder_path: Path) -> bool:
    for config_name in ("next.config.js", "next.config.ts", "next.config.mjs"):
        ...  # root-level check — PASSES for src/ layouts
```

```python
def _parse_app_router(self, folder_path: Path) -> None:
    """Parse app/ directory for pages, layouts, and API routes."""
    app_dir = folder_path / "app"      # <-- hard-coded root app/
    if not app_dir.is_dir():
        return                          # <-- silent no-op for src/app/
```

and the route-string helper assumes the same prefix:

```python
def _next_route_from_path(file_path: str, app_dir: str = "app") -> str:
```

Secondary occurrence — `jcodemunch_mcp/context/framework_profiles.py` (Next.js profile):

```python
entry_point_patterns=[
    "app/**/page.tsx",
    "app/**/route.ts",
    "app/layout.tsx",
    ...
]
```

## Suggested fix

In `_parse_app_router`, resolve the app dir from the two official locations and thread the matching prefix through:

```python
for candidate in ("app", "src/app"):
    app_dir = folder_path / candidate
    if app_dir.is_dir():
        break
else:
    return
# pass `candidate` down so _next_route_from_path strips the right prefix
```

and mirror `src/app/**` variants in the framework profile's `entry_point_patterns` (Next.js explicitly supports both layouts; when both exist, Next.js itself prefers root `app/`, so keeping root-first candidate order matches framework semantics).

`_parse_middleware` may deserve the same treatment: with the `src/` layout, middleware lives at `src/middleware.ts`.

## Impact

- Silent: `detect()` succeeds, so users see the framework recognized while all route intelligence is missing.
- Affects any `create-next-app --src-dir` project (a first-class, commonly chosen layout).
- Scope is enrichment/stats only — core AST indexing and retrieval keep working, which is why this goes unnoticed.

lilubot:
Fixes #432.

## Problem

`NextjsContextProvider.detect()` passes on any root `next.config.{js,ts,mjs}`, but `_parse_app_router` only probed `<root>/app` and silently returned when absent. Projects using the official [`src/` directory layout](https://nextjs.org/docs/app/building-your-application/configuring/src-directory) (`create-next-app --src-dir`) therefore indexed with **zero** page/API routes while still reporting the framework as detected. Real-world hit that motivated #432: a Next.js repo with 6 `page.tsx` + 171 `route.ts` under `src/app/` and `page_routes: 0, api_routes: 0` after indexing.

## Change

- `_parse_app_router` probes `app/` then `src/app/` and threads the matched prefix into `_next_route_from_path`, so route strings stay identical between layouts. Root `app/` wins when both exist — the same precedence Next.js itself applies.
- `_parse_middleware` additionally probes `src/middleware.{ts,js}` (where the `src/` layout places middleware).
- The `next` framework profile mirrors `src/app/**` entry-point patterns and `src/`-prefixed layer / high-value paths.
- Docstrings updated to describe both layouts.

## Tests

6 new tests (`pytest tests/test_nextjs_provider.py tests/test_context_providers.py` → 51 passed):

- `TestNextRouteFromPath.test_src_app_prefix` — prefix stripping with `src/app`.
- `TestNextjsSrcAppLayout` — src-layout page routing (routes + stats), API handler (endpoint `/api/users` + methods + stats), **root-app-wins precedence when both layouts exist**, `src/middleware.ts` detection with matcher extraction, and metadata (`nextjs_routes` / `nextjs_api`) for the src layout.

All pre-existing root-`app/` tests pass unchanged — behavior for the root layout is byte-identical (`_next_route_from_path`'s default parameter is untouched).

### DIFF
--- a/src/jcodemunch_mcp/parser/context/framework_profiles.py
+++ b/src/jcodemunch_mcp/parser/context/framework_profiles.py
@@ -119,7 +119,9 @@ _NEXT = FrameworkProfile(
         Layer("lib",        ["lib/", "src/lib/"]),
         Layer("api",        ["app/api/", "src/app/api/"]),
     ],
-    high_value_paths=["app/", "src/app/", "lib/", "components/"],
+    high_value_paths=[
+        "app/", "src/app/", "lib/", "src/lib/", "components/", "src/components/",
+    ],
 )
 
 _VUE_SPA = FrameworkProfile(
--- a/src/jcodemunch_mcp/parser/context/framework_profiles.py
+++ b/src/jcodemunch_mcp/parser/context/framework_profiles.py
@@ -107,15 +107,19 @@ _NEXT = FrameworkProfile(
         "app/**/page.tsx",
         "app/**/route.ts",
         "app/layout.tsx",
+        "src/app/**/page.tsx",
+        "src/app/**/route.ts",
+        "src/app/layout.tsx",
         "middleware.ts",
+        "src/middleware.ts",
     ],
     layer_definitions=[
-        Layer("pages",      ["app/"]),
-        Layer("components", ["components/"]),
-        Layer("lib",        ["lib/"]),
-        Layer("api",        ["app/api/"]),
+        Layer("pages",      ["app/", "src/app/"]),
+        Layer("components", ["components/", "src/components/"]),
+        Layer("lib",        ["lib/", "src/lib/"]),
+        Layer("api",        ["app/api/", "src/app/api/"]),
     ],
-    high_value_paths=["app/", "lib/", "components/"],
+    high_value_paths=["app/", "src/app/", "lib/", "components/"],
 )
 
 _VUE_SPA = FrameworkProfile(
--- a/src/jcodemunch_mcp/parser/context/nextjs.py
+++ b/src/jcodemunch_mcp/parser/context/nextjs.py
@@ -1,9 +1,10 @@
 """Next.js context provider — detects Next.js projects and enriches symbols with framework metadata.
 
 When a Next.js project is detected (via next.config.js/ts/mjs), this provider:
-1. Parses app/ for App Router file-based routing → enriches components with route metadata
-2. Parses app/api/ for API route handlers → enriches with endpoint metadata
-3. Detects middleware.ts at project root
+1. Parses the App Router directory (app/ or src/app/) for file-based routing
+   → enriches components with route metadata
+2. Parses <app dir>/api/ for API route handlers → enriches with endpoint metadata
+3. Detects middleware.ts at the project root (or src/ for src-layout projects)
 4. Exposes route metadata via get_metadata()
 """
 
@@ -75,10 +76,11 @@ def _next_api_methods_from_content(content: str) -> list[str]:
 class NextjsContextProvider(ContextProvider):
     """Context provider for Next.js projects (App Router).
 
-    Detects next.config.js/ts/mjs, then parses:
-    - app/**/page.tsx    → page routes with route metadata
-    - app/**/layout.tsx  → layout nesting
-    - app/api/**/route.ts → API handlers with HTTP methods
+    Detects next.config.js/ts/mjs, then parses (app/ or src/app/ — both
+    official layouts; root app/ wins when both exist, matching Next.js):
+    - <app>/**/page.tsx    → page routes with route metadata
+    - <app>/**/layout.tsx  → layout nesting
+    - <app>/api/**/route.ts → API handlers with HTTP methods
     """
 
     def __init__(self) -> None:
@@ -106,9 +108,15 @@ class NextjsContextProvider(ContextProvider):
         self._parse_middleware(folder_path)
 
     def _parse_app_router(self, folder_path: Path) -> None:
-        """Parse app/ directory for pages, layouts, and API routes."""
-        app_dir = folder_path / "app"
-        if not app_dir.is_dir():
+        """Parse the App Router directory (app/ or src/app/)."""
+        # Next.js supports both the root app/ layout and the src/ directory
+        # layout (create-next-app --src-dir). When both exist, Next.js itself
+        # uses root app/ and ignores src/app — probe in the same order.
+        for app_prefix in ("app", "src/app"):
+            app_dir = folder_path / app_prefix
+            if app_dir.is_dir():
+                break
+        else:
             return
 
         for src_file in sorted(app_dir.rglob("*")):
@@ -119,7 +127,7 @@ class NextjsContextProvider(ContextProvider):
                 continue
 
             rel_path = str(src_file.relative_to(folder_path)).replace("\\", "/")
-            route = _next_route_from_path(rel_path)
+            route = _next_route_from_path(rel_path, app_prefix)
 
             if stem == "route":
                 # API route handler
@@ -169,8 +177,8 @@ class NextjsContextProvider(ContextProvider):
         )
 
     def _parse_middleware(self, folder_path: Path) -> None:
-        """Detect middleware.ts at project root."""
-        for name in ("middleware.ts", "middleware.js"):
+        """Detect middleware.ts at the project root or under src/."""
+        for name in ("middleware.ts", "middleware.js", "src/middleware.ts", "src/middleware.js"):
             mw_path = folder_path / name
             if mw_path.exists():
                 try:
--- a/src/jcodemunch_mcp/parser/context/nextjs.py
+++ b/src/jcodemunch_mcp/parser/context/nextjs.py
@@ -77,7 +77,8 @@ class NextjsContextProvider(ContextProvider):
     """Context provider for Next.js projects (App Router).
 
     Detects next.config.js/ts/mjs, then parses (app/ or src/app/ — both
-    official layouts; root app/ wins when both exist, matching Next.js):
+    official layouts; a root app/ or pages/ directory makes Next.js ignore
+    src/ entirely, and parsing matches):
     - <app>/**/page.tsx    → page routes with route metadata
     - <app>/**/layout.tsx  → layout nesting
     - <app>/api/**/route.ts → API handlers with HTTP methods
@@ -110,14 +111,18 @@ class NextjsContextProvider(ContextProvider):
     def _parse_app_router(self, folder_path: Path) -> None:
         """Parse the App Router directory (app/ or src/app/)."""
         # Next.js supports both the root app/ layout and the src/ directory
-        # layout (create-next-app --src-dir). When both exist, Next.js itself
-        # uses root app/ and ignores src/app — probe in the same order.
-        for app_prefix in ("app", "src/app"):
-            app_dir = folder_path / app_prefix
-            if app_dir.is_dir():
-                break
-        else:
-            return
+        # layout (create-next-app --src-dir). Root directories win: if app/
+        # or pages/ exists at the root, Next.js ignores src/ entirely
+        # (https://nextjs.org/docs/app/api-reference/file-conventions/src-folder),
+        # so an ignored src/app tree must not be published as live routes.
+        app_prefix = "app"
+        if not (folder_path / app_prefix).is_dir():
+            if (folder_path / "pages").is_dir():
+                return
+            app_prefix = "src/app"
+            if not (folder_path / app_prefix).is_dir():
+                return
+        app_dir = folder_path / app_prefix
 
         for src_file in sorted(app_dir.rglob("*")):
             if src_file.suffix not in (".tsx", ".ts", ".jsx", ".js"):
--- a/src/jcodemunch_mcp/parser/context/nextjs.py
+++ b/src/jcodemunch_mcp/parser/context/nextjs.py
@@ -105,19 +105,25 @@ class NextjsContextProvider(ContextProvider):
         if self._folder is None:
             self._folder = folder_path
 
-        self._parse_app_router(folder_path)
-        self._parse_middleware(folder_path)
+        # Root directories win: when app/ or pages/ exists at the project
+        # root, Next.js ignores src/ entirely
+        # (https://nextjs.org/docs/app/api-reference/file-conventions/src-folder).
+        # Resolved once, here, so the router and middleware parsers share one
+        # precedence rule and cannot drift apart.
+        src_ignored = (folder_path / "app").is_dir() or (folder_path / "pages").is_dir()
 
-    def _parse_app_router(self, folder_path: Path) -> None:
+        self._parse_app_router(folder_path, src_ignored)
+        self._parse_middleware(folder_path, src_ignored)
+
+    def _parse_app_router(self, folder_path: Path, src_ignored: bool) -> None:
         """Parse the App Router directory (app/ or src/app/)."""
-        # Next.js supports both the root app/ layout and the src/ directory
-        # layout (create-next-app --src-dir). Root directories win: if app/
-        # or pages/ exists at the root, Next.js ignores src/ entirely
-        # (https://nextjs.org/docs/app/api-reference/file-conventions/src-folder),
-        # so an ignored src/app tree must not be published as live routes.
+        # Root app/ wins over src/app. A root pages/ directory (src_ignored
+        # without a root app/) means any src/app tree is dead — publishing
+        # it would invent routes, and inventing routes is a worse failure
+        # than missing them.
         app_prefix = "app"
         if not (folder_path / app_prefix).is_dir():
-            if (folder_path / "pages").is_dir():
+            if src_ignored:
                 return
             app_prefix = "src/app"
             if not (folder_path / app_prefix).is_dir():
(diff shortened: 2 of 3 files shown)


jgravelle (changes requested):
Thanks, Lilu. Filing #432 with a real repro (6 `page.tsx` and 171 `route.ts` under `src/app/`, and the provider still reporting `page_routes: 0, api_routes: 0`) and then landing the fix six minutes later is about as clean as a report-plus-patch gets. CLA signed without being asked, too. Verdict: the approach is right and I want this in. Two changes before merge, plus a rebase.

## What I verified

I didn't take any of this on trust, so here's what I actually ran:

- **Your root-`app/` claim holds.** `_next_route_from_path(file_path, app_dir="app")` already carried the `app_dir` parameter at `nextjs.py:26` before this PR, so the default path really is untouched. I checked the pre-PR source rather than reading your note and moving on.
- **Targeted run:** `pytest tests/test_nextjs_provider.py tests/test_context_providers.py` gives 51 passed, matching the number you reported exactly.
- **Non-vacuity.** I ran your six new tests against unmodified `main` in a detached worktree. Four of them fail there for real reasons (`KeyError: 'nextjs_routes'` and friends), which is what I want to see. Two pass on `main`: `test_src_app_prefix` passes because the `app_dir` parameter predates the PR, and `test_root_app_wins_over_src_app` passes trivially because `main` never looks at `src/app` at all. I'm keeping both. Worth being precise about what the second one buys us, though: it guards against a future reordering, it doesn't demonstrate today's fix. That's a good test to have written, just not evidence for this change.
- **Full suite on your branch:** 7442 passed, 7 skipped, 0 failed. `uv run ruff check src/` clean.
- **CI**: I approved the workflow runs on your SHA (they were held behind the first-time-contributor gate, nothing to do with your changes). Tests, Replay and Health Radar all green on `01cd104`.

## Change 1 (blocker): root `pages/` has to win too

The new loop is:

```python
for app_prefix in ("app", "src/app"):
```

That defers to a root `app/` directory, but not to a root `pages/` one. Next.js's own docs are explicit ([src-folder](https://nextjs.org/docs/app/api-reference/file-conventions/src-folder)): "src/app or src/pages will be ignored if app or pages are present in the root directory."

So a repo with a root `pages/` directory alongside `src/app/`, which is exactly what a half-finished App Router migration looks like, gets its `src/app` tree parsed and published as live routes that Next.js ignores at runtime.

Why that matters more here than it would in most projects: missing routes is the bug you're fixing, but inventing routes is the worse failure mode for jCodeMunch. The whole product claim is that absence is honest. If we report a route, it has to exist. A wrong answer costs us more than a blank one.

Should be about three lines: bail out when a root `pages/` directory is present and no root `app/` exists.

## Change 2: `high_value_paths` is internally inconsistent

You extended `layer_definitions` to cover both layouts:

```python
Layer("components", ["components/", "src/components/"]),
Layer("lib",        ["lib/", "src/lib/"]),
```

but `high_value_paths` only picked up `src/app/`:

```python
high_value_paths=["app/", "src/app/", "lib/", "components/"],
```

`src/lib/` and `src/components/` should be in there as well. One line.

## Not yours, and we're not asking you to fix it

The automated reviewer flagged missing `.js` / `.jsx` variants in `entry_point_patterns`. That gap is pre-existing: the root patterns are `app/**/page.tsx`, `app/**/route.ts`, `app/layout.tsx`, `middleware.ts`, with no JS variants either. You mirrored the patterns that were already there, which is the correct thing to have done. Filed separately as #435. Please don't absorb it into this PR.

Worth telling you what else came out of reading your patch: the Nuxt provider has the identical defect (`nuxt.py:129` probes root `pages/` and nothing else), and because Nuxt 4 moved the default `srcDir` to `app/`, a stock Nuxt 4 project has been indexing zero routes. Nobody had reported it. That's #434. You found a defect class and we'd been treating it as one framework's problem.

## Rebase

Your merge base is v1.108.268 and `main` is now at v1.108.270. GitHub reports no textual conflict, but two releases have landed underneath you. House convention is that the author rebases, so please pull `main` into your branch when you make the two changes above.

## Timebox

**2026-08-24.** If I haven't heard back by then, I'll make the two changes myself on top of your commit, preserving your authorship and crediting you in the CHANGELOG, and merge. That's a guarantee the work lands with your name on it either way, not a deadline you're at risk of losing anything to. If you'd rather do it yourself, the branch is yours until that date and I'd prefer that outcome.

 -jjg

jgravelle:
Not a nudge, and nothing here changes what's asked of you. The two requested changes still stand (root `pages/` precedence, and `src/lib/` plus `src/components/` in `high_value_paths`), the review comment has the detail, and your timebox is unchanged at 2026-08-24. Not shortened.

This is one thing you'd otherwise find out by accident. Reviewing your PR surfaced the identical defect in the Nuxt provider, which nobody had reported: `NuxtContextProvider` probed `pages/` at the project root, and Nuxt 4 moved the default `srcDir` to `app/`, so stock Nuxt 4 projects reported the framework detected and zero routes. That fix shipped today as #434 in 1.108.271, credited to your report in both the CHANGELOG and the GitHub release notes. You found a defect class. We'd been treating it as one framework's problem.

One related note: #435 (the JS-extension gap the automated reviewer flagged) shipped for `nuxt` and `nestjs` but deliberately not for `next`. Your PR edits those exact lines, and we're not going to force a conflict onto the rebase we asked you for. `test_ts_patterns_carry_js_variants` exempts `next` by name, so the exemption gets deleted when your PR lands rather than sitting there unnoticed.

https://pypi.org/project/jcodemunch-mcp/1.108.271/

 -jjg

jgravelle:
@lilubot **The Health Radar comment on this PR is wrong. That's our bug, not yours.**

The bot posted `Composite: C -> D (-10.1 pts)` and `REGRESSION on 1 axis`, driven entirely by `churn_surface` going `60 -> 0`. Every other axis moved by 0.4 or less. Nothing in your patch caused that.

### What actually happened

1. `.github/workflows/health-radar.yml` checks out the PR with `fetch-depth: 0`, so the head side has full history.
2. `.github/actions/health-radar/action.yml` then fetched the base branch with `git fetch origin "$BASE" --depth=1`. Against an already complete clone, that doesn't just limit a download. It shortens the repo and writes `.git/shallow`. Measured on this repo: 1390 commits, 482 of them in the last 90 days, before. 1 and 1 after.
3. `churn_surface` is complexity x log(1 + churn), and churn comes from `git log --since=<N> days ago`. With one commit visible, every file scored churn <= 1, so the base came back artificially healthy. Every PR then got charged for the gap.
4. `_score_churn_surface` is a step function (<500 -> 60, <1500 -> 40, <3000 -> 20). So the artifact jumps whole buckets instead of nudging a number.

### Controlled proof

Same commit, identical git tree hash on both sides, only history depth differing. Shallow side: composite 82.2, grade B. Full side: composite 75.5, grade C. Five of six axes bit-identical. `churn_surface` 60.0 vs 20.0. The same commit graded B against itself as a base and C as a head.

One honesty note on scope. That reproduction ran jcm 1.108.265, not the 1.88.0 this CI action pins, so my absolute numbers aren't the ones in the bot comment above. What transfers is the mechanism and the direction, not the digits. I did not reproduce your specific raw of 2589.9647 and I'm not claiming I did. Your baseline raw of 386.78 and my shallow side raw of 318.85 land in the same <500 bucket and both score 60, which is consistent with the same artifact.

### Fixed

Commit `52d0791` on `main`, CI green. Both sides now unshallow when `.git/shallow` exists. Four guard tests added in `tests/test_health_radar_action.py`.

Root cause on why it sat unnoticed: that test file used to declare the Action's shell and YAML steps testable only in real CI. So the one component that can publish a grade on someone else's work had no local coverage at all. That's on us, and it's what the new tests close.

### What this means for your branch

The workflow resolves the action from the PR's own checkout (`uses: ./.github/actions/health-radar`), and your branch is based on v1.108.268. So it still carries the broken step, and the sticky comment above will keep re-rendering the same false verdict on every push until your branch includes `52d0791`. The rebase already discussed brings it in. Until then, treat that grade as void.

### Nothing new is being asked of you

The two requested changes are unchanged: root `pages/` precedence, and `src/lib/` plus `src/components/` in `high_value_paths`. The timebox is unchanged at **2026-08-24**. This adds no scope and moves no goalposts. It retracts a grade we shouldn't have published against your work.

 -jjg

jgravelle:
Thanks for turning both of these around, and for doing the rebase on top. The root-wins rule in `_parse_app_router` is exactly the shape I was hoping for, and rebasing rather than merging is what made the rest of this comment possible. Four things: one correction I owe you, one new finding, one blocker that isn't yours, and one thing you offered to file that you shouldn't.

### First, a correction I owe you

My first pass at reproducing your root-`pages/` fix concluded it was broken. It wasn't. `git fetch origin pull/433/head` had handed me the stale pre-force-push commit `01cd1042`, so I was measuring the defect you'd already fixed. I caught it by diffing the worktree SHA against the PR head before I said anything, so nothing wrong got posted. Flagging it anyway. We've published wrong line numbers on this PR's issues once already, and you should know when a review of your branch nearly went sideways on our end rather than yours.

### The Health Radar false regression is fixed, and your branch is what proves it

Your rebase pulled in `52d0791`, so the radar re-ran with both sides at equal history depth. It now posts:

- Composite: D -> D (0.0 pts)
- Verdict: no meaningful change
- `churn_surface`: 0 -> 0, every axis flat

The sticky comment was replaced in place, so the old "REGRESSION, C -> D, `churn_surface` 60 -> 0" text is no longer what the thread shows. Your PR is both the thing that exposed that bug and the thing that confirms the fix.

Being straight about what the honest number says: `churn_surface` is 0 on both sides, and the composite is D, not the C the old comment showed as a baseline. The shallow base wasn't just noisy, it was systematically flattering us. Our real grade on that axis was always 0. That's a finding about our codebase, not a win.

### New finding: `_parse_middleware` didn't get the same precedence rule

Change 1 put the root-wins decision in `_parse_app_router`. `_parse_middleware` wasn't changed to match, and it still probes unconditionally:

    for name in ("middleware.ts", "middleware.js", "src/middleware.ts", "src/middleware.js"):

Next.js ignores `src/` entirely when a root `app/` or root `pages/` exists, which is the same doc line change 1 cites. So a stale `src/middleware.ts` gets published with its matcher even though it never runs.

Reproduced at your actual head `dd1489b3`, in an isolated worktree, in both shapes:

- root `app/` present: `src/app/page.tsx` correctly ignored, `src/middleware.ts` published (`tags=['nextjs-middleware']`)
- root `pages/` present: `src/app/page.tsx` correctly ignored, `src/middleware.ts` published (`tags=['nextjs-middleware']`)

So change 1 is correct in both shapes. This is strictly the middleware method.

It's the same failure class you already agreed on, one method over: inventing a route or handler that doesn't serve is worse than missing one. That isn't a knock on the patch. It's the kind of thing that only becomes visible once the first fix makes the precedence rule explicit.

On the fix shape, your call entirely. `_parse_middleware` needs the same root-wins decision. The cleanest version probably resolves "does this project use `src/` at all" once in `load()` and hands it to both methods, so the two can't drift apart again.

### The CLA gate re-opened, and it's a hard merge blocker

Not something you did wrong. Commit `261c3a03` is authored `lilubot`, and `dd1489b3` is authored `dmdorta1111`, so the sticky CLA comment flipped from "All committers have signed the CLA" back to not_signed. That's almost certainly the same person on a second machine with a different git config. The bot keys on the email address, nothing more.

It does block merge, because the project is dual-licensed. Two ways out, whichever is less annoying: sign for the second identity, or re-author the commit under the identity that already signed.

### The 3 test failures are ours. Please don't file them.

You offered to open a separate issue for the 3 failures in `tests/test_count_cap_end_to_end.py`, all `[env_var]` parametrizations, on the theory that they reproduce on clean `origin/main` so it must be your environment.

It isn't your environment. Our test reads developer machine state. Reproduced deterministically here.

The `env_var` fixture calls `config_module.load_config()` with no arguments. That resolves to `CODE_INDEX_PATH` or `~/.code-index/config.jsonc` and loads the developer's real global config. If you have `max_folder_files` set in yours, it overrides the env var the fixture just set, and `file_count` comes back 6 instead of the cap of 2.

Proof: pointing `CODE_INDEX_PATH` at a temp config containing `{"max_folder_files": 5000}` fails exactly your three parametrizations, by name:

    test_the_cap_a_user_set_is_the_cap_the_walk_uses[full-walk-env_var]
    test_the_cap_a_user_set_is_the_cap_the_walk_uses[explicit-paths-env_var]
    test_a_truncated_index_says_so[env_var]

Our suite is green only because `max_folder_files` happens to be commented out in this machine's config. CI is green on your branch for the same reason: the runner has no global config setting it. A test that passes on two machines and fails on a third, for a reason none of them can see, is the defect.

The irony is worth naming out loud: that file's docstring is "The file-COUNT cap, tested where the USER stands", and it reads the developer's config. We'll take it. Don't spend time on it. Thanks for reporting it precisely enough to be reproducible. "3 failures, all `env_var`, and they also fail on clean main" is what made it findable.

### Status and next

All three checks on `dd1489b3` are green: Tests, Replay, Health Radar.

Nothing new is being asked beyond the middleware finding. The timebox is unchanged at **2026-08-24**, and there's no expectation you hit it early. If you'd rather not take the middleware change, just say so and we'll do it on top of your commit with your authorship preserved, either way.

 -jjg

jgravelle:
CLA status is the only mechanical blocker left: 1 of 2 committers signed. Your `lilubot` identity has signed. `dd1489b3` is authored as `dmdorta1111`, which is the second committer.

Two routes, either works. Sign as `dmdorta1111` at the cla-assistant link in the bot's comment. Or, since `dd1489b3` is the branch tip, re-author it under the identity that already signed, which drops the second committer from the PR:

```
git commit --amend --author="lilubot <lilubot@users.noreply.github.com>" --no-edit
git push --force-with-lease
```

That's one command and no second signature. Your call, no preference beyond convenience.

If you amend, the SHA changes, so the fork workflows need maintainer approval again. I'll approve them promptly, so don't worry if checks sit idle for a bit.

All three checks (Tests, Replay, Health Radar) are green on `dd1489b3` right now.

 -jjg

jgravelle (approved):
Verified rather than taken on trust, and everything checks out.

**Identity rewrite.** `git rev-parse <old>^{tree} <new>^{tree}` agrees for both pairs on my side too — `261c3a0`/`bde6e19` at `c707878341b0`, `dd1489b`/`06bc7bb` at `7c7a665d4f82`. Byte-identical content, identity-only change, and all three commits now author and commit as the signed identity. Rewriting rather than signing for a misconfigured second identity was the right call, and the repo-local override is the fix that keeps it from recurring.

**Middleware precedence.** `93e6046` is the shape I suggested: one `src_ignored` resolution in `load()`, handed to both parsers. I checked the thing that could have quietly changed behaviour — inside `_parse_app_router` the root `app/` directory is already known absent at that point, so `src_ignored` reduces to the original `pages/` test. Semantically identical, no regression smuggled in alongside the fix.

**Non-vacuity, re-run here.** I checked out the pre-fix `nextjs.py` under the post-fix tests: both new cases fail with exactly the signature you quoted — `FileContext(description='Next.js middleware', tags=['nextjs-middleware'])` published for `src/middleware.ts` in both root shapes. Restored, both pass. Your three cited files: 84 passed. `ruff check src/` clean.

Asserting that the live root tree still publishes (`page_routes: 1`) in the root-`app/` case is the detail I appreciate most — suppression bugs love to over-reach, and that test is what would catch it.

Approving. Fork workflows are approved and running against `93e6046`; I will merge once they are green.

jgravelle:
Merged as `4bbf5c5`. All three workflows green at `93e6046` (Tests 9/9 including lint, Replay, Health Radar), and `main` re-verified after the merge: 84 passed on the three affected test files, `ruff check src/` clean.

Thanks for the identity rewrite — reproducing the tree hashes locally made that trivial to accept, and preserving byte-identical diffs was the detail that made it reviewable rather than a leap of faith. The `src_ignored` consolidation is exactly right: with one resolution point the two precedence decisions genuinely cannot drift, which is more than the original suggestion asked for.

Confirming your last two notes: `f03bc6b` did land the `env_var` fix on `main`, so nothing to file there. Nothing outstanding on this one.

Issue #436, opened 2026-08-10
get_dead_code_v2 warnings tell callers to pass entry_point_patterns, which no layer accepts


jgravelle:
`get_dead_code_v2` emits two warnings telling the caller to pass `entry_point_patterns`. It doesn't accept that parameter, at any layer. The advice can't be followed, and it's emitted exactly when the tool has degraded and the caller most needs a way out.

Found while auditing `entry_point_patterns` for #435. Filed separately because it's a different verdict on a different tool.

## The advice

`src/jcodemunch_mcp/tools/get_dead_code_v2.py:810-825` builds `signal_warning`, and `:823-824` appends:

> Pass entry_point_patterns to make unreachable_file discriminate, or lower min_confidence to see the weaker verdicts.

`:827-831` sets `framework_warning`:

> No standard entry points detected (e.g. main.py, app.py, `__main__.py`). Signal 1 (unreachable_file) fires for every symbol, inflating dead code counts. Pass entry_point_patterns to identify framework-specific roots (e.g. handler functions for AWS Lambda, route modules for FastAPI).

## Why it can't be followed

Three layers, none of which accept it:

1. **Signature** (`get_dead_code_v2.py:505-513`) takes `repo`, `min_confidence`, `include_tests`, `max_results`, `file_pattern`, `storage_path`, `degeneracy_cutoff`. No `entry_point_patterns`.
2. **MCP schema** (`server.py:3112-3148`) exposes that set minus `storage_path`, which the server supplies. Still no `entry_point_patterns`.
3. **Dispatcher** (`server.py:6114-6127`) forwards those six caller arguments plus `storage_path`, so a client that sends the key anyway has it dropped.

The parameter is real, it just belongs to a different tool. `find_dead_code` accepts it in the signature (`tools/find_dead_code.py:164`), exposes it in the schema (`server.py:3722`) and receives it from the dispatcher (`server.py:6326`). The warning text appears to have been carried across without the parameter.

## A second problem in the same string

`signal_warning` fires whenever any signal lands in `uninformative` (`get_dead_code_v2.py:806`), but the sentence names `unreachable_file` unconditionally. Per the #408 measurement in `CHANGELOG.md`, the signal that's degenerate most often is `not_barrel_exported`. So on the common case the advice names a parameter that doesn't exist and a signal that isn't the one that failed. Same string, same fix, noted because whoever rewrites the text has to decide what it should say instead.

## Why this is worse than a stale docstring

[...]

PR #439, opened 2026-08-11
fix: allow shallow Windows Git roots

### LINKED ISSUE #438: Windows drive-root child Git repositories trip broad-root guard

On Windows, `index_folder()` rejects an explicit absolute Git working tree located directly under a drive root, such as `F:\example-repo`, as too broad.

## Reproduction

With a Git repository whose working-tree root is directly below a drive root:

```powershell
git -C F:\example-repo rev-parse --show-toplevel
# F:/example-repo

py -3.13 -c "from pathlib import Path; print(Path(r'F:\example-repo').parts)"
# ('F:\\', 'example-repo')

jcodemunch-mcp index F:\example-repo --no-ai-summaries
```

Actual result:

```text
Resolved path 'F:\example-repo' is too broad to index safely
(fewer than 3 path components).
```

The broad-root guard uses the path component count. A local drive anchor plus one child has only two `Path.parts`, so an exact Git working-tree root is classified the same way as a broad non-project directory.

## Expected vs actual

Expected:

- an explicit path that is exactly a Git working-tree root at `X:\repo` proceeds through normal indexing;
- the drive root `X:\` remains rejected;
- a shallow non-Git directory remains rejected.

Actual:

- every local-drive path with two components is rejected unless it is explicitly listed in `trusted_folders`.

## Environment

- jCodeMunch version: 1.108.271
- Python version: 3.13.7
- OS: Microsoft Windows NT 10.0.26200.0

## Anything else you found along the way

Related work:

- #125 introduced the broad-root guard after a relative path could resolve against the MCP server working directory.
- #175 added the explicit `trusted_folders` escape hatch.
- #321 / #322 corrected UNC share depth while intentionally leaving local-drive behavior unchanged, so this is a distinct local-drive case.

A project-scoped `.jcodemunch.jsonc` containing `{"trusted_folders": ["."]}` is a working local workaround.

A focused fix can preserve the guard and allow only a two-component Windows path with a `.git` directory or file at the requested root. I have a regression test and patch in progress.

JayceeB1:
## Summary

Fixes #438.

- allow an exact Git working-tree root directly below a local Windows drive root (for example, `F:\repo`)
- keep drive roots and shallow non-Git directories blocked by the broad-root guard
- add Windows-specific regression coverage for all three cases
- document the behavior change in the changelog

## Root cause

On Windows, `Path.parts` represents `F:\repo` with only two components: the drive anchor and the repository directory. The broad-root guard therefore rejects a valid repository located directly under a drive, even though the selected directory is an exact Git working-tree root.

## Safety

This does not lower the global minimum-depth threshold. The exception is limited to Windows local-drive paths with a safety depth of two whose selected directory contains a `.git` directory or file. The drive root itself, shallow non-Git directories, POSIX paths, and UNC handling remain unchanged. Existing `trusted_folders` behavior is also preserved.

## Testing

- red/green regression: the child-drive Git-root case failed before the production change and passes afterward
- review regression: the non-Git test was shown to fail when the mocked runner exposed `.git`, then isolated with an explicit negative filesystem probe
- `uv run pytest tests/test_tools.py -q -k WindowsDriveRootPathSafety` — 3 passed
- `uv run pytest tests/ -q` — 7618 passed, 17 skipped, 1 pre-existing warning
- `uv run ruff check src/` — passed
- real Windows smoke test indexed a separate repository directly below `F:\` without a project `.jcodemunch.jsonc`: 35 files and 984 symbols; the new narrow Git-root exception was reported

## Related work checked

- #125 introduced the broad-root guard
- #175 added explicit `trusted_folders`
- #321 / #322 cover UNC logical depth and intentionally left local-drive behavior unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,24 @@ The Windows regression coverage keeps the positive repository case beside both
 negative cases and mocks the filesystem probes, so the runner's drive layout
 cannot decide the result.
 
+⚠⚠ **Review note, recorded because the wrong version of this nearly shipped on
+maintainer advice.** The first review asked for the UNC scope predicate to be
+dropped as redundant with the shared depth helper. It is not, and the two are
+not the same kind of rule: `_path_safety_part_count` measures DEPTH, while
+`not drive.startswith("\\")` bounds SCOPE. A UNC share root has one real part
+and the depth helper adds one for the `\\server\share` anchor, so it computes to
+**exactly two -- the same depth as `C:\repo`**. With the predicate gone, a share
+root holding a `.git` would have been admitted, handing a whole file server to
+the indexer through the guard that exists to prevent it (#321/#322).
+
+⚠ **The regression test for it was ALSO wrong on its first run, in a way that
+passed.** It patched `os.path.exists` to a blanket `True`, which answers
+`_is_container()`'s `/.dockerenv` probe as well -- that drops `_MIN_PATH_PARTS`
+from three to two, so `2 < 2` skips the guard and the assertion never reaches
+the code under test. The probe is narrowed to the `.git` path and the reason is
+recorded at the patch site. **A mock broad enough to satisfy the assertion can
+be broad enough to bypass what the assertion is about.**
+
 ## [1.108.275] - 2026-08-12 - A pattern that matches nothing now says so
 
 ### `entry_point_patterns` failed silently ([#446](https://github.com/jgravelle/jcodemunch-mcp/issues/446))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [Unreleased] - A Windows drive-root child can prove it is a repository
+
+### Exact Git working trees no longer trip the broad-root guard ([#438](https://github.com/jgravelle/jcodemunch-mcp/issues/438))
+
+On Windows, an explicit repository at `X:\repo` has only two logical path
+components, so `index_folder` rejected it alongside genuinely broad paths. The
+guard now accepts that narrow case only when `.git` exists at the selected root.
+Drive roots, shallow non-Git directories, POSIX paths, and UNC depth handling are
+unchanged.
+
+The Windows regression coverage keeps the positive repository case beside both
+negative cases and mocks the filesystem probes, so the runner's drive layout
+cannot decide the result.
+
 ## [1.108.275] - 2026-08-12 - A pattern that matches nothing now says so
 
 ### `entry_point_patterns` failed silently ([#446](https://github.com/jgravelle/jcodemunch-mcp/issues/446))
@@ -136,7 +150,6 @@ Item 1 of the QA pass — the `install-pack` archive guard missing drive-absolut
 member names — is [#447](https://github.com/jgravelle/jcodemunch-mcp/issues/447),
 with @elfrost's [PR #443](https://github.com/jgravelle/jcodemunch-mcp/pull/443)
 open against it.
-
 ## [1.108.273] - 2026-08-12 - A pattern that names two extensions and matches neither
 
 ### v1.108.271's #435 fix matched nothing ([#445](https://github.com/jgravelle/jcodemunch-mcp/issues/445))
--- a/src/jcodemunch_mcp/tools/index_folder.py
+++ b/src/jcodemunch_mcp/tools/index_folder.py
@@ -329,12 +329,21 @@ def _is_shallow_windows_git_root(path: Path) -> bool:
     selected a working-tree root rather than the drive itself or a broad parent.
     The corresponding POSIX case was considered and intentionally left unchanged
     so this exception remains scoped to the reported Windows drive-root behavior.
+
+    ⚠ The depth check and the UNC check answer different questions and neither
+    is redundant with the other. ``_path_safety_part_count`` is a DEPTH rule;
+    ``not drive.startswith("\\\\")`` is a SCOPE rule. A UNC share root has one
+    real part and the depth helper adds one for the ``\\server\share`` anchor,
+    so it computes to exactly two -- the same value as ``C:\repo``. Dropping the
+    UNC test would therefore admit ``\\server\share`` itself, which #321/#322
+    classify as too broad whatever it happens to contain.
     """
     drive = str(path.drive)
     return (
         os.name == "nt"
         and _path_safety_part_count(path) == 2
         and bool(drive)
+        and not drive.startswith("\\\\")
         and os.path.exists(path / ".git")
     )
 
--- a/src/jcodemunch_mcp/tools/index_folder.py
+++ b/src/jcodemunch_mcp/tools/index_folder.py
@@ -321,6 +321,24 @@ def _path_safety_part_count(path: Path) -> int:
     return count
 
 
+def _is_shallow_windows_git_root(path: Path) -> bool:
+    r"""Return whether ``path`` is a Git root directly below a local drive.
+
+    ``C:\repo`` has a safety depth of two and normally trips the broad-root
+    guard. A ``.git`` directory or file at that exact path proves the caller
+    selected a working-tree root rather than the drive itself or a broad parent.
+    The corresponding POSIX case was considered and intentionally left unchanged
+    so this exception remains scoped to the reported Windows drive-root behavior.
+    """
+    drive = str(path.drive)
+    return (
+        os.name == "nt"
+        and _path_safety_part_count(path) == 2
+        and bool(drive)
+        and os.path.exists(path / ".git")
+    )
+
+
 @lru_cache(maxsize=512)
 def _is_trusted(
     folder_path: Path, trusted_folders: tuple, whitelist_mode: bool = True
@@ -1445,7 +1463,8 @@ def index_folder(
     _MIN_PATH_PARTS = 2 if container else 3
     path_part_count = _path_safety_part_count(folder_path)
     if path_part_count < _MIN_PATH_PARTS:
-        if not is_trusted:
+        shallow_windows_git_root = _is_shallow_windows_git_root(folder_path)
+        if not is_trusted and not shallow_windows_git_root:
             error_msg = (
                 f"Resolved path '{folder_path}' is too broad to index safely "
                 f"(fewer than {_MIN_PATH_PARTS} path components). "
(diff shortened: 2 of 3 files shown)


jgravelle:
Reviewed the draft. The approach is right and the scope discipline is exactly what this guard needs: a narrow exception, both negative cases tested alongside the positive one, and the prior art in #125 / #175 / #321 / #322 checked before touching it. The `.git`-at-that-exact-path proof is the correct discriminator — it distinguishes "the caller selected a working tree" from "the caller selected a broad parent" without lowering the global threshold. Thank you for the real smoke test on `F:\` too; that is the part we could not have run here.

Four things before this comes out of draft. Two are substantive, two are small.

**1. Two notions of depth in one guard.** `_is_shallow_windows_git_root` uses `len(path.parts)`, but the guard it gates uses `_path_safety_part_count(path)`, and those two deliberately disagree: the helper adds 1 when the drive is a UNC anchor, so `\server\share\repo` is `2` by `len(parts)` and `3` by the helper.

Your `not drive.startswith("\\\\")` does correctly exclude UNC, so the behaviour is right today. The concern is that it is right *twice over* rather than once — the UNC case is handled by a second, independent condition instead of by the depth function that already knows about it. If `_path_safety_part_count` learns another anchor form later, the two silently drift.

`_path_safety_part_count(path) == 2` reads as the same rule the guard applies, and lets the UNC clause go. We have precedent for caring about this specifically: #436 landed with `_matches_any_pattern` **imported** rather than reimplemented, on the grounds that two definitions of one concept is the original defect wearing a new costume.

**2. `test_shallow_non_git_directory_remains_too_broad` reaches the real filesystem.** That test patches `Path.exists` but not `os.path.exists`, and the helper uses `os.path.exists`. So the assertion depends on `F:\Users\.git` not existing on whatever machine runs it. It passes on CI because `F:` is absent there, and on your box because that directory has no `.git` — but it is the runner's drive layout deciding, not the test.

This is worth more than its size, because we have same-day evidence of the failure mode. Earlier today `TestWindowsUNCPathSafety::test_unc_repo_at_share_child_is_not_too_broad` — the sibling test your class is modelled on, and reasonably so — went red on `windows-latest / 3.10` with `OSError: [WinError 64] The specified network name is no longer available`. It patches `Path.resolve` to a fake UNC path, and something downstream still made a real call, so Windows went looking for `\server\share` over the network. It passed on re-run of the identical commit. Same class, less dangerous here because a missing local drive fails fast instead of hanging.

Patching `os.path.exists` in that third test the way you already do in the first one closes it.

**3. Minor: compute inside the branch.** `shallow_windows_git_root` is evaluated on every `index_folder` call, but only read when `path_part_count < _MIN_PATH_PARTS`. Moving it inside the `if` saves a stat on every index on Windows.

**4. Needs a CHANGELOG entry to merge.** Behaviour change to a security guard. Not a draft-stage concern, just flagging it now so it is not a surprise later.

**One open question, genuinely yours to answer rather than a change request.** The exception is Windows-only, and #438 is a Windows report, so that is defensible. But "an exact Git working-tree root is project-scoped, not a broad parent" is not a Windows-specific fact — `/repo` on Linux is also two parts and also rejected outside a container. The POSIX gap is narrower, since `_MIN_PATH_PARTS` already drops to 2 in containers, which covers most of where `/repo` shows up. Happy either way; if you keep it Windows-only, a line in the docstring saying the POSIX case was considered and left alone would stop the next reader re-deriving it.

**Timebox, so this does not sit:** if you would like to make these changes, we will hold the fix for you until **2026-08-26**. If we have not heard by then, we will land the change ourselves with your authorship on the commit and credit in the CHANGELOG and release notes, and you keep the report and the design either way. Say the word at any point if you would rather we take it from here — a preempted contributor is a cost we would rather pay ourselves than pass to you.

JayceeB1:
I’ll make the four requested changes: reuse `_path_safety_part_count()` as the single depth definition, isolate the negative test from the runner filesystem, move the Git-root probe inside the shallow-path branch, and add the CHANGELOG entry. 

I’ll keep the exception Windows-only to match the reported case and preserve the narrow security boundary, with a docstring note that the POSIX case was considered and intentionally left unchanged. 

I’ll move the PR out of draft after the full suite is green again.

I’m genuinely happy to contribute, and I appreciate the thoughtful review and the care you take not to pass the cost of a handoff on to contributors.

JayceeB1:
Implemented the four review changes on 56b319f: the helper now reuses _path_safety_part_count() as the single depth definition, the negative test explicitly mocks the .git probe, the probe is evaluated only inside the shallow-path branch, and the behavior change is recorded in CHANGELOG.md. The docstring also records the decision to keep POSIX behavior unchanged. Local verification on the rebased 1.108.274 base: 7618 passed, 17 skipped; 
uff check src/ passes. The complete GitHub matrix is green across Ubuntu and Windows on Python 3.10–3.13, along with lint, radar, retrieval quality, and CLA. Moving the PR out of draft.

jgravelle:
Thank you — all four landed cleanly, and I verified the matrix myself: 9/9 green, CLA signed, radar and retrieval-quality gates pass. The docstring note recording that POSIX was considered and left alone is exactly what I hoped for; the next reader will not have to re-derive it.

**Then I checked my own advice, and item 1 was wrong. It opens a hole, and it is my mistake rather than yours.**

## `_path_safety_part_count(path) == 2` does not subsume the UNC exclusion

I told you the helper already knows about UNC so the `not drive.startswith("\\\\")` clause could go. It does know — but it knows in the direction that breaks this. The helper *adds* one for a UNC anchor, and a UNC **share root** has one part:

```
path                        drive               parts   safety_count   == 2 ?
'C:\repo'                   'C:'                2       2              True
'C:/'                       'C:'                1       1              False
'\server\share'            '\server\share'    1       2              True    <-- share ROOT
'\server\share\repo'       '\server\share'    2       3              False
```

So `\server\share` now satisfies the depth test, `bool(drive)` is true for it, and the only remaining condition is `.git` at that path. **A UNC share root that happens to be a Git working tree would be admitted as narrow.** That contradicts #321 / #322, which decided deliberately that share-root depth stays too broad, and it is the one direction a broad-root guard must never move.

Before my suggestion your two-clause version refused it. After it, the guard depends on `\server\share\.git` not existing.

## Where my reasoning went wrong, since it matters for the fix

I called it "two notions of depth". Half of that was right: `len(path.parts) == 2` genuinely *was* a second depth definition, it disagreed with the guard's own, and replacing it with the helper is a real improvement worth keeping.

But `not drive.startswith("\\\\")` was never a depth rule. It is the **scope** predicate — "this exception applies to local drives only" — and it is load-bearing on its own. One depth definition plus one scope predicate is not duplication, and I collapsed the two when I asked you to drop it.

So the shape I should have suggested:

```python
return (
    os.name == "nt"
    and _path_safety_part_count(path) == 2   # one depth definition (this part was right)
    and bool(drive)
    and not drive.startswith("\\\\")          # scope: local drives only — NOT redundant
    and os.path.exists(path / ".git")
)
```

## The suite cannot see this, which is the second lesson

`test_unc_share_root_remains_too_broad` still passes — and passes for the wrong reason. It patches `Path.exists` but not `os.path.exists`, so the `.git` probe hits the real filesystem and returns False because no such share exists on the runner. **That is precisely the defect I asked you to fix in `test_shallow_non_git_directory_remains_too_broad`, sitting in the pre-existing test that would otherwise have caught my bad advice.**

Green CI is therefore not evidence for this case. A regression test needs to mock the probe so `.git` appears present at `\server\share`, and assert the guard still refuses.

## How would you like to handle it

This is my error and I do not want it to cost you another round trip. Three options, your pick:

1. **You push the two-line change** plus the UNC regression test, and I merge.
2. **I push to your branch**, keeping your authorship on the existing commit and adding mine on top.
3. **You are done and I take it from here** — the PR merges as your work with credit, and I land the correction as a follow-up commit.

Any of those is fine and none of them costs you standing. The timebox from my earlier comment is void; nothing here is on a clock while I am the one who introduced the problem.

For what it is worth, the underlying change is good and I still want it. The narrow `.git`-at-that-exact-path proof is the right discriminator, and everything about the reuse, the test isolation and the branch placement is now where it should be.

jgravelle:
Pushed the correction to your branch (`51bea3c`), keeping your commit and authorship intact. Option 3 of the three I offered, taken because the fault was mine and it should not cost you another round trip. If you would rather carry it yourself, force-push over it and I will re-review.

**What changed, and the measurement behind it.**

The UNC predicate is restored beside the depth check, because the two are not the same kind of rule:

```python
os.name == "nt"
and _path_safety_part_count(path) == 2   # DEPTH
and bool(drive)
and not drive.startswith("\\\\")          # SCOPE
and os.path.exists(path / ".git")
```

Run against the shipped helper rather than a reimplementation:

| path | `drive` | `len(parts)` | safety depth | `_is_shallow_windows_git_root` (pre-fix) |
|---|---|---|---|---|
| `C:\repo` | `C:` | 2 | 2 | True |
| `\server\share` | `\server\share` | **1** | **2** | **True** |
| `\server\share\repo` | `\server\share` | 2 | 3 | False |

A share root has one real part and the helper adds one for the anchor, so it lands on exactly the value `C:\repo` does. My "the depth helper subsumes it" claim was measuring the wrong thing. With the predicate gone, `\server\share` holding a `.git` is admitted, which hands the indexer a whole file server through the guard that exists to stop that (#321/#322).

**The test was wrong too, in a way that passed.**

Worth flagging because it is the more transferable half. My first regression test patched `os.path.exists` to a blanket `True`. That also answers `_is_container()`'s `/.dockerenv` probe, which drops `_MIN_PATH_PARTS` from three to two — so `2 < 2` skips the guard entirely and the assertion never reached the code under test. It failed identically with and without the fix, which is what gave it away. The probe is now narrowed to the `.git` path with the reason recorded at the patch site.

**Added:**
- `test_unc_share_root_with_git_remains_too_broad` — fails against the predicate-less version, and fails on the *effect* (the share root gets indexed), not on the shape of the condition. It asserts the depth premise first, so if `_path_safety_part_count` ever changes, the test says why it no longer applies instead of quietly passing.
- `test_unc_repository_below_share_root_is_unaffected` — a control pinning that `\server\share\repo` clears the guard on depth alone and never reaches the exception. Passes both sides. It exists so nobody can justify widening the exception later with "UNC repos need it."

Also rebased onto `main` (the CHANGELOG conflicted with .275) and recorded the review error in the CHANGELOG entry rather than deleting the bad advice, so the next reader does not re-derive it.

Your four original changes are untouched and were all correct. Local suite green on the Windows class (5/5), `ruff` clean. Once CI comes back I will merge this and close #438.

PR #443, opened 2026-08-12
fix(install-pack): confine archive members to the install directory

elfrost:
Item 1 from the QA pass in the companion issue. Independent of the other two items there, which are `SECURITY.md` edits and yours to make or decline.

## The gap

`install-pack` pre-scans archive members and rejects a leading `/` or a `..` anywhere. On Windows that is necessary but not sufficient: a member named `C:/Windows/Temp/x` — or its backslash form — contains no `..` and does not start with `/`, so it passes. Then at the join, `base / relative` with an absolute `relative` **discards `base` entirely**.

Reproduced against the guard and join exactly as written, on Windows:

```
guard_pass=True   rel='C:/Windows/Temp/evil.txt'     escapes base? True
guard_pass=True   rel='C:\Windows\Temp\evil2.txt'    escapes base? True
guard_pass=True   rel='\\server\share\evil.txt'      escapes base? True
guard_pass=False  rel='../escape.txt'                (correctly blocked)
guard_pass=False  rel='/pack/abs.txt'                (correctly blocked)
```

## The fix, in your own idiom

`IndexStore._safe_content_path()` already does this correctly, and already treats repository-supplied paths as untrusted on every cached write: resolve the candidate, compare `commonpath` against the base. The same three inputs through that helper are all rejected.

So this adds `_safe_extract_path()` in the same shape and uses it at the join, rather than introducing a new pattern. Catching the escape by *resolution* instead of by *pattern* is the part that matters — it covers UNC-shaped names and any separator form without having to enumerate them, which a string test can never finish doing.

The string pre-scan is deliberately kept. It aborts early with a clear message before any file is opened; the confinement check is the authoritative backstop.

## Scope and honesty about the threat path

Low severity, nobody is blocked, and I would not want this to read as more than it is. The archive comes from your own endpoint over HTTPS, so exploiting it needs a hostile response from that endpoint or its redirect chain (`follow_redirects=True`, content-type is the only validation) — not merely a network position.

It seemed worth closing anyway because of where it sits relative to the rest of the posture: release wheels are Sigstore-signed precisely so a downloaded artifact ties back to the workflow that built it, and packs are downloaded and extracted into `~/.code-index/` — a directory `SECURITY.md` itself flags for file-integrity monitoring — with neither a signature nor a complete extraction guard.

## Tests

Five new cases in `tests/test_install_pack.py`: an ordinary member resolving under the base, parent traversal (direct and interior), and the three absolute forms that previously passed the string test.

```
[...]

Issue #435, opened 2026-08-10
entry_point_patterns is TypeScript-only in the next, nuxt and nestjs profiles: JS projects get no reachability seed


jgravelle:
`entry_point_patterns` in the `next`, `nuxt` and `nestjs` framework profiles listed only TypeScript extensions. Equivalent JavaScript projects get no matching entry points, so a caller feeding these patterns into reachability analysis can classify genuine framework roots as unreachable.

Raised by the automated Codex reviewer on [#433](https://github.com/jgravelle/jcodemunch-mcp/pull/433) against the `next` profile. A sweep of `framework_profiles.py` found it in three profiles, not one. Filed separately because it predates [#433](https://github.com/jgravelle/jcodemunch-mcp/pull/433) and isn't that PR's to fix.

**Status: two of three fixed in v1.108.271. `next` is still open and is deliberately sequenced after [#433](https://github.com/jgravelle/jcodemunch-mcp/pull/433).**

## Current state

| profile | JS variants | state |
|---------|-------------|-------|
| `vue-spa` | present (sibling globs) | was already correct |
| `react-spa` | present (sibling globs) | was already correct |
| `express` | present (sibling globs) | was already correct |
| `nuxt` | present (brace alternation) | **fixed in v1.108.271** |
| `nestjs` | present (brace alternation) | **fixed in v1.108.271** |
| `next` | **missing** | **open**, deferred behind [#433](https://github.com/jgravelle/jcodemunch-mcp/pull/433) |

Line numbers are omitted on purpose. The file moved in v1.108.271, and the original body of this issue already cited stale ones once. Read the profiles by symbol name (`_NUXT`, `_NEXT`, `_NESTJS`).

### Fixed

`_NUXT` entry patterns now use `{ts,js,mjs}` brace alternation, and they name both root-relative and `app/`-relative paths so Nuxt 3 and Nuxt 4 defaults are both covered. `server/` is deliberately **not** mirrored under `app/`, because it stays at the project root in both layouts. Layer definitions cover both layouts too.

`_NESTJS` now uses `{ts,js}` in its entry patterns and in all four `Layer` globs. NestJS is TypeScript-first by convention, so this was the weakest of the three, but `nest new --language js` is a supported scaffold.

### Still open

[...]

PR #451, opened 2026-08-12
fix(storage): bind process creation time against PID reuse (#450)

### LINKED ISSUE #450: Process-registry/lock liveness check trusts recycled PIDs: dead servers report as live after PID reuse
## Summary

`_is_pid_alive()` (`src/jcodemunch_mcp/storage/process_locks.py`) verifies that *a* process with the recorded PID is running, but never that it is the *same* process that registered it. After PID reuse, a registry entry for a long-dead server reads as live indefinitely: `live_processes()` (`src/jcodemunch_mcp/storage/process_registry.py`) skips pruning it and `get_session_stats` reports it as an active jcodemunch process.

Observed on 1.108.272 (Windows 11, Python 3.11).

## What happened

`get_session_stats` reported two "live" jcodemunch servers, versions 1.108.193 and 1.108.198, `started_at` 2026-07-28 / 2026-07-29 — two weeks old. Checking the actual processes:

```
ProcessId : 20720  CommandLine : "C:\Program Files\AMD\CNext\CNext\cncmd.exe" watch 21044
ProcessId : 37688  CommandLine : "C:\Program Files\Google\Chrome\Application\chrome.exe" --type=renderer ...
```

Both original servers were long dead (hard-killed with their sessions, so `unregister()` never ran — the case the registry's liveness check exists for). Windows recycled both PIDs to unrelated processes; those are `STILL_ACTIVE`, so `GetExitCodeProcess` truthfully answers "alive" and the stale rows survive every prune pass, forever.

Near-miss worth noting: acting on the report, an operator (me, via an agent) came one command-line check away from killing a Chrome renderer and an AMD service believing they were orphaned jcodemunch servers. The docstring's own rationale — "readers verify PID liveness instead of trusting the file" — is defeated by reuse: the check answers "is this PID taken?", not "is my process still there?".

## Why the existing Windows fix doesn't cover this

The `GetExitCodeProcess` improvement in `_is_pid_alive` correctly handles *exited process whose PID stays queryable through an open handle*. PID reuse is the opposite failure: a **different, genuinely running** process now owns the PID. No exit-code check can distinguish it; only an identity comparison can.

## Suggested fix

Bind identity at `register()` time and compare at read time:

- Record the process **creation time** alongside the PID (Windows: `GetProcessTimes`, already used and documented in `runtime_identity`; POSIX: `/proc/<pid>/stat` field 22, or `psutil.Process(pid).create_time()` where available).
- In `live_processes()` / lock-holder inspection, treat *pid alive but creation time mismatch* (beyond a small tolerance) as **dead → prune**, same as a missing process.
- The registry entry already carries `started_at`; today it is display-only. Even comparing that wall-clock value against the live process's creation time (with generous skew tolerance, e.g. ±60 s) would have caught this two-week mismatch.

The same weakness applies to every consumer of `_is_pid_alive`: scope/watcher lock holders (`process_locks.LockHolder`) can equally be "held" by a recycled PID, which makes a genuinely stale lock look owned and blocks takeover.

Related: #371 (on-demand process identity) — that feature landing would provide exactly the identity primitive this check needs.

## Repro sketch

1. Start a jcodemunch server; note its registry row in `<storage>/_processes/<pid>.json`.
2. Hard-kill it (no `unregister()`).
3. Cause/wait for PID reuse (on Windows, PIDs recycle aggressively; for a deterministic test, monkeypatch `_is_pid_alive` input or spawn processes until the PID is reassigned).
4. `live_processes()` returns the dead server's row; `get_session_stats` lists it as a live process with the old version string.

phantom-man:
Fixes #450.

## What

`_is_pid_alive()` verifies that *a* process with the recorded PID is running, but not that it is the *same* process that wrote the registry row or lock file. After the OS recycles a PID, entries for long-dead holders read as live indefinitely — in the field we watched `get_session_stats` report two-week-old servers whose PIDs had been recycled to a Chrome renderer and an AMD service, and cleanup nearly killed those innocent processes. The same weakness lets a recycled PID "hold" a watcher/index-write lock and block takeover.

## How

Identity is bound at write time and checked at read time:

- New `process_create_time(pid)` in `process_locks.py` — OS creation time for an arbitrary PID. Windows: `OpenProcess` + `GetProcessTimes` (the same primitive `runtime_identity` already uses for self), with full `argtypes`/`restype` per the module's existing warning. Linux: `/proc/<pid>/stat` starttime + `/proc/stat` btime. Other platforms: `None`.
- New `_is_live_holder(pid, expected_create_time)` — PID alive **and** creation time within 2s of the recorded value. The tolerance only absorbs float rounding and btime jitter; a recycled PID mismatches by minutes-to-weeks.
- `register()` (process registry) and `acquire()` (locks) record `create_time`; `live_processes()`, `inspect()`, and `acquire()`'s stale-recovery treat alive-but-mismatched as dead → stale → prune/reclaim.

## Compatibility

Rows and locks written by earlier versions carry no `create_time`; readers fall back to the existing liveness-only behavior for them, so mixed-version stores degrade gracefully rather than mass-pruning. Platforms without a creation-time source behave identically to today. If the holder is alive but its creation time cannot be read, the check also falls back to liveness-only rather than declaring a possibly-live holder dead.

## Tests

13 new tests in `tests/test_pid_reuse_identity.py`, written first and confirmed failing, including the exact bug shape: a registry row / lock file naming a *genuinely alive* PID (the test's own) with a mismatched `create_time` is pruned and the lock reclaimed, plus back-compat (no `create_time` → old behavior) and roundtrip cases. `test_process_locks.py`, `test_v1_108_172.py`, and `test_watcher_lock.py` all pass unchanged on this branch (93 passed locally on Windows).


phantom-man:
recheck

jgravelle:
Reviewed. **This is a careful implementation and it gets the hardest part right.** I have approved the CI runs — they were sitting at `action_required` because it is your first PR here, so nothing had executed on this branch yet.

What is right, specifically:

- **The unknown case resolves in the safe direction, in both consumers.** `_is_live_holder` falls back to liveness-only both when no `create_time` was recorded and when the current value cannot be read. For locks that means unknown → *held* (never steal from a possibly-live holder); for the registry it means unknown → *keep the row* (never prune a live server). That was the thing I most wanted from this change and you got it without being told the second half.
- Back-compat degrades rather than mass-prunes, which matters because every existing store is pre-fix.
- The Windows path carries the full `argtypes`/`restype` set. That is not decoration — the module's own warning documents the 64-bit HANDLE truncation trap, and it is easy to copy the call and drop the declarations.
- Linux field indexing is correct: `rsplit(b")", 1)` past the comm field, then index 19 for stat field 22.
- `maybe_takeover` inherits the fix for free through `acquire`, so watcher takeover is covered without touching it.
- Tests written first and confirmed failing.

## The one substantive finding: a clock step can make the check confidently wrong on Linux

Your tolerance comment says both sides derive the value with the same code from the same OS constant, so a genuine match is near-exact. On Windows that is exactly true — FILETIME is absolute and stable.

On Linux it is not, and you half-identified why: you compute `btime + starttime/HZ`, and you note btime can shift ~1s under clock sync. **The shift is not bounded at 1s.** `/proc/stat` btime is derived from wall clock minus uptime, so any *step* correction moves it — chrony stepping at startup, a VM resuming from suspend, a container host resyncing after drift. A step larger than `_CREATE_TIME_TOLERANCE_S` moves every recorded `create_time` out of tolerance at once.

The consequence lands on the dangerous side of the asymmetry this whole change is built around: a **live** holder is judged recycled, so its lock is reclaimed. That is two watchers on one repo, or two concurrent index writes. By the reasoning in #450, blocking work is the bug we are fixing and corrupting it is strictly worse — and unlike the unknown case, this one bypasses your safe fallback, because the value reads fine and simply disagrees.

Worth noting it is a fleet-wide simultaneous failure rather than a per-process one: one clock step invalidates every lock and every registry row on the machine at the same moment.

**A fix that keeps your design:** on Linux, record the boot-relative `starttime` ticks rather than the absolute epoch value. Ticks-since-boot are immune to clock steps, and PIDs do not survive a reboot, so cross-boot comparability is not needed — though pairing it with btime (or a boot id) so a stale-across-reboot row is recognised rather than accidentally matched would be tidier. Windows needs no change.

If you would rather not take that on, widening the tolerance is *not* the answer I would pick — it trades a rare wrong-steal for a permanently weaker identity check, and the whole value here is that a recycled PID mismatches by minutes-to-weeks.

## Two small ones

**`watcher._is_pid_alive` is now a loaded gun.** ⚠ To be clear about severity: it is **not a live defect** — it is a thin wrapper with no production caller, referenced only by three assertions in `test_watcher_lock.py`. But it is a public-looking helper in the watcher module that bypasses the new identity check, and its tests make it look supported. The next person who reaches for it gets pre-fix semantics silently. Either delete it and have those tests call `process_locks` directly, or make it delegate to `_is_live_holder`.

**`process_registry`'s module docstring is now stale.** Lines 12-15 still say readers "filter by PID liveness" and that the module "deliberately reuses `process_locks._is_pid_alive` rather than inventing a second liveness notion". Both describe the pre-fix design.

That one is a little pointed at us rather than you: v1.108.274 shipped yesterday specifically because two `SECURITY.md` sentences were true when written and were not revisited as the code grew. Same class, one day later.

## Not blocking

- `_is_live_holder` opens the process twice — once in `_is_pid_alive`, once in `process_create_time` — and `live_processes()` loops over every row. Foldable into one handle, but correctness first; I would not churn it now.
- Needs a CHANGELOG entry before merge. Happy to add it myself if you would rather stop here.

## Status

CI is running now. **The Linux `/proc` path has never executed anywhere** — your local verification was Windows-only and CI had not run — so that is the result I am most interested in, independent of the btime question above.

Nothing here is on a clock. If you want to take the Linux starttime change, it is yours; if you would rather this land as-is and we follow up, say so and I will handle it with your authorship intact.

jgravelle:
Following up on the `btime` finding with a concrete patch, so this is not blocked on a design discussion.

**The tolerance comment already names the mechanism, and I think it under-sizes it.** It says btime "can shift ~1s under clock sync", which is true of *slewing* — chrony and ntpd correct small offsets by adjusting tick rate, and that stays inside 2.0s. The case the tolerance does not cover is a **step**: `/proc/stat`'s btime is derived from the current wall clock minus boot-relative time, so `settimeofday`-class adjustments move it by the full step. Those are routine rather than exotic — laptop suspend/resume, VM snapshot restore, a container host correcting its clock, or the first real NTP sync on a machine that booted with a bad RTC.

⚠ **Stated limit: I have reasoned this from how the kernel derives btime and from your own jitter note. I have not measured a clock step on Linux** — this box is Windows, and CI will not reproduce it either. I am not asking you to take the magnitude on faith, because the fix removes the dependency entirely and is therefore correct whether the step is 3 seconds or 300.

**Why it matters more than an ordinary edge case:** every recorded `create_time` moves *at once*, so after a step a genuinely live holder reads as recycled. Locks get stolen from a running server and registry rows get pruned. That is the corrupting direction, and it routes around the safe fallback you built — the value parses fine and the process is alive, so nothing reaches the unknown branch. It is also a failure mode `main` does not have today, which is the part that makes it a merge blocker rather than a follow-up.

**The fix is to drop btime and stay in the boot-relative domain:**

```python
             start_ticks = int(fields[19])
-            with open("/proc/stat", "rb") as fh:
-                btime = next(
-                    int(line.split()[1]) for line in fh if line.startswith(b"btime")
-                )
-            return btime + start_ticks / os.sysconf("SC_CLK_TCK")
+            # ⚠ Deliberately NOT converted to wall clock. /proc/stat's btime is
+            # derived from the current wall clock, so a settimeofday-class step
+            # (suspend/resume, VM restore, first NTP sync) moves every recorded
+            # value at once and reads live holders as recycled. This value is
+            # only ever compared against one produced by this same function on
+            # this same machine, so the epoch is irrelevant to the comparison
+            # and seconds-since-boot is strictly safer than seconds-since-epoch.
+            return start_ticks / os.sysconf("SC_CLK_TCK")
```

The value stops being comparable across a reboot, which is worth stating plainly rather than glossing: a lock file surviving a reboot could match a new process that happens to share both the PID and a similar start offset. **That is not a regression** — it degrades to exactly the bare-PID behaviour `main` has today, in a case far rarer than the PID reuse this PR fixes, whereas the btime version is *worse* than today under a condition that is common.

Two consequences worth folding in:
- `test_own_create_time_is_sane`'s "created after 2020" assertion becomes Windows-only; on Linux the sane bound is `0 <= ct <= uptime`.
- The docstring and the tolerance comment should say the Linux value is seconds-since-boot and the Windows value is seconds-since-epoch, since the two platforms no longer share an epoch. They were never compared across platforms, but the next reader should not have to derive that.

I am happy to push this to your branch and keep your authorship, the same as I did on #439 earlier today. **Timebox so this does not sit: if I have not heard from you by 2026-08-19, I will push the patch above, merge, and credit you in the CHANGELOG and release notes.** Say the word before then and it is yours instead — and if you disagree with the analysis, say that too and I will hold. Everything else in this PR is right and I would rather not have it wait on one function.

phantom-man:
Taken, in `d5804f7` — your analysis is right and I'd rather own the fix than have it land around me. Point by point:

- **Linux is now boot-relative**, exactly your patch: `starttime/HZ`, btime dropped, with your warning comment kept nearly verbatim so the next reader inherits the reasoning. You're right that my tolerance comment named the mechanism and under-sized it — slewing stays inside 2.0s, a step doesn't, and the failure it produces is fleet-wide simultaneous *and* lands on the corrupting side of the asymmetry while routing around the unknown-case fallback. No disagreement to hold for.
- **Cross-reboot degradation stated in the code**: a lock file surviving a reboot falls back to bare-PID matching — pre-fix behavior, in a case far rarer than a clock step.
- **`test_own_create_time_is_sane` is per-domain now**: after-2020 stays Windows-only; Linux asserts `0 <= ct <= uptime` read from `/proc/uptime`.
- **`watcher._is_pid_alive` deleted** rather than delegated — a liveness-only wrapper can't honestly delegate to `_is_live_holder` without an `expected_create_time`, so removal was the option that doesn't leave a second notion lying around. Its tests now exercise `process_locks._is_pid_alive` directly, with a comment saying why.
- **`process_registry` docstring updated** — readers verify liveness + creation-time identity via `_is_live_holder`. Fair hit; same one-day-later class as your SECURITY.md example.
- **CHANGELOG entry added under `[Unreleased]`** — happy to have you reword it to house style at merge.
- **Double `OpenProcess` left as-is** per your note — correctness first.

The 93 tests across `test_pid_reuse_identity` / `test_process_locks` / `test_v1_108_172` / `test_watcher_lock` pass locally on Windows against this branch; the Linux `/proc` path is now genuinely CI's to judge, and I'm as interested in that result as you are — this box is Windows too.

For transparency: this account's contributions are authored with Claude Code operating the workflow end to end, including this reply.

jgravelle:
Could you rebase this onto current `main`? Specifics, so you are not hunting for it:

**Only `CHANGELOG.md` conflicts.** I checked with `git merge-tree` against current `main` — all four source and test files (`process_locks.py`, `process_registry.py`, `watcher.py`, `test_pid_reuse_identity.py`, `test_watcher_lock.py`) merge clean. The merge base is `ce2c667` (v1.108.273), and three releases have landed since: .274, .275 and .276 today.

The conflict is mechanical. Your entry sits under `## [Unreleased]`, and 1.108.276 consumed that section this morning — it became `## [1.108.276] - 2026-08-13`. On rebase, put your entry under a fresh `## [Unreleased]` heading above it.

**There is a second reason this is worth doing now, and it is not cosmetic.** A conflicting PR has no `refs/pull/451/merge`, and `pull_request` workflows run against the *merge* ref rather than the head. So this PR cannot run the test matrix at all in its current state — the absence of checks here is that, not a queue. Rebasing is what makes CI able to run.

Related: we found today that fork PRs from first-time contributors were having their runs created and then silently held for approval (`conclusion=action_required`), which never surfaces in `gh pr checks`. That is now relaxed, but it means your #459 merged on a local trial-merge run of mine rather than its own green matrix. Your work was not the problem; our CI configuration was, and it was invisible from the place anyone would have looked.

⚠ Apologies for the close/reopen notifications on this PR and on #443 earlier — that was me testing whether reopening re-provokes a held workflow run. It does not, which is now written into our checklist.

**On the fix itself:** I have not reviewed the diff in detail yet and will do that properly once it rebases. One thing that may be useful — I hit a real instance of #450 on this machine this morning after a reboot: twelve `_watcher_*.lock` files in `~/.code-index/`, all naming PID 87576 from 2026-06-27, plus a stale `_processes/47292.json`. In that case the recorded PIDs happened to be genuinely dead, so nothing falsely read as live, but it is the exact on-disk shape your change is about. I have the files backed up if a fixture would help; say the word and I will attach them.

No rush, and thanks for the two PRs today.

phantom-man:
Rebased onto current `main` (`8d2def6`) and force-pushed — `1c7fa62` + `f94799e`. As you said, only `CHANGELOG.md` conflicted; the entry now sits under a fresh `## [Unreleased]` above `[1.108.276]`, and the five source/test files applied clean. 97 tests across the identity/locks/registry/watcher suites (plus the html file-class tests now on this base) pass locally on Windows. The merge ref should exist now, so the matrix is CI's to run — including the Linux `/proc` path's first genuine execution, which we're both waiting on.

**Yes to the fixture files, please.** Twelve watcher locks all naming one PID from six weeks prior plus a stale registry row is exactly the on-disk shape worth pinning: a fixture test can drop those files into a tmp store and assert `live_processes()` prunes the row and `acquire()` reclaims every lock — covering both the genuinely-dead case you hit and, by editing the recorded `create_time`, the recycled-PID case this PR exists for. Attach them on the issue or here, whichever is less friction, and I'll fold the fixture in as a follow-up commit on this branch.

Also appreciated the transparency on the held-workflow discovery and the #459 trial-merge — good to know its green came from your run rather than a phantom matrix.

jgravelle:
Thanks for rebasing — and **your code is fine.** The matrix reads `1 failed, 7763 passed, 21 skipped`, and the single failure is not in anything you wrote.

## What failed

```
FAILED tests/test_claude_md_rotation.py::test_older_releases_boundary_starts_below_the_rotation
AssertionError: boundary says 1.108.273; the release below the rotation is 1.108.272.
```

**The rebase dropped the entire `## [1.108.273]` section from `CHANGELOG.md`.** On `main` that heading is at line 324; on this branch the file goes `## [1.108.274]` straight to `## [1.108.272]`. Your `[Unreleased]` section and the `.276` heading both survived — only `.273` was lost, which is the signature of a conflict resolution that took one side wholesale across a hunk boundary rather than a mistake in your entry.

`CLAUDE.md` keeps the three newest releases in full and says `Older releases (1.108.273 and earlier)`. The gate cross-checks that boundary against the 4th-newest heading in `CHANGELOG.md`. With `.273` missing, the 4th-newest becomes `.272`, so the two disagree and the build fails.

## The fix

Restore the `## [1.108.273] - 2026-08-12 - A pattern that names two extensions and matches neither` section from `main` — it is lines 324–396 there, ending where `## [1.108.272]` begins. Nothing else needs touching.

```bash
git fetch origin main
git checkout origin/main -- CHANGELOG.md   # then re-add your [Unreleased] entry on top
```

Or restore just that section by hand if you would rather not re-apply your own entry.

## Two things worth saying

⚠ **This is not a style gate and it was not being pedantic.** It caught the deletion of a released version's changelog entry — real history loss that a reviewer skimming a rebase diff would very plausibly wave through, because the diff looks like "changelog churn" and the file is 16,000 lines. That is precisely the case it exists for.

⚠ **It is also not Ubuntu-specific.** I initially told our maintainer this was failing on the four Ubuntu jobs; the Windows jobs simply had not finished yet. It fails identically everywhere, as a file-content assertion should. Correcting that here so nobody chases a platform difference that does not exist.

**On the CI itself:** this run is the first green-lit matrix on any of your PRs. Your fork's runs were previously being created and then silently held for approval, which never surfaces in `gh pr checks` — that is fixed, and your push is what proved it works end to end. Worth stating plainly that #459 merged without its own matrix ever running, on a local trial-merge of mine instead. That was our configuration, not anything you did.

Push the changelog restore and this should go green. I will review the fix itself properly then — I have not read the `process_locks` / `process_registry` changes in detail yet.

phantom-man:
Restored in `f1ba33b` — the `[1.108.273]` section back verbatim from `main` (73 lines), plain push, no history rewrite. Your read of the mechanism was exact: my resolution took the upstream side wholesale across the hunk boundary, and my post-resolution ordering check then validated against a *prose mention* of the `.273` heading elsewhere in the file rather than the heading line itself — so it passed while the section was gone. The re-check is line-anchored now (`^## \[` only), and `test_claude_md_rotation` passes locally on this branch.

Point taken about what the gate is for — a released version's history vanishing inside 16,000 lines of "changelog churn" is exactly what a human reviewer waves through. It earned its keep here.

And noted with some satisfaction: 1 failed / 7763 passed means the Linux `/proc` path survived its first genuine execution on all four Ubuntu jobs. The matrix should go fully green on this push.

jgravelle:
Thanks for the quick turnaround — but hold on merging: **CI is green and `CHANGELOG.md` is still wrong.** Not your code, again; the restore landed twice.

## What is in the file now

The whole `## [1.108.273]` section appears **twice**:

- **Line 344** — heading fused to the end of the `.274` section's closing sentence:
  ```
  with @elfrost's [PR #443](...)
  open against it.## [1.108.273] - 2026-08-12 - A pattern that names two extensions and matches neither
  ```
- **Line 417** — a correctly formed heading, blank line above, content identical to `main`.

Counted rather than eyeballed: the section title text and its first subheading (`### v1.108.271's #435 fix matched nothing`) each occur **2** times.

The rendered effect is that `.274` now ends mid-sentence in a mangled paragraph, immediately followed by a duplicate release section.

## Why CI passed anyway

The rotation gate matches `^## \[(\d+\.\d+\.\d+)\]` — **line-anchored**. The copy at 344 is not at line start, so the regex cannot see it. The gate finds exactly one well-formed `.273`, agrees with `CLAUDE.md`'s boundary, and passes. All nine jobs green.

⚠⚠ **I made the same mistake checking your fix, which is the part worth recording.** I verified the `.273` section was byte-identical to `main` (4287 chars both sides) and that all 517 released versions were present, and reported it as correct. Both facts were true. I checked for *absence* and used a line-anchored regex to do it — **the same instrument as the gate, carrying the same blind spot**, so the two agreeing told us nothing. Duplication was invisible to both. A verification that shares an assumption with the thing it is verifying is not independent evidence.

## The fix

Delete the fused copy: lines 344–415 inclusive, then restore the `.274` section's final line to end cleanly at `open against it.` — the truncated `.274` tail and the duplicate are one contiguous block. The good copy at 417 needs no changes.

```bash
git fetch origin main
git checkout origin/main -- CHANGELOG.md
# then re-apply ONLY your [Unreleased] block at the top
```

That is the lower-risk route: take `main`'s file wholesale and re-add your own entry, rather than surgically removing a 70-line region. Your `[Unreleased]` entry is the only thing you need to carry over, and it is self-contained.

⚠ Worth checking the rendered file after, not just the diff. Both failures so far have been in the seams between sections, which is precisely where a diff view is least legible and a 16,000-line file is least reviewable.

**Your implementation continues to look good and is not implicated in either round.** I have now read the `process_locks` / `process_registry` / `watcher` changes; comments on those to follow separately so they are not tangled up with the changelog mechanics.

phantom-man:
Fixed in `d737648`, your lower-risk route exactly: `main`'s `CHANGELOG.md` taken wholesale, only the `[Unreleased]` block re-applied. The diff of the commit is `1 insertion, 73 deletions` — the fused block gone, the `.274` tail ending cleanly at `open against it.`

Your postmortem point landed, and it's the sharpest lesson of this whole PR: my "restore" verified the `.273` section's *absence* with a line-anchored scan — the same instrument as the gate, same blind spot — so it couldn't see the fused copy and confidently inserted a duplicate. Two agreeing instruments that share an assumption are one instrument. The verification set this time deliberately doesn't share it: substring counts of the section title and first subheading (1 each), a scan for any `## [` not at line start (zero), and the decisive one — **`git diff upstream/main -- CHANGELOG.md` is a single 21-line hunk containing only the `[Unreleased]` block**, which bounds every possible seam error at once. Rotation gate green locally.

There's some irony in a PR about not trusting a PID without checking its identity being held up twice by checks that didn't verify what they thought they were verifying. Looking forward to the storage-changes comments.

jgravelle:
Code review, as promised — separate from the changelog mechanics. **The implementation is sound and I have no blocking objections.** Details below, including one hole I went looking for and did not find.

## The design is right

Recording the creation time at *write* time and comparing at *read* time is the correct shape. PID reuse is only detectable against something the OS assigns and never reissues for that slot, and creation time is the cheapest such anchor available without a dependency.

## The Linux epoch choice is the best part of this

Keeping `/proc/<pid>/stat` starttime **boot-relative** rather than converting to wall clock via `btime` is right, and the comment explaining why is the kind I want in the tree. `btime` is derived from the current wall clock, so a `settimeofday`-class step — suspend/resume, VM restore, first NTP sync on a fresh boot — moves every recorded value at once and reads every live holder as recycled. **A fix that mass-invalidates on clock adjustment would be worse than the bug**, because it converts a rare stale row into a fleet-wide false negative at exactly the moment a machine comes back up.

⚠ Worth noting the deliberate asymmetry with `runtime_identity._linux_process_start`, which *does* add `btime`. That is correct there — it needs a globally meaningful timestamp for a payload — and correct to avoid here, where the value is only ever compared against one this same function produced on this same machine. Two callers, two epochs, for a stated reason. I would keep them separate rather than DRY them into one helper; the shared-primitive temptation is what would silently reintroduce `btime` here later.

Verified the field arithmetic: after `rsplit(b")", 1)[-1]`, `fields[0]` is stat field 3 (`state`), so starttime (field 22) is index 19. Correct.

## The hole I went looking for and did not find

`_is_live_holder` returns `True` when the PID is alive but `process_create_time` returns `None`. I expected that to defeat the fix in the exact reported case: on Windows, if `OpenProcess` fails for a recycled PID now owned by another user, the identity check silently degrades to liveness-only and the stale row stays "live" forever. Your own report cites a two-week-old row resolving to an AMD service, which would typically run as SYSTEM.

**Measured on Windows 11, unelevated shell, `PROCESS_QUERY_LIMITED_INFORMATION`:**

```
csrss.exe     pid=1592  create_time=1786613651.43
wininit.exe   pid=1708  create_time=1786613653.49
services.exe  pid=1784  create_time=1786613653.95
lsass.exe     pid=1812  create_time=1786613653.98
svchost.exe   pid=1944  create_time=1786613654.16
```

All succeed, including `lsass.exe`, a protected process. `PROCESS_QUERY_LIMITED_INFORMATION` is specifically the access right that crosses that boundary, and choosing it over `PROCESS_QUERY_INFORMATION` is what makes the fallback rare rather than routine. **The concern does not hold and the fallback direction is right anyway** — treating an unreadable-but-alive PID as live is the conservative error, since the alternative is reclaiming a lock from a running process.

## Interaction with `_is_pid_alive`, checked

`_is_live_holder` calls `_is_pid_alive` first, which uses `GetExitCodeProcess` rather than trusting `OpenProcess` success — so an exited-but-still-handle-held process is correctly dead before creation time is ever consulted. No interaction bug. The two Windows probes agree on access right, which also keeps their failure modes aligned.

## Smaller notes, none blocking

⚠ **`process_create_time` is public surface.** Under the 1.x no-removal contract a public name in a shipped module is a permanent commitment. It is a storage-internal helper with one consumer; `_process_create_time` would cost nothing today and reserve nothing forever. Your call — I will take it either way.

⚠ **The `actual is None` branch is the one untested path.** Your nine tests cover the recycled case, the legacy no-`create_time` case, roundtrip and lock reclaim, which is the right coverage. The permission-denied fallback is hard to provoke deliberately — given the measurement above it may be unreachable on a normal Windows box — but it is the branch that decides "treat as live", so a monkeypatched `process_create_time -> None` test would pin the intent cheaply.

⚠ **`_CREATE_TIME_TOLERANCE_S = 2.0`** is justified in the comment and I agree with the reasoning: both sides come from the same function and the same OS constant, so a true match is near-exact and a recycled PID misses by minutes at least. No change wanted; noting that the *reason* is what makes 2.0 safe rather than arbitrary, and the comment says so.

⚠ Removing the `watcher._is_pid_alive` wrapper is right — it bypassed the identity check and had no production caller, so leaving it would have been a working path back to the bug.

**Verdict: approve on the code.** The only thing standing between this and merge is the `CHANGELOG.md` seam, which is unrelated to anything you wrote.

phantom-man:
Both suggestions taken in `4131398`:

- **`_process_create_time`** — you're right that the 1.x contract makes the free choice today a permanent one tomorrow; it's storage-internal with one consumer, so it goes private now. Also taking your point about NOT DRYing it with `runtime_identity._linux_process_start` — the comment there and here both state their epoch choice, and keeping them separate is what keeps those reasons attached to the code they govern.
- **The fallback branch is pinned** — `test_unreadable_create_time_treats_alive_pid_as_live` monkeypatches the reader to `None` and asserts an alive PID with a recorded creation time still resolves live. Your measurement is quoted in spirit in its docstring: on a normal box `PROCESS_QUERY_LIMITED_INFORMATION` makes the real condition nearly unreachable, which is exactly why the intent deserves a test rather than waiting for nature to provide one.

Thank you for going looking for the hole — an approve that includes "here's the refutation I attempted and why it failed" is worth more than the approve alone. 94 tests green locally; the changelog seam was fixed in `d737648` just before your review posted, so as far as I can tell everything between this and merge is now CI's to say.

jgravelle:
Merged — your commits are in `main` (`4131398` is an ancestor), and this PR shows **Merged** rather than closed.

⚠ **It landed via [#464](https://github.com/jgravelle/jcodemunch-mcp/pull/464) rather than this PR's merge button, and the reason is worth knowing for next time.** This PR went `CONFLICTING` because of *our* merges — #462 and #463 both touched `CHANGELOG.md`, and #462 added an `[Unreleased]` entry in exactly the region yours occupies. A conflicting PR has no `refs/pull/N/merge`, so CI could not run and the merge button was unavailable.

The normal fix is for a maintainer to push the resolution to your branch. **That is impossible here: `Nexusmill` is an Organization, and GitHub's "Allow edits by maintainers" does not apply to org-owned forks.** The PR reports `maintainerCanModify: true` and the push still returns 403 with full `repo` scope — I tried. So the choice was to send you back to that seam a third time or land it from our side, and after two rounds there I was not going to ask again for something our merge cadence caused.

**What I changed:** `CHANGELOG.md` only, all code merged clean. Your `### Process liveness…` section now sits inside `[Unreleased]` alongside the #461 entry; `## [Unreleased]` is untitled because one title cannot describe three unrelated fixes. Built programmatically rather than by hand — hand-editing that seam is what failed twice, and I was not going to make it three.

Verified with the #463 gate's own predicates rather than a line-anchored regex, which is the specific mistake I made reviewing your earlier fix. No glued headings, no duplicates. Full suite **7788 passed, 9 skipped, 0 failed**, ruff clean, reconciled by same-tree collect (7797 total, 7787 without `test_pid_reuse_identity.py` = its 10).

Thanks for both review notes landing before it merged — private `_process_create_time` and the monkeypatched fallback test. And thank you for the boot-relative `/proc` starttime reasoning; that comment is going to save someone a very confusing bug report after a VM restore.

Issue #453, opened 2026-08-12
TestWindowsUNCPathSafety reaches the real network for a fake UNC path: two release-blocking flakes in one day, and it hides a real gap


jgravelle:
### The one thing that is wrong

`tests/test_tools.py::TestWindowsUNCPathSafety` builds a fake UNC path
(`\server\share\repo`) and patches `Path.resolve` / `Path.exists` / `Path.is_dir`
on the `index_folder` module — but **something downstream still makes a real
filesystem call**, so Windows tries to reach `\server\share` over the network. When
the runner's network says no, the test fails with `WinError 64`.

### Why this is worth an issue rather than a shrug

**It has blocked two releases today, on two different Python versions.**

| release | job | result |
| --- | --- | --- |
| v1.108.272 | `windows-latest / 3.10` | `OSError: [WinError 64] The specified network name is no longer available: '\server\share\repo'` |
| v1.108.275 | `windows-latest / 3.11` | identical |

Both passed on re-run of the identical SHA, and in both runs the same test passed on
the other three Windows cells. So it is not version-specific — it is whichever
runner happens to have unlucky network conditions when that call goes out.

Each occurrence costs a full CI cycle plus the diagnosis to prove it is unrelated to
the change under test. That diagnosis is not free and it is not skippable: on
v1.108.274 a red `lint` job looked exactly like the [#259-#262 lint
disaster](https://github.com/jgravelle/jcodemunch-mcp/blob/main/CHANGELOG.md) and was
actually a cpython download failure, so "it is probably the usual flake" is not a
safe read. **A test that cries wolf twice a day degrades the signal that protects the
irreversible release steps.**

### The worse half: it currently passes for the wrong reason, and that hid a real gap

`test_unc_share_root_remains_too_broad` — the sibling in the same class — patches
`Path.exists` but **not** `os.path.exists`. On a branch where the broad-root guard
grows an `os.path.exists(path / ".git")` probe, that test's verdict is decided by
whether the runner happens to have `\server\share\.git`, not by the guard.

That is not hypothetical. Reviewing [PR #439](https://github.com/jgravelle/jcodemunch-mcp/pull/439)
I advised replacing a `not drive.startswith("\\\\")` clause with
`_path_safety_part_count(path) == 2`, believing the helper subsumed the UNC case. It
does not — a UNC **share root** has one part and the helper adds one for the anchor,
so it computes to exactly `2`:

```
path                     drive               parts   safety_count
[...]

PR #454, opened 2026-08-13
feat(parser): register markdown and html; markdown sections become container symbols

### LINKED ISSUE #452: Markdown/HTML files are not indexable as first-class sources - doc reads cost whole-file tokens
## Verdict to judge

Should jCodeMunch register `markdown` and `html` as first-class languages, with markdown sections extracted as container symbols?

## Problem

Documentation is where agents burn tokens. A repo's operating docs (state files, design bibles, runbooks) are often the largest files in the tree, and today jCodeMunch cannot index `.md`/`.html` at all - `search_text` misses them entirely unless they're proxied through an unrelated grammar via `extra_extensions`, and there is no symbol-level access, so an agent that needs one section of a large doc must read the whole file.

Real numbers from our deployment (Windows 11, tree-sitter-language-pack already bundled):

- Our main state doc is **274 KB**. With markdown sections as symbols, `get_file_outline` lists its sections and `get_symbol_source` returns a single section for **~400 tokens** instead of a whole-file read.
- Enabling markdown+html across one production repo took its symbol count from **5,153 to 7,563** - every one of those new symbols is a doc section an agent can now retrieve surgically.
- This is exactly the economics your own `TOKEN_SAVINGS.md` sells for code; docs are the same win, arguably bigger because doc files are bigger.

## Proposed mechanism (implementation in hand, PR ready)

- `MARKDOWN_SPEC`: tree-sitter markdown block grammar's nested `section` nodes (heading through content) become container symbols, so one section round-trips through `get_symbol_source`. Section names come from the `atx_heading > inline` text (plain children, not named fields - needs a small `_extract_name` special case).
- `HTML_SPEC`: registers `.html`/`.htm` on the same bundled grammar `RAZOR_SPEC` already rides, making html files indexable/text-searchable.
- No new dependencies - both grammars ship in tree-sitter-language-pack today.

Tests included (fixture + name-extraction + registry coverage), plus updates to the coverage/fixture tests that assumed markdown is unparseable. No new failures against the full suite on Windows at v1.108.274 (a handful of environment-dependent failures pre-exist there without this change; can file separately).

Happy to split anything at triage per the one-issue-one-verdict policy.

phantom-man:
Closes #452.

## What

Three commits:

1. **feat(languages): register markdown (section symbols) and html** — adds `.md`/`.markdown`/`.html`/`.htm` to `LANGUAGE_EXTENSIONS`, plus `MARKDOWN_SPEC` (nested `section` nodes → container symbols, verified against tree-sitter-language-pack's block grammar) and `HTML_SPEC` (same bundled grammar `RAZOR_SPEC` already rides).
2. **feat(extractor): markdown section names from atx heading inline text** — the heading text lives in `atx_heading > inline` as plain children, not named ts fields, so `_extract_name` gets a small markdown special case. Heading-less preamble sections produce no symbol by design.
3. **test: update expectations for markdown/html as indexed languages** — markdown/html join the data-format exemption in the refactor-pattern coverage tests (document structure, no import/def syntax), and three fixtures that used README.md as "a file the parser ignores" switch to files that still are ignored (comment-only `.py`, extensionless `LICENSE`).

## Why

Docs are the biggest files agents touch and today they're invisible to the index. With sections as symbols, `get_file_outline` on a 274 KB operating doc lists its sections, and `get_symbol_source` returns one section for ~400 tokens instead of a whole-file read. Enabling this across one production repo: 5,153 → 7,563 symbols. Same token economics as TOKEN_SAVINGS.md, applied to documentation.

## Tests

`tests/test_markdown.py` (5 tests: extension routing, registry specs, section symbols from a fixture, heading-name extraction, preamble exclusion) + `tests/fixtures/markdown/sample.md`. Full suite run with `PYTHONPATH=src` on Windows 11 against v1.108.274: **13 failed, 7688 passed, 30 skipped**. The 13 failures are environment-dependent and **pre-exist on pristine `upstream/main` in this environment** (verified A/B on the same machine — identical set with the feature commits absent; e.g. a temp-path case-sensitivity assertion `pytest-of-user` vs `pytest-of-User`). Happy to file those separately per one-issue-one-verdict.

## Notes

- No new dependencies; both grammars ship in tree-sitter-language-pack already.
- Deployed in production on our side since 2026-08-12 (watcher reindex included) — the numbers above are measured, not projected.

### DIFF
--- a/src/jcodemunch_mcp/parser/extractor.py
+++ b/src/jcodemunch_mcp/parser/extractor.py
@@ -928,6 +928,18 @@ def _extract_name(node, spec: LanguageSpec, source_bytes: bytes) -> Optional[str
                             return source_bytes[name_node.start_byte:name_node.end_byte].decode("utf-8")
         return None
 
+    # Markdown sections: the heading text lives in atx_heading > inline,
+    # which tree-sitter exposes as plain children, not named fields.
+    if spec.ts_language == "markdown" and node.type == "section":
+        for child in node.children:
+            if child.type == "atx_heading":
+                for sub in child.children:
+                    if sub.type == "inline":
+                        text = source_bytes[sub.start_byte:sub.end_byte].decode("utf-8").strip()
+                        return text or None
+                break
+        return None  # heading-less section (e.g. preamble) -> not a symbol
+
     if node.type not in spec.name_fields:
         return None
     
--- a/src/jcodemunch_mcp/parser/languages.py
+++ b/src/jcodemunch_mcp/parser/languages.py
@@ -247,6 +247,10 @@ LANGUAGE_EXTENSIONS = {
     ".swagger.yaml": "openapi",
     ".swagger.yml": "openapi",
     ".swagger.json": "openapi",
+    ".md": "markdown",
+    ".markdown": "markdown",
+    ".html": "html",
+    ".htm": "html",
 }
 
 
@@ -1946,6 +1950,35 @@ DLANG_SPEC = LanguageSpec(
     type_patterns=[],
 )
 
+MARKDOWN_SPEC = LanguageSpec(
+    ts_language="markdown",
+    # Nested `section` nodes span a heading through its content (verified
+    # against tree-sitter-language-pack's block grammar 2026-08-12). Each
+    # becomes a container symbol so get_symbol_source returns one section.
+    symbol_node_types={"section": "class"},
+    name_fields={},  # heading text has no ts field; see _extract_name special case
+    param_fields={},
+    return_type_fields={},
+    docstring_strategy="preceding_comment",
+    decorator_node_type=None,
+    container_node_types=["section"],
+    constant_patterns=[],
+    type_patterns=[],
+)
+
+HTML_SPEC = LanguageSpec(
+    ts_language="html",   # same bundled grammar RAZOR_SPEC already rides
+    symbol_node_types={},
+    name_fields={},
+    param_fields={},
+    return_type_fields={},
+    docstring_strategy="preceding_comment",
+    decorator_node_type=None,
+    container_node_types=[],
+    constant_patterns=[],
+    type_patterns=[],
+)
+
 
 # Language registry
 LANGUAGE_REGISTRY = {
@@ -2024,6 +2057,8 @@ LANGUAGE_REGISTRY = {
     "nim": NIM_SPEC,
     "tcl": TCL_SPEC,
     "dlang": DLANG_SPEC,
+    "markdown": MARKDOWN_SPEC,
+    "html": HTML_SPEC,
 }
 
 # Template-engine languages (jinja/twig) all route through the
--- a/tests/fixtures/markdown/sample.md
+++ b/tests/fixtures/markdown/sample.md
@@ -1,0 +1,16 @@
+# Alpha Top
+
+intro text under alpha
+
+## Beta Sub
+
+beta body line one
+beta body line two
+
+## Gamma Sub
+
+gamma body
+
+# Delta Top
+
+delta body
--- a/tests/test_markdown.py
+++ b/tests/test_markdown.py
@@ -1,0 +1,25 @@
+"""Markdown + HTML language registration and section-symbol extraction."""
+from pathlib import Path
+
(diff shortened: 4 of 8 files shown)


jgravelle:
**Our messages crossed** — I posted the triage verdict on #452 at 01:11 UTC and this landed at 01:43, so you were almost certainly mid-flight. Reading [that comment](https://github.com/jgravelle/jcodemunch-mcp/issues/452#issuecomment-5274712045) first will save you re-reading the argument here; the short version is that **the HTML half is accepted and I want it, and the markdown half is declined.**

I am not going to restate the whole triage. But this PR adds a piece of evidence I did not have when I wrote it, and it is the strongest one yet — **against the markdown half, and it comes from your own diff.**

## Your diff is the blast-radius argument

Three test files outside the feature had to change, and one of the edits removes a shipped invariant:

```diff
-    assert "README.md" not in files  # Not a source file
```

That is `discover_source_files` in `test_tools.py` — the **GitHub repo indexer**. Deleting it does not just make docs retrievable; it **changes what the corpus is** for every indexed repo, on the next reindex, with no opt-in. Index size, symbol counts, token costs and every derived metric move at once, for users who asked for none of it.

The other two are the same shape:

- `test_suggest_queries.py` — a fixture had to stop using `README.md` as "a file the parser ignores", with the comment *"markdown headings ARE symbols now"*.
- `test_plan_refactoring.py` — `markdown` and `html` had to join the refactor-pattern exemption set.

**Three unrelated tests broke because they encoded an assumption the whole codebase shares: a symbol is code.** That is exactly the invisible blast radius the triage predicted, and your patch found it empirically in an afternoon. The consumers I named there — `find_dead_code` (`find_dead_code.py:316` iterates `index.symbols` with no kind filter, and nothing imports a `.md`, so all 2,410 new symbols land in dead code), the `dead_code` radar axis, and the public observatory grades — are not covered by any of those three edits and are still unaddressed. A well-documented repo would get a worse public grade.

⚠ **None of that says your patch is wrong.** It says doc sections are a different kind of thing from code symbols, and jdocmunch already models them as one: `index_local`, then `search_sections` / `get_document_outline` / `get_section`. Your 274 KB doc is exactly its use case. **If it is not delivering that for you, file it there with these same numbers** — that report is worth more to the suite than this feature is.

## What I would like

**Split the HTML half into its own PR and I will review it on arrival.** Reasons it stands on its own, including one stronger than the report made:

`flow_edges._resolve_template` (`flow_edges.py:225`) resolves a rendered template to a file **only when that file is indexed**. Since `.html` is not indexable today, that resolution silently returns `None` on every Django, Flask, Express and Rails repo we touch — the `views` annotation on `get_signal_chains` and the render edges in `get_endpoint_impact` are degraded, and nobody noticed because the failure is silent.

So please include:

- `.html` / `.htm` on the existing html grammar, `symbol_node_types` empty so it stays a text-searchable file class emitting no symbols.
- **A test proving `_resolve_template` now resolves a template it returns `None` for today.** That is the test that makes the change worth having — registry-coverage assertions would pass just as happily against a version that resolves nothing.
- A note on the dead-file interaction (an indexed `.html` with no importers is still a dead *file* under the current rule; teaching `find_dead_code` about render edges is a separate issue).

Rebase onto current `main` when you do — `tests/test_tools.py` moved under it about an hour ago (#453).

⚠ **On the 13 failures: thank you for A/B-ing them against pristine `upstream/main` on the same machine rather than reporting a raw count.** That is the right method and it is not the common one. Please do file them — separately, per one-issue-one-verdict. A case-sensitivity assertion on `pytest-of-user` vs `pytest-of-User` is a real test-isolation bug on our side, and it is the same family as #411.

I will keep #452 open until the HTML PR exists, then close it there. The timebox from the triage stands: **if I have not seen it by 2026-08-26 I will implement the HTML half myself and credit you** in the CHANGELOG and release notes.

phantom-man:
Superseded by #459 per the triage on #452: the accepted HTML half is there, re-scoped exactly as requested (zero-symbol file class, the `_resolve_template` red/green test, dead-file note in the spec comment, rebased past #453 onto v1.108.275). The markdown half stays declined and I'm not contesting it - the dead-code blast-radius argument held up against our own diff. Closing this one.

Issue #440, opened 2026-08-12
`identity_hit` is always 0 for `search_symbols` rows in `ranking_events`


rknighton:
### The one thing that is wrong

`ranking_events.identity_hit` is recorded as 0 on every `search_symbols` row, including
exact symbol-name matches, because the list passed to `extract_ledger_features` carries
only scores.

### Why this matters

`identity_hit` is meant to record whether any of the top three results matched the query
by name, counting an exact match, a name prefix, or an ID segment
(`_identity_score`, `tools/search_symbols.py:367-407`). On the default and semantic search
paths that value is always 0, whatever the identity channel found.

Two shipped features read the column.

`analyze_perf` reports an `identity_hits` count per repository. Rows written by
`search_symbols` never contribute to it, so it undercounts by however many name matches
those searches made. Fusion rows still count correctly, so the number is not obviously
broken from the outside.

`regret` has a signal called vocabulary gap, meaning the agent's wording did not match a
symbol name but semantic search found the right thing anyway. It looks for rows where
`identity_hit` is 0 and `semantic_used` is 1. On the semantic search path `semantic_used`
is passed as `True` (`:1607`) and `identity_hit` is always 0, so both halves are satisfied
by construction rather than by measurement. The signal does not fire on a single row: it
also requires confidence at or above 0.30 and the same query hash appearing at least twice
(`VOCAB_CONF_FLOOR` and `VOCAB_RECUR`, `retrieval/regret.py:49-50`). But for any repeated
query above that floor, the condition it is meant to test is not actually being tested. Per
that function's docstring these clusters feed `suggest_corrections`, which shows
configuration patches to the user.

There is already a module for this class of problem. `retrieval/ledger_trust.py` marks rows
whose column holds a default rather than a measurement, so consumers can set them aside.
`identity_label_is_trustworthy` returns True for any tool other than
`get_ranked_context_fusion` (`:105-106`), so these rows are treated as genuine measurements.

### Reproduction

`PYTHONPATH=src python repro_identity_hit.py`

```
default      top result = calculate_invoice_total
fusion=True  top result = calculate_invoice_total

ranking_events:
  tool=search_symbols           identity_hit=0  top1_score=51.73
  tool=search_symbols_fusion    identity_hit=1  top1_score=0.06
```

Same repo, same query, same top result. The query is the exact symbol name.

The script in full:

```python
"""identity_hit is recorded False for an exact name match on the default path."""
import os, sqlite3, tempfile
[...]

PR #457, opened 2026-08-13
fix(telemetry): record the true result count (#441), stop reopening the db per event (#442)

### LINKED ISSUE #441: `returned_ids` is capped at 50 with no record of the true result count
### The one thing that is wrong

`ranking_events.returned_ids` keeps the first 50 IDs and the row carries nothing that says
how many there actually were.

### Why this matters

A row holding 50 IDs could mean the search returned exactly 50. It could also mean it
returned 500 and the rest were dropped. Nothing on the row distinguishes those two cases,
so an analysis cannot exclude the truncated rows or say how many it dropped.

Being precise about what is and is not lost. The cap keeps the first 50 in order
(`storage/token_tracker.py:786`), so rank positions inside the stored prefix are intact.
Anything answerable from the top 50 is unaffected. What cannot be answered is anything
whose denominator is the full returned set: how large the result set actually was, how
often truncation happened at all, and what the distribution looks like beyond position 50.

The blast radius today is analysis, not runtime. No shipped feature depends on the exact
count. `regret` reads `returned_ids` only to ask whether it is empty or holds more than one
entry (`retrieval/regret.py:122` and `:138-144`), and `ledger_trust` only whether it is
empty (`:112-115`). Truncation begins above 50, so none of those are reached.

The truncation itself is documented: `SECURITY.md:274-275` describes stored ranking events
as carrying a "truncated result-id list." What is missing is any marking on the individual
row, so a reader of the table cannot tell which rows were affected.

A case from working with this data. Across 131 recorded searches, 6 had stored lists of
exactly 50 IDs, and nothing in the table says whether any of them returned more.

In [Discussion #430](https://github.com/jgravelle/jcodemunch-mcp/discussions/430) I called
those six "hit the 50-item cap", asserting a truncation I had only inferred from the
length. That is the bug in miniature, and catching it on re-verification is why this is
filed now rather than with that comment.

### Reproduction

`PYTHONPATH=src python repro_returned_ids.py`

```
results returned to caller : 60
ids stored in ranking_events: 50
a column recording the true count: no
columns: ts, repo, tool, query_hash, query, returned_ids, top1_score, top2_score, confidence, semantic_used, identity_hit, repo_is_stale
```

This goes through the shipped `search_symbols` path rather than calling the storage helper,
so the cap is shown to be reachable in normal use. `max_results` is capped at 100
(`tools/search_symbols.py:665`), and token-budget mode can raise the effective limit
further (`:851`), so results above 50 are ordinary rather than contrived.

The script in full:

```python
"""A real search returning more than 50 results stores only 50, with no true count."""
import json, os, sqlite3, tempfile
from pathlib import Path

os.environ["JCODEMUNCH_PERF_TELEMETRY"] = "1"

from jcodemunch_mcp.tools.search_symbols import search_symbols

with tempfile.TemporaryDirectory(prefix="jcm-retention-repro-") as tmp:
    root = Path(tmp)
    project = root / "sample"
    project.mkdir()

    # 60 functions that all match the query term "invoice".
    body = "\n\n".join(
        f"def invoice_handler_{i:03d}(x):\n    return x + {i}"
        for i in range(60)
    )
    (project / "invoices.py").write_text(body + "\n", encoding="utf-8")
    store = str(root / "store")

    repo = index_folder(path=str(project), use_ai_summaries=False,
                        storage_path=store)["repo"]

    out = search_symbols(repo=repo, query="invoice", storage_path=store,
                         max_results=60)
    returned = len(out["results"])

    conn = sqlite3.connect(str(Path(store) / "telemetry.db"))
    try:
        columns = [r[1] for r in conn.execute("PRAGMA table_info(ranking_events)")]
        stored = json.loads(
            conn.execute("SELECT returned_ids FROM ranking_events").fetchone()[0]
        )
    finally:
        conn.close()

    print(f"results returned to caller : {returned}")
    print(f"ids stored in ranking_events: {len(stored)}")
    print(f"a column recording the true count: "
          f"{'yes' if 'returned_count' in columns else 'no'}")
    print(f"columns: {', '.join(columns)}")
```

### Expected vs actual

Actual: 60 results reach the caller, the cap at `storage/token_tracker.py:786`
(`list(returned_ids)[:50]`) stores 50, and no other column records the original length.

Expected: some way to tell a truncated row from a complete one. Keeping the bounded payload
is fine. This does not need more IDs retained, only the count.

### Environment

- jCodeMunch version: 1.108.271 (`main` at `4bbf5c5`, and release v1.108.271 at `8922e69`)
- Python version: 3.13.7
- OS: Windows 11

Every `file.py:line` reference in this issue is against
[`4bbf5c5`](https://github.com/jgravelle/jcodemunch-mcp/tree/4bbf5c509a310e461ebe7dd97efb7a944b5c164a).

### Anything else you found along the way

Nothing beyond the two filed separately from the same write path.

### LINKED ISSUE #442: Opt-in ranking telemetry: ~80% of each write is unnecessary repeated database setup
### The one thing that is wrong

With `perf_telemetry_enabled` on, each `record_ranking_event` call opens a new SQLite
connection and executes all eight schema statements before writing its single row.

### Why this matters

Scope first, because it bounds everything below: this only reaches installs running with
`perf_telemetry_enabled`, which is off by default.

On those installs, recording one search costs about 4 ms, and roughly 3.3 ms of that is the
reopen and schema replay. Every event opens a fresh database connection and re-runs all
eight table and index creation statements before inserting one row. All eight are
`IF NOT EXISTS`, so after the first event none of them does anything productive.

Measured on the same machine, in the same process, with the two arms interleaved:

| Cost of recording one event | median | p95 |
| --- | ---: | ---: |
| as written today | 3.99 ms | 4.83 ms |
| connection reused, everything else identical | 0.74 ms | 0.97 ms |

So about 82% of the cost of recording a search is setup that is discarded immediately.

Both arms call the real `record_ranking_event`, so the configuration lookup, lock
acquisition, path resolution, query hashing, JSON serialization, numeric coercion and error
handling are identical and cancel out. The only variable is whether
`_ensure_perf_db_locked` opens a connection and replays the schema or hands back one that
is already open.

It can also delay concurrent `_State` operations. The write holds `self._lock`
(`storage/token_tracker.py:766-767`), the single lock created at `:91` and used by the
other public `_State` operations, including `record_latency` at `:538`, whose docstring
says it is "Called from server.call_tool". Searches are dispatched through `asyncio.to_thread`
(`server.py:5624`), so overlap is possible. I did not measure contention, so this is a
property of the code rather than something observed in the benchmark, which is
single-threaded.

On volume, this is just arithmetic on the median above: 100 recorded searches is about
0.33 seconds of avoidable work, and 1,000 is about 3.3 seconds. I have no data on how many
searches a typical session makes, so those are conversions rather than a claim about anyone's
usage.

### Reproduction

`PYTHONPATH=src python repro_schema_per_event.py`

```
500 events per arm, both through record_ranking_event.

Arm A opened 500 connections and ran 4000 schema statements
  (arm B opened 0 and ran 0 after its first event)

Cost of one event                 median       p95
  as written today                3.99ms     4.83ms
  connection reused               0.74ms     0.97ms

  difference                      3.26ms (82% of the write)
```

Three runs gave 3.48, 3.23 and 3.26 ms, at 82%, 81% and 82%. The two counters are exact and
identical every run. The timings are from one Windows laptop and I have not run this on
Linux or macOS, so both the milliseconds and the ratio are this machine's numbers rather
than portable ones. Individual maximums swing widely in both arms, so nothing here says
anything about the tail.

On how it isolates the variable: it does not reimplement the write. Both arms call the real
`record_ranking_event`, and arm B patches `_ensure_perf_db_locked` to return an already-open
connection whose `close()` is a no-op. Everything else in the method runs identically in
both arms. The two are interleaved so any drift lands on both.

The script in full:

```python
"""Every ranking event reopens the telemetry db and replays the whole schema.

Both arms call the real record_ranking_event, so every other per-event cost
(config lookup, lock, path resolution, SHA-1 hash, JSON serialization, numeric
coercion, error handling) is identical and cancels out. The ONLY difference is
that arm B's _ensure_perf_db_locked hands back an already-open connection
instead of opening one and replaying the schema.
"""
import os, statistics, tempfile, time
from pathlib import Path

os.environ["JCODEMUNCH_PERF_TELEMETRY"] = "1"

from jcodemunch_mcp.storage import token_tracker

WARMUP, N = 20, 500

counts = {"opens": 0, "schema": 0}
real_connect = token_tracker.sqlite3.connect


def counting_connect(*args, **kwargs):
    counts["opens"] += 1
    conn = real_connect(*args, **kwargs)
    conn.set_trace_callback(
        lambda sql: counts.__setitem__("schema", counts["schema"] + 1)
        if sql.lstrip().upper().startswith(("CREATE TABLE", "CREATE INDEX", "ALTER TABLE"))
        else None
    )
    return conn


token_tracker.sqlite3.connect = counting_connect


class _KeepOpen:
    """Passes calls through but ignores close(), so the caller can reuse it."""

    def __init__(self, conn):
        self._conn = conn

    def execute(self, *args, **kwargs):
        return self._conn.execute(*args, **kwargs)

    def close(self):
        pass


reuse = {"on": False}
cache = {}
real_ensure = token_tracker._State._ensure_perf_db_locked


def patched_ensure(self, base_path=None):
    if not reuse["on"]:
        return real_ensure(self, base_path)
    if base_path not in cache:
        cache[base_path] = real_ensure(self, base_path)
    return _KeepOpen(cache[base_path])


token_tracker._State._ensure_perf_db_locked = patched_ensure


def stats(samples):
    s = sorted(samples)
    return statistics.median(s) / 1e6, s[int(0.95 * len(s)) - 1] / 1e6


with tempfile.TemporaryDirectory(prefix="jcm-schema-repro-") as tmp:
    root = Path(tmp)
    store_a = str(root / "a")
    store_b = str(root / "b")
    state = token_tracker._State()

    def write(store, i):
        state.record_ranking_event(tool="benchmark", repo="local/demo", query=f"q-{i}",
                                   returned_ids=["a", "b", "c"], top1_score=3.0,
                                   top2_score=2.0, confidence=1.0, identity_hit=True,
                                   base_path=store)

    for i in range(WARMUP):
        reuse["on"] = False
        write(store_a, i)
        reuse["on"] = True
        write(store_b, i)
    before = dict(counts)

    # Interleaved so any drift lands on both arms rather than one.
    a_samples, b_samples = [], []
    for i in range(N):
        reuse["on"] = False
        t0 = time.perf_counter_ns()
        write(store_a, WARMUP + i)
        a_samples.append(time.perf_counter_ns() - t0)

        reuse["on"] = True
        t0 = time.perf_counter_ns()
        write(store_b, WARMUP + i)
        b_samples.append(time.perf_counter_ns() - t0)

    for conn in cache.values():
        conn.close()

    opens = counts["opens"] - before["opens"]
    schema = counts["schema"] - before["schema"]
    a_med, a_p95 = stats(a_samples)
    b_med, b_p95 = stats(b_samples)

    print(f"{N} events per arm, both through record_ranking_event.\n")
    print(f"Arm A opened {opens} connections and ran {schema} schema statements")
    print(f"  (arm B opened 0 and ran 0 after its first event)\n")
    print(f"Cost of one event              {'median':>9} {'p95':>9}")
    print(f"  as written today            {a_med:8.2f}ms {a_p95:8.2f}ms")
    print(f"  connection reused           {b_med:8.2f}ms {b_p95:8.2f}ms")
    print()
    print(f"  difference                  {a_med - b_med:8.2f}ms "
          f"({(a_med - b_med) / a_med * 100:.0f}% of the write)")
```

### Expected vs actual

Actual: `record_ranking_event` calls `_ensure_perf_db_locked` per event
(`storage/token_tracker.py:767`) and closes the connection in its `finally` (`:795-799`).
`_ensure_perf_db_locked` connects at `:619` and then runs three `CREATE TABLE IF NOT EXISTS`
and five `CREATE INDEX IF NOT EXISTS` statements every time (`:621-664`). That is the eight
per event the counter shows.

Expected: recording an event does not repeat one-time setup work on every event.

I am deliberately not prescribing how. Holding a connection open has its own constraints
here that I have not worked through: the connection is created with the default
`check_same_thread=True`, `base_path` can select a different database per call, and the
lifecycle would need to survive the failure handling at `:666-672` that intentionally keeps
one unwritable caller-supplied path from disabling telemetry everywhere.

### Environment

- jCodeMunch version: 1.108.271 (`main` at `4bbf5c5`, and release v1.108.271 at `8922e69`)
- Python version: 3.13.7
- OS: Windows 11

Every `file.py:line` reference in this issue is against
[`4bbf5c5`](https://github.com/jgravelle/jcodemunch-mcp/tree/4bbf5c509a310e461ebe7dd97efb7a944b5c164a).

### Anything else you found along the way

Nothing beyond the two filed separately from the same write path.

jgravelle:
Fixes #441. Fixes #442. Both from @rknighton's batch against the same write path .272 touched. No version bump; rides `[Unreleased]`.

## #441 — a stored list of 50 could mean 50 or 500

`returned_ids` keeps the first 50 ids and the row carried nothing saying how many there really were, so **a complete result set and a truncated one were byte-identical in every stored field**. Rows now carry `returned_count`, the true size recorded before the cap. The id list stays bounded; only the count is added.

Against your repro script, unmodified:

```
results returned to caller : 60
ids stored in ranking_events: 50
a column recording the true count: yes
returned_count value        : 60
```

⚠⚠ **Pre-existing rows keep `NULL`, meaning UNKNOWN, and nothing backfills them.** Reading `count == len(returned_ids)` off an old row is the defect itself — the same inference you made in Discussion #430 and caught on re-verification. Same treatment `ledger_trust` gives its unseparable history. `0` and `NULL` are pinned as distinct, because `0` is a measurement and `NULL` is an absence.

⚠ Your severity read was checked and holds: this is analysis-only. `regret` reads `returned_ids` for emptiness and `> 1` (`regret.py:122`, `:138-144`), `ledger_trust` for emptiness (`:112-115`), and truncation starts above 50, so none is reached.

## #442 — and the measurement that changed the design

**Your report made me measure the safe fix first, and that is what chose the risky one.** There is an obvious low-risk shape here: remember the schema is ready, skip the eight `IF NOT EXISTS` statements, keep the per-write open/close. No connection lifetime to manage — none of the constraints you flagged as unresolved. Three arms, 400 interleaved events:

| arm | median |
|---|---:|
| connect + `PRAGMA` + 8 DDL + insert | 15.441ms |
| connect + insert, schema ensured once | 15.166ms |
| cached connection + insert | 3.214ms |

**Skipping the DDL captures 2% of the available saving.** The open/close is the other 98%. So the cheap fix does essentially nothing and caching the connection is the only thing that works.

⚠ **That also corrects the framing in the issue, and it is the useful half.** You attributed the cost to "reopen and schema replay" jointly; only the reopen carries it. Eight DDL statements *look* like the expense next to one `connect()` and are noise. Worth knowing before anyone reaches for the tempting version.

Shipped path measures **3.455ms against a 16.615ms pre-fix baseline, a 79% cut** — agreeing with your 82%. (My absolutes are ~4x yours on a slower box; the ratio is what transfers, which is the right way to read both.)

## The constraints you left open

You said you were deliberately not prescribing how, and listed three. All three are real:

- **`check_same_thread`** — `False` is *required*, because searches dispatch through `asyncio.to_thread` and a cached connection outlives its opening thread. It is *safe* because every caller holds `_State._lock`, so the serialisation the flag would enforce is already there. That reasoning is recorded at the call site, since a future edit could quietly invalidate it.
- **Per-call `base_path`** — the cache is keyed by resolved path. A single connection would send one store's rows to another, which is the defect .188 fixed.
- **The `:666-672` failure handling** — `_perf_db_failed` still latches for the default path only, so one unwritable caller-supplied path cannot disable telemetry process-wide.

## Two silent failure modes caching introduces, which the report did not name

The first surfaced because my benchmark crashed, not because I designed for it.

⚠⚠ **A stray `close()` poisons the cache.** The old contract had callers closing after every write, so any missed call site leaves a dead handle that every later caller receives — telemetry off for the process while every write still reports success.

⚠⚠ **A deleted database file gets written into an unlinked inode.** SQLite keeps writing happily; rows land nowhere and nothing raises. Before caching, the next event simply recreated the file, so caching would have *introduced* this. A liveness probe cannot catch it — the connection is healthy, the *file* is gone — so the guard checks `exists()` as well.

Both checks together cost **0.344ms, 2.1% of the pre-fix write**, against the 79% saved.

⚠ **Windows cannot produce the orphan case at all** — it refuses to unlink a file with an open handle. So the end-to-end test is POSIX-only and says so in its skip reason, with a portable unit test covering the predicate. Flagging that rather than letting a green Windows run imply coverage it does not have.

`close_perf_dbs()` is public because the connections now outlive a write, and on Windows an open handle blocks removal of the directory holding the database. It is registered with `atexit` **before** `flush` on purpose: `atexit` runs LIFO, so registering it second would close the database out from under the final flush.

## Verification

- Full suite **7763 passed, 9 skipped, 0 failed**, `ruff check src/` clean, sync + rotation gates 37/37.
- Reconciled by same-tree collect: 7772 total, 7753 with `test_v1_108_276.py` ignored (= its 19). Nothing else moved.
- Non-vacuity by reverting **only the two call sites** on the same tree: **11 of 19 fail**. The 7 passing both sides are the schema, migration idempotence, the telemetry-disabled scope control, and the public surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,93 @@ design:
 Proven non-vacuous by removing the `is_file` patch: the guard fires and names the
 call, rather than the run going red somewhere else an hour later.
 
+### A stored list of 50 could mean 50 or 500 ([#441](https://github.com/jgravelle/jcodemunch-mcp/issues/441))
+
+`ranking_events.returned_ids` keeps the first 50 ids and the row carried nothing
+saying how many there really were, so **a complete result set and a truncated one
+were byte-identical in every stored field**. An analysis could neither exclude the
+truncated rows nor say how many it dropped.
+
+Rows now carry `returned_count`: the true size, recorded before the cap. The id
+list stays bounded — only the count is added.
+
+⚠⚠ **Pre-existing rows keep `NULL`, which means UNKNOWN and must never be read as
+"the count equals `len(returned_ids)`".** That inference is the defect, not the
+fix, and no heuristic backfills it — the same treatment `ledger_trust` gives its
+unseparable history. ⚠ `0` is a measurement and `NULL` is an absence; the test
+pins that they do not collide.
+
+⚠ **Blast radius is analysis, not runtime, and that was checked rather than
+assumed.** `regret` only asks whether `returned_ids` is empty or holds more than
+one entry (`regret.py:122`, `:138-144`) and `ledger_trust` only whether it is
+empty (`:112-115`) — truncation begins above 50, so none of them is reached.
+
+⚠ **@rknighton filed this against his own earlier claim.** In Discussion #430 he
+described six rows as having "hit the 50-item cap", an assertion he had inferred
+from the stored length rather than measured, and he caught it on re-verification.
+The defect and the mistake it invites are the same shape.
+
+### ~79% of a telemetry write was reopening the database ([#442](https://github.com/jgravelle/jcodemunch-mcp/issues/442))
+
+With `perf_telemetry_enabled` on — off by default, so this reaches opt-in installs
+only — every event opened a fresh connection and replayed all eight
+`IF NOT EXISTS` statements before inserting one row. Connections are now cached
+per resolved path for the process.
+
+⚠⚠ **The schema replay was NOT the expensive part, and measuring that is what
+killed the cheap fix.** The obvious low-risk change — remember the schema is ready
+and skip the DDL, needing no connection lifetime at all — was measured first,
+across 400 interleaved events:
+
+| arm | median |
+|---|---:|
+| connect + `PRAGMA` + 8 DDL + insert | 15.441ms |
+| connect + insert, schema ensured once | 15.166ms |
+| cached connection + insert | 3.214ms |
+
+Skipping the DDL captures **2%** of the available saving. The other 98% is the
+open/close itself, so caching the connection is the only thing that works. The
+shipped path measures **3.455ms against a 16.615ms pre-fix baseline, a 79%
+reduction**. (Absolute figures are this machine's and slower than the reporter's
+3.99ms/0.74ms; the ratio is what transfers, and it agrees with his 82%.)
+
+⚠ `check_same_thread=False` is **required**, because searches dispatch through
+`asyncio.to_thread` and a cached connection outlives the thread that opened it. It
+is **safe** because every caller holds `_State._lock` — the serialisation the flag
+would otherwise enforce is already there. That reasoning is recorded at the call
+site, because it is the thing a future edit could quietly invalidate.
+
+### Two ways a process-lifetime connection goes bad, both silent
+
+Neither was in the report; the first surfaced when the benchmark crashed.
+
+⚠⚠ **A closed cached connection poisoned the cache.** The old contract had callers
+closing after every write, so any missed call site — or third-party caller — left
+a dead handle that every later caller received. Telemetry would be off for the
+process while every write still reported success.
+
+⚠⚠ **A deleted database file would be written into the void.** SQLite keeps
+writing happily to an unlinked inode; rows land nowhere and nothing raises. Before
+caching, the next event simply recreated the file, so caching would have
+introduced this. ⚠ A liveness probe cannot catch it — the connection is perfectly
+healthy, it is the file that is gone — so the guard checks `exists()` too.
+⚠ **Windows cannot produce this case at all** (it refuses to unlink a file with an
+open handle), so the end-to-end test is POSIX-only and a portable unit test covers
+the predicate. Stating that rather than implying cross-platform coverage.
+
+Both checks together cost **0.344ms, 2.1% of the pre-fix write**, against the 79%
+saved. Paying 2 to keep 79, where the alternative failure is silent.
+
+`close_perf_dbs()` is public because the connections now outlive a write: on
+Windows an open handle blocks removal of the directory holding the database.
+⚠ It is registered with `atexit` **before** `flush` on purpose — `atexit` runs
+LIFO, so registering it second would close the database out from under the final
+flush.
+
+`tests/test_v1_108_276.py` (19, 1 skipped), **11 fail against the pre-fix call
+sites**; the 7 passing both sides are the schema, the migration's idempotence, the
+telemetry-disabled control and the public surface.
+
 ## [1.108.275] - 2026-08-12 - A pattern that matches nothing now says so
 
 ### `entry_point_patterns` failed silently ([#446](https://github.com/jgravelle/jcodemunch-mcp/issues/446))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,22 @@ flush.
 sites**; the 7 passing both sides are the schema, the migration's idempotence, the
 telemetry-disabled control and the public surface.
 
+⚠⚠ **Writing these tests exposed an unrelated ranking defect, filed as
+[#458](https://github.com/jgravelle/jcodemunch-mcp/issues/458).** The
+`Retrieval-quality gate` failed on this branch -- a telemetry-only change that
+cannot affect ranking -- because the replay harness indexes **this repo itself**,
+so a new test file changes the corpus. Isolated by removing the file from the same
+tree and re-indexing: `mrr 1.0` becomes `0.95` when it is present.
+
+The displacement is real: a fixture named `state` scored `identity_type: "exact"`
+for the query `_State`, identically to the class literally named `_State`, and won
+the tie on field length and having a docstring. **`identity_type` graded a
+normalised match as exact** -- the same shape as #440, a column reporting a grade
+it did not measure. ⚠ The fixture is renamed here so the gate passes; **that
+unblocks a PR and fixes nothing**, and the test file says so at the rename site.
+⚠ The gate's self-indexing sensitivity will produce false reds on ordinary
+test-adding PRs. It also caught this. Both are true; neither is addressed here.
+
 ## [1.108.275] - 2026-08-12 - A pattern that matches nothing now says so
 
 ### `entry_point_patterns` failed silently ([#446](https://github.com/jgravelle/jcodemunch-mcp/issues/446))
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -290,6 +290,41 @@ not measure is what that costs in practice.
 a RESUMED conversation counts accumulated context on every step, so the total is
 dominated by how much the agent read early on, which compounds.
 
+**In flight 2026-08-13: #441 + #442 (@rknighton), ranking-ledger write path.**
+Same path .272 touched. Unreleased; see CHANGELOG `[Unreleased]`.
+⚠⚠ **The reusable lesson is that measuring the SAFE fix first is what chose the
+risky one.** #442 has an obvious low-risk shape — remember the schema is ready,
+skip the eight `IF NOT EXISTS` statements, keep the per-write open/close, no
+connection lifetime to manage. **Measured: it captures 2% of the available
+saving.** The other 98% is the open/close itself. Had that not been measured, the
+cheap fix would have shipped, looked principled, and delivered nothing.
+⚠ Shipped path is **3.455ms vs 16.615ms**, a **79%** cut, agreeing with his 82%
+(absolutes are machine-local; the ratio transfers). ⚠ `check_same_thread=False` is
+REQUIRED (searches dispatch via `asyncio.to_thread`, so a cached connection
+outlives its opening thread) and SAFE only because every caller holds
+`_State._lock` — recorded at the call site because a future edit could quietly
+invalidate it.
+⚠⚠ **Caching a connection introduces TWO silent failure modes the report did not
+name, and the benchmark found the first by crashing**: a stray `close()` poisons
+the cache so every later caller gets a dead handle (telemetry off, every write
+still reporting success), and a DELETED db file gets written into an unlinked
+inode forever (pre-caching, the next event just recreated it). A liveness probe
+catches only the first; the `exists()` check is what catches the second. Together
+**0.344ms, 2.1%** of the pre-fix write. ⚠ **Windows cannot produce the orphan case
+at all** (it refuses to unlink a file with an open handle) — the end-to-end test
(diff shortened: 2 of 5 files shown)


jgravelle:
**The `Retrieval-quality gate` failed, and it was not infra — it caught a real defect that has nothing to do with this PR.** Pushed a fix for the collision; the defect itself is filed as #458 rather than papered over.

`mrr 0.95 vs 1.0`, `ndcg 0.9631 vs 1.0`, on a telemetry-only change that cannot affect ranking. Reproduced locally, then isolated by removing the new test file from the same tree and re-indexing:

| tree | mrr | ndcg |
|---|---:|---:|
| with `tests/test_v1_108_276.py` | 0.95 | 0.9631 |
| same tree, file removed | **1.00** | **1.00** |

The replay harness indexes **this repo itself**, so adding a test file changes the corpus. The displaced query was `_State`:

```
tests/test_v1_108_276.py::state#function | 58.749
   {'name': 8.012, 'signature': 7.389, 'summary': 5.992, 'identity': 50.0, 'identity_type': 'exact'}
token_tracker.py::_State#class           | 58.394
   {'name': 6.996, 'signature': 6.153, 'summary': 0.0,   'identity': 50.0, 'identity_type': 'exact'}
```

**A pytest fixture named `state` was graded `identity_type: "exact"` for the query `_State`** — identically to the class literally named `_State` — because the matcher normalises away case and the leading underscore. The identity channel then cannot separate them, and the tie falls to BM25, where the shorter name with a docstring wins by 0.355 points out of ~58.

⚠ Same shape as #440: a column reporting a grade it did not measure. `identity_type` also feeds `compute_confidence`, so this is a published number and not only a search result.

**What I changed here, and what I did not.** The fixture is renamed `state` → `tracker`, which makes the gate pass locally (`mrr 1.0, ndcg 1.0`). **That unblocks this PR and fixes nothing.** The rename site carries a comment pointing at #458 so nobody later reads it as the fix. I did not touch the golden baseline, and I did not loosen the gate — either would have buried a real regression.

⚠ Worth a separate decision, not taken here: the gate's self-indexing means **any PR that adds symbols can fail it**, which is a real tax on ordinary test-adding work. It also caught this. Both are true; #458 records the tension rather than resolving it.

PR #473, opened 2026-08-14
fix(telemetry): key the perf db connection cache on the resolved path (#465)

### LINKED ISSUE #465: A relative `storage_path` sends perf telemetry rows to another store's database

`_perf_db_path` never resolves the path it returns, so the connection cache is keyed on the
caller's spelling. With a relative spelling that key means a different database once the
working directory changes, and a row recorded for one store lands in another store's file.

On the ranking path it hides well. The obvious check, one query with two spellings, comes back
clean, because `search_symbols` returns on a result-cache hit
([`:759-823`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/search_symbols.py#L759-L823))
well before it reaches the ledger write, so only the first spelling records anything at all.

## Why this matters

PR #457 named this case, listing the constraints #442 left open:

> **Per-call `base_path`** — the cache is keyed by resolved path. A single connection would
> send one store's rows to another, which is the defect .188 fixed.

That is the failure this reports. The design is right; only the resolution step is missing. The
same intent is in the docstring
([`token_tracker.py:719`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L719))
and the v1.108.276 release note
([`CHANGELOG.md:223`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/CHANGELOG.md#L223)).

This was easy to miss. The same PR works through *"Two silent failure modes caching introduces,
which the report did not name"*: a stray `close()` poisoning the cache, and a deleted file
leaving writes in an unlinked inode. Both are real, both are guarded, and I had not thought of
either. This is a third of the same family, and the one that looked already covered.

`_perf_conns` is keyed on `str(path)`
([`:729`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L729),
[`:800`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L800)),
and `_perf_db_path` returns `root / _PERF_DB_FILE` unresolved at both exits
([`:602`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L602),
[`:611`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L611)).

With a relative spelling, the key names a different file after a working-directory change while
the cached connection still points at the original. `_perf_conn_usable` probes `exists()` at the
**new** location and asks the cached connection for `SELECT 1`
([`:642`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L642)),
and both can pass while the connection points somewhere else. In the A/B/A sequence below it is
store A's own database, created legitimately on the first write, that satisfies the `exists()`
probe when the process comes back to A, so the stale B connection is judged healthy and A's
second row is written into B.

v1.108.188's note is still in the tree: *"the store this call was told to use, not whichever one
the first savings write of the process happened to pin"*
([`search_symbols.py:1234-1236`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/search_symbols.py#L1234-L1236)).
The writers still name the right store. The cache redirects the row after they do.

That table is read by
[`ranking_db_query`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L1401),
which `WeightTuner` and `analyze_perf` consume.

`ranking_events` is not the only table affected, and the latency path reaches the same helper
more directly. `call_tool` takes the raw tool argument, `_store_arg = arguments.get("storage_path")`
([`server.py:7032-7034`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/server.py#L7032-L7034)),
and hands it straight to `record_tool_latency(base_path=...)`, which reaches
`_persist_perf_locked` and its `INSERT INTO tool_calls`
([`token_tracker.py:951-957`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L951-L957))
through the same cache, with no `IndexStore` normalising anything on the way.
`_persist_session_yield_locked`
([`:812`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L812))
uses the helper too; I did not trace whether a relative spelling reaches its `base_path`.

## Scope

Small:

- Installs without `perf_telemetry_enabled` are unaffected. It is off by default.
- The default store is `Path.home() / ".code-index"`, already absolute. This starts with a
  caller-supplied relative `storage_path`.
- **Cross-store misrouting also needs the working directory to change while one `_State`
  survives.** There is no `chdir` anywhere in `src/`. I checked unfiltered, and every match in
  the tree is in `tests/`. A long-running server with a fixed CWD does not hit it, and two
  servers launched from two project directories have separate caches. It matters for embedded or
  library use, or any host that moves the process.
- Nothing here touches indexed source content or the repo databases, and nothing is corrupted.
  A row arrives intact, in the wrong file. The effect is confined to local performance and
  analysis telemetry.

The duplicate-connection half needs none of that, only two spellings, and it is milder. Two
spellings take two entries, and `_perf_conns`
([`:161`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L161))
has no cap and no eviction, so each spelling holds a connection for the life of the process.

## Reproduction

`PYTHONPATH=src python repro_perf_alias.py`

```
cache keys held : ['store\\telemetry.db']
project_a rows  : 1  (2 written)
project_b rows  : 2  (1 written)
```

Three ranking writes, two from project A and one from project B, each naming its own store as
`"store"`. Two rows land in B and one in A.

```python
"""A relative storage_path sends one store's ranking rows into another store's database."""
import os, sqlite3, tempfile
from pathlib import Path

os.environ["JCODEMUNCH_PERF_TELEMETRY"] = "1"

from jcodemunch_mcp.storage.token_tracker import _State


def ranking_rows(db: Path) -> int:
    conn = sqlite3.connect(str(db))
    try:
        return conn.execute("SELECT COUNT(*) FROM ranking_events").fetchone()[0]
    finally:
        conn.close()


with tempfile.TemporaryDirectory(prefix="jcm-perf-alias-") as tmp:
    root = Path(tmp).resolve()
    project_a, project_b = root / "project_a", root / "project_b"
    (project_a / "store").mkdir(parents=True)
    (project_b / "store").mkdir(parents=True)

    state = _State()
    cwd = os.getcwd()
    try:
        # Three ranking writes, each from its own project, each naming its own
        # store the same relative way a project-local config would.
        for project, query in (
            (project_a, "q-a1"), (project_b, "q-b1"), (project_a, "q-a2"),
        ):
            os.chdir(project)
            state.record_ranking_event(
                tool="search_symbols", repo="local/demo", query=query,
                returned_ids=["demo.py::symbol#function"], base_path="store",
            )
    finally:
        os.chdir(cwd)

    print(f"cache keys held : {sorted(state._perf_conns)}")
    print(f"project_a rows  : {ranking_rows(project_a / 'store' / 'telemetry.db')}  (2 written)")
    print(f"project_b rows  : {ranking_rows(project_b / 'store' / 'telemetry.db')}  (1 written)")
    state.close_perf_dbs()
```

The script drives `_State` directly to keep it to one file, but the spelling reaches the sink
untouched from the tool surface: `IndexStore.__init__` stores `Path(base_path)` without
resolving
([`sqlite_store.py:788`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/sqlite_store.py#L788)),
`ledger_base_path` returns `str(store.base_path)`
([`_utils.py:23-35`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/_utils.py#L23-L35)),
and that is what the three `search_symbols` ledger exits pass
([`:1236`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/search_symbols.py#L1236),
[`:1650`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/search_symbols.py#L1650),
[`:2055`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/tools/search_symbols.py#L2055)).
Three `search_symbols` calls against one store spelled absolute, relative, and with a `..`
segment produced three cache entries for one file, each with a distinct query so the result
cache did not answer first.

## Expected vs actual

Actual: the key is the caller's spelling. Rows can land in another store's telemetry database,
and aliases each hold a connection.

Expected: one resolved path, one cache entry, and the row in the store the caller named.

On #442 you said not prescribing a fix was the right call, because the constraints were the
whole problem. That does not apply here. The constraint was already worked through and settled
in the PR, so the fix is the missing call at both exits.

```python
-                return root / _PERF_DB_FILE
+                return (root / _PERF_DB_FILE).resolve()
...
-            path = root / _PERF_DB_FILE
+            path = (root / _PERF_DB_FILE).resolve()
```

Three regression checks fail on `8afb119` and pass with it: aliases of one directory share a
single cached connection, the A/B/A sequence leaves two `ranking_events` rows in A and one in B,
and the same holds for `tool_calls` through `record_latency`. On that tree the full suite is
7,790 passed, 10 skipped, and `ruff check src/` is clean.

I have the change and both tests on a branch off `8afb119`, if a PR would help.

The reader helpers are unaffected and deliberately untouched: `ranking_db_query` and the
`analyze_perf` reader both go through the module-level `perf_db_path`
([`:1381`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L1381))
and open a fresh connection per call, so SQLite resolves at connect time and no key can go stale.

## When it started

Both `_perf_db_path` exits have lacked `.resolve()` since the helper arrived in
[v1.74.0](https://github.com/jgravelle/jcodemunch-mcp/commit/d99f90027dd28747d23f5b7bb68d290df8e5ab53),
where it was harmless. The cache is what made the key persist.

| Revision | project_a | project_b |
|---|---:|---:|
| [`1761710`](https://github.com/jgravelle/jcodemunch-mcp/commit/176171045bb5ae113684e79c981dce4813572b64), the commit before the cache | 2 | 1 |
| [`8afb119`](https://github.com/jgravelle/jcodemunch-mcp/commit/8afb119f884975af267937f1ad5384f9dc6d5c19), current `main` | 1 | 2 |
| `8afb119` with both exits resolved | 2 | 1 |

Before the cache, each event opened its own connection, so SQLite resolved the relative path
every time and the row landed correctly. The pre-cache arm was run with a variant reporting only
row counts, since `_perf_conns` does not exist there.

The cache arrived in
[`e9d5995`](https://github.com/jgravelle/jcodemunch-mcp/commit/e9d5995b3bb4d40def30dc25fecd9c0b3f7c288d)
via [PR #457](https://github.com/jgravelle/jcodemunch-mcp/pull/457), released as v1.108.276,
fixing [#441](https://github.com/jgravelle/jcodemunch-mcp/issues/441) and
[#442](https://github.com/jgravelle/jcodemunch-mcp/issues/442). Nothing has touched
`token_tracker.py` since.

## Environment

- jCodeMunch version: 1.108.277 (`main` at `8afb119`, release v1.108.277 at the same commit)
- Python version: 3.13.7
- OS: Windows 11

Every `file.py:line` reference is against
[`8afb119`](https://github.com/jgravelle/jcodemunch-mcp/tree/8afb119f884975af267937f1ad5384f9dc6d5c19).

## Anything else you found along the way

Both of these turned up in a final sweep, after this issue was already written. Had they
surfaced earlier they would have been separate reports rather than a note at the end. The
brevity here is about when I found them, not what they weigh.

**The same unresolved-identity assumption is in the core store caches, and there it is not
telemetry.** `SQLiteIndexStore._initialized_dbs`
([`sqlite_store.py:814-849`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/sqlite_store.py#L814-L849))
keys on `str(db_path)`, and that key gates schema creation and the whole v4→v17 migration chain.
Indexing one repo into two stores, both named `"store"` relatively, with the working directory
moved between them, leaves the second database with **zero tables** and fails with
`Indexing failed: no such table: symbols`. It fails loudly, which bounds it, but it leaves an
empty `.db` behind. `_VERIFIED_PATHS`
([`:791-794`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/sqlite_store.py#L791-L794))
has the same shape. The fix in this issue does not address either; they key off
`IndexStore.base_path`, not `_perf_db_path`.

**`_perf_rows_since_trim` is one counter for every telemetry database.** It is a single int on
`_State`
([`:959-970`](https://github.com/jgravelle/jcodemunch-mcp/blob/8afb119f884975af267937f1ad5384f9dc6d5c19/src/jcodemunch_mcp/storage/token_tracker.py#L959-L970))
while `_perf_conns` holds many, so the trim runs against whichever database happens to receive
the 1,000th process-wide write. Two stores written alternately, cap set to 1,000: the second
store absorbs every trim and the first reaches 2,000 rows and keeps growing. No relative path or
directory change needed, only the multiple storage paths the code already supports. Present
since v1.74.0 and unrelated to the cache.

## Duplicate check

No existing issue, pull request, discussion, or `ROADMAP.md` entry found. Searched open and
closed issues for `resolve`, `resolved path`, `cache key`, `relative path`, `storage_path`,
`telemetry.db`, `_perf_conns`, `symlink`, `connection cache`, `ranking_events`, `wrong
database`, `perf telemetry`, and `base_path`; pull requests for `perf db`; all five discussions;
and the full unfiltered open-issue list read whole rather than searched.

rknighton:
Closes #465.

`_perf_db_path` now resolves at both exits, so `_perf_conns` is keyed by the path the docstring
already claimed it was keyed by. Two source lines changed, three regression tests added.

```python
-                return root / _PERF_DB_FILE
+                return (root / _PERF_DB_FILE).resolve()
...
-            path = root / _PERF_DB_FILE
+            path = (root / _PERF_DB_FILE).resolve()
```

## Why at the helper rather than at the cache

`_perf_db_path` has exactly one caller, `_ensure_perf_db_locked`
([`:726`](https://github.com/jgravelle/jcodemunch-mcp/blob/362579d/src/jcodemunch_mcp/storage/token_tracker.py#L726)),
and all three telemetry sinks reach the cache through that one helper — session yield
([`:830`](https://github.com/jgravelle/jcodemunch-mcp/blob/362579d/src/jcodemunch_mcp/storage/token_tracker.py#L830)),
ranking
([`:900`](https://github.com/jgravelle/jcodemunch-mcp/blob/362579d/src/jcodemunch_mcp/storage/token_tracker.py#L900)),
and latency
([`:951`](https://github.com/jgravelle/jcodemunch-mcp/blob/362579d/src/jcodemunch_mcp/storage/token_tracker.py#L951)).
Resolving where the path is built fixes one place and every consumer inherits it, including
`_persist_session_yield_locked`, whose reachability the issue left open.

## The tests

One per consequence, in the order the issue names them.

- **`test_aliases_of_one_store_share_one_cached_connection`** — one directory spelled two ways
  yields one cache entry and two rows in one file. `_perf_conns` has no cap and no eviction, so
  before the fix each spelling held a connection for the life of the process.
- **`test_relative_spelling_does_not_write_into_another_store`** — the A/B/A sequence. Two
  `ranking_events` rows in A, one in B. Before the fix, one and two.
- **`test_relative_spelling_does_not_misroute_latency_rows`** — the same through `record_latency`,
  which `call_tool` reaches with the raw `storage_path` argument and no `IndexStore` normalising it
  on the way. `analyze_perf` reads both tables through one base path, so a split between them is
  worse than either alone.

The module docstring carries that mapping.

## Verification

- Focused suite: **3 passed**.
- **Reverting either `.resolve()` on its own turns the suite red.** The two exits are asserted
  separately: every row-level test supplies an explicit `base_path`, so the no-argument exit needs
  its own assertion rather than inheriting coverage from them.
- Full suite green, `ruff check src/` clean.

⚠ **Reported as a delta, because absolute suite totals are not comparable across machines here.**
Ten test modules call `pytest.importorskip` at module level, gated on `watchfiles`, `pyyaml`,
`starlette` and `tiktoken`. When one of those is missing, pytest reports a **single skip and
collects nothing from that module** — so 209 tests can leave the total with nothing failing and no
per-test skip to show for it. `uv sync` and the `pip install -e .` line in `CONTRIBUTING.md` do not
install the same set, so two correct runs of the same commit can report totals a hundred apart.

Base and candidate ran back to back from one checkout: **+3 passed, 0 failed on both arms, no skip
or warning change.**

## Existing installs

Unaffected. The default store is `Path.home() / ".code-index"`, and `.resolve()` on an
already-absolute path with no symlinks or `..` returns it unchanged. Nothing backfills or migrates
rows that already landed in the wrong file.

### DIFF
--- a/src/jcodemunch_mcp/storage/token_tracker.py
+++ b/src/jcodemunch_mcp/storage/token_tracker.py
@@ -599,7 +599,7 @@ class _State:
             try:
                 root = Path(base_path)
                 root.mkdir(parents=True, exist_ok=True)
-                return root / _PERF_DB_FILE
+                return (root / _PERF_DB_FILE).resolve()
             except Exception:
                 logger.debug("Failed to resolve perf db path at %s", base_path, exc_info=True)
                 return None
@@ -608,7 +608,7 @@ class _State:
         try:
             root = Path(self._base_path) if self._base_path else Path.home() / ".code-index"
             root.mkdir(parents=True, exist_ok=True)
-            path = root / _PERF_DB_FILE
+            path = (root / _PERF_DB_FILE).resolve()
             self._perf_db_path_cached = path
             return path
         except Exception:
--- a/tests/test_perf_db_path_resolution.py
+++ b/tests/test_perf_db_path_resolution.py
@@ -1,0 +1,133 @@
+"""The perf-db connection cache must key on the resolved path.
+
+``_ensure_perf_db_locked`` documents its cache as "Keyed by resolved path", but
+``_perf_db_path`` returned ``root / _PERF_DB_FILE`` unresolved, so the key was
+whatever spelling the caller happened to use. Three consequences, one per test:
+
+* Aliases of one directory each got their own process-lifetime connection, and
+  ``_perf_conns`` has no cap and no eviction.
+* A RELATIVE spelling made the key depend on the process CWD. After a chdir the
+  same key names a different database, and ``_perf_conn_usable``'s ``exists()``
+  probe consults the new location while the cached connection still points at
+  the old file -- so a row recorded for one store is written into another's.
+  That is the failure v1.108.188 fixed for the writers; an unresolved cache key
+  reintroduces it one layer down.
+* The latency sink takes it too, by a shorter route: ``call_tool`` hands the raw
+  ``storage_path`` argument to ``record_latency`` with no ``IndexStore`` in
+  between. ``analyze_perf`` reads ``tool_calls`` and ``ranking_events`` through
+  one base path, so a split between them is worse than either alone.
+"""
+
+import os
+import sqlite3
+from pathlib import Path
+
+import pytest
+
+from jcodemunch_mcp.storage import token_tracker
+
+
+@pytest.fixture
+def perf_enabled(monkeypatch):
+    real_get = token_tracker._config.get
+
+    def fake_get(key, default=None, **kwargs):
+        if key == "perf_telemetry_enabled":
+            return True
+        return real_get(key, default, **kwargs)
+
+    monkeypatch.setattr(token_tracker._config, "get", fake_get)
+
+
+def _rows(db: Path, table: str) -> int:
+    conn = sqlite3.connect(str(db))
+    try:
+        return conn.execute(f"SELECT COUNT(*) FROM {table}").fetchone()[0]
+    finally:
+        conn.close()
+
+
+def _ranking_rows(db: Path) -> int:
+    return _rows(db, "ranking_events")
+
+
+def test_aliases_of_one_store_share_one_cached_connection(perf_enabled, tmp_path):
+    store = tmp_path / "store"
+    (store / "nested").mkdir(parents=True)
+
+    state = token_tracker._State()
+    # The no-argument exit is the other half of the fix and the writes below never
+    # reach it: they all pass an explicit base_path. `self._base_path` keeps whatever
+    # spelling the first `_ensure_loaded` was handed, so it needs its own assertion or
+    # reverting that one `.resolve()` leaves this file green.
+    state._base_path = str(store / "nested" / "..")
+    assert state._perf_db_path() == (store / "telemetry.db").resolve()
+
+    try:
+        for i, spelling in enumerate((str(store), str(store / "nested" / ".."))):
+            state.record_ranking_event(
+                tool="t", repo="local/demo", query=f"q-{i}",
+                returned_ids=["a"], base_path=spelling,
+            )
+        assert len(state._perf_conns) == 1, (
+            f"one directory, {len(state._perf_conns)} cached connections: "
+            f"{sorted(state._perf_conns)}"
+        )
+        assert _ranking_rows(store / "telemetry.db") == 2
+    finally:
+        state.close_perf_dbs()
+
+
+def test_relative_spelling_does_not_write_into_another_store(perf_enabled, tmp_path):
+    project_a, project_b = tmp_path / "project_a", tmp_path / "project_b"
+    (project_a / "store").mkdir(parents=True)
+    (project_b / "store").mkdir(parents=True)
+
+    state = token_tracker._State()
+    cwd = os.getcwd()
+    try:
+        for project, query in (
+            (project_a, "q-a1"), (project_b, "q-b1"), (project_a, "q-a2"),
+        ):
+            os.chdir(project)
+            state.record_ranking_event(
+                tool="t", repo="local/demo", query=query,
+                returned_ids=["a"], base_path="store",
+            )
+    finally:
+        os.chdir(cwd)
+        state.close_perf_dbs()
+
+    assert _ranking_rows(project_a / "store" / "telemetry.db") == 2
+    assert _ranking_rows(project_b / "store" / "telemetry.db") == 1
+
+
+def test_relative_spelling_does_not_misroute_latency_rows(perf_enabled, tmp_path):
+    """Same defect on the latency sink, which reaches the cache more directly.
+
+    ``call_tool`` passes the raw ``storage_path`` argument straight through to
+    ``record_latency``, with no ``IndexStore`` normalising it on the way, so
+    ``tool_calls`` takes the same misrouting as ``ranking_events``. Its own
+    docstring is the reason this matters: ``analyze_perf`` reads BOTH tables
+    through one base path.
+    """
+    project_a, project_b = tmp_path / "project_a", tmp_path / "project_b"
+    (project_a / "store").mkdir(parents=True)
+    (project_b / "store").mkdir(parents=True)
+
+    state = token_tracker._State()
+    cwd = os.getcwd()
+    try:
+        for project, tool in (
+            (project_a, "search_symbols"), (project_b, "find_references"),
+            (project_a, "get_file_outline"),
+        ):
+            os.chdir(project)
+            state.record_latency(tool, 1.0, ok=True, repo="local/demo",
+                                 base_path="store")
+    finally:
+        os.chdir(cwd)
+        state.close_perf_dbs()
+
+    assert _rows(project_a / "store" / "telemetry.db", "tool_calls") == 2
+    assert _rows(project_b / "store" / "telemetry.db", "tool_calls") == 1


jgravelle (approved):
Merging. The fix is at the right layer: _perf_db_path is the single place the key is built, and both exits needed it, which your first test asserts separately rather than letting the row-level tests imply.

Two things I checked rather than took on trust. _perf_conns has no cap and no eviction, so the alias case really was a per-spelling connection for the life of the process, not just a duplicate lookup. And the module-level perf_db_path() helper at :1382 still builds the path unresolved -- it stays that way on purpose: it never touches the cache, and an unresolved spelling opens the same file on disk.

The delta-not-totals note is the right call, and the reason you give is one we hit here too.

PR #474, opened 2026-08-15
feat(telemetry): correlate perf rows by session and call (#456)

### LINKED ISSUE #456: Add nullable `session_uid` and `call_uid` columns to the perf telemetry tables
## The ask

`ranking_events` and `tool_calls` share no identifier, so a ranking event cannot be
attributed to the call that produced it. Please add nullable `session_uid` and
`call_uid` `TEXT` columns to both tables.

Scope: no backfill, no new tool, no index, no foreign key, no retention change, no
reader change, no new outbound data, no new dependency. Existing rows keep `NULL`.

## What is missing

`ranking_events` records ranking detail and `tool_calls` records duration, but nothing
relates the two. `ts`, `repo`, and `tool` are not unique. This is visible in the table
definitions and needs no dataset to establish.

`session_uid` already exists. It is the primary key of `session_yield` and is already
generated per process, so this request reuses an identifier rather than inventing one.
Only the per-call value is new.

## What it enables

Any query whose unit is a call or a process: how long the call that produced a given
ranking event took, which events belong to the same run, how a session's ranking
behaviour relates to the yield already recorded in `session_yield`.

## Requested shape

- Add both columns to `tool_calls` and `ranking_events`, nullable, no default.
- `session_uid`: both sinks are `_State` methods, so they read `self._session_uid`
  directly. Nothing new is needed.
- `call_uid`: set a fresh value on each entry into the registered `call_tool` and read
  it inside the sinks. A module-level `ContextVar` works; `contextvars` is already
  imported and already used for request-scoped state in `server.py`. Both dispatch
  styles see it: offloaded tool work goes through `asyncio.to_thread`, which copies the
  context, and the front-door tools that run inline on the event loop share it directly.
- Reset with the token returned by `set()`, not by setting `None`. See re-entrancy below.
- Add the columns using the `PRAGMA table_info` plus conditional `ALTER TABLE` pattern
  already in `sqlite_store.py`.
- Writes outside a dispatched call record `NULL`, which is why the columns are nullable.

No tool signature and no call site changes. The six `record_ranking_event` call sites
and the single latency call site stay as they are, because both values are read at the
sink rather than passed in.

## Three things worth knowing before implementing

**The schema probe must not run per write.** `_ensure_perf_db_locked` is called on every
telemetry write, so an unguarded probe would add cost to a path that already has an open
issue against it (linked below). Guarding it with a process-local set of migrated paths
keeps it to once per path per process. The set must be keyed by resolved path, not a
single flag, because `base_path` allows several telemetry databases in one process.

**The dispatcher is re-entrant.** `order` and executable `route` both re-enter the
registered `call_tool`, so one client request can produce two `tool_calls` rows. Giving
each entry its own `call_uid` keeps a ranking event joined to exactly one latency row,
which is why token-based reset matters: setting `None` in the `finally` would clear the
outer entry's value and silently write `NULL` for it. The consequence to expect is that
`COUNT(DISTINCT call_uid)` counts dispatcher entries rather than client requests, and
the front-door row has no matching ranking event.

**The join is best-effort, by design.** `tool_calls` is trimmed to a rolling cap while
`ranking_events` is not, so older events will outlive their `tool_calls` row. Queries
should use `LEFT JOIN` and expect misses. These are correlation keys, not referential
integrity. Aligning retention is a separate concern and not part of this request.

## Acceptance criteria

- [ ] A fresh database has both columns on both tables.
- [ ] A legacy database gains them without altering existing rows, which keep `NULL`.
- [ ] The probe runs once per resolved path per process; a second path migrates
      independently.
- [ ] A ranking row and a latency row from the same dispatcher entry carry the same
      non-null values.
- [ ] Concurrent calls receive distinct `call_uid` values.
- [ ] On re-entrant dispatch through `order` or executable `route`, the inner entry gets
      its own `call_uid`, the outer entry still has its original one after the inner
      completes, and the context returns to what it was before the outermost entry once
      that entry exits. Holds when a call fails at either level.
- [ ] A write outside any dispatched call stores `NULL`.
- [ ] Behaviour with telemetry disabled, and the outbound savings payload, are unchanged.

## Source

All references are against
[`186c519`](https://github.com/jgravelle/jcodemunch-mcp/commit/186c5191c75b06cf3ac37561422ca672aa5c19ab),
which declares version `1.108.275`.

| What | Where |
| --- | --- |
| The three table definitions | [`token_tracker.py:611-665`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L611-L665) |
| Per-write call sites into that function, in the ranking sink and the latency sink | [`token_tracker.py:767`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L767) and [`:811`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L811) |
| Existing per-process `session_uid` | [`token_tracker.py:140-143`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L140-L143) |
| Existing `ContextVar` for request state | [`server.py:496-504`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/server.py#L496-L504) |
| Existing additive migration pattern | [`sqlite_store.py:465-479`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/sqlite_store.py#L465-L479) |
| Registered dispatcher | [`server.py:5279-5286`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/server.py#L5279-L5286) |
| `order` re-entry | [`server.py:5099-5110`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/server.py#L5099-L5110) |
| `route` re-entry | [`server.py:5134-5176`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/server.py#L5134-L5176) |
| `tool_calls` trim | [`token_tracker.py:815-830`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L815-L830) |
| Outbound payload, for the no-new-data claim | [`token_tracker.py:1004-1014`](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/storage/token_tracker.py#L1004-L1014) |

Finer HTTP-client attribution is out of scope; [`_session_key()` at server.py:540](https://github.com/jgravelle/jcodemunch-mcp/blob/186c5191c75b06cf3ac37561422ca672aa5c19ab/src/jcodemunch_mcp/server.py#L540) already exists for that if it is ever wanted.

This touches the same schema function as
[`returned_ids` is capped at 50 with no record of the true result count](https://github.com/jgravelle/jcodemunch-mcp/issues/441)
and
[Opt-in ranking telemetry: ~80% of each write is unnecessary repeated database setup](https://github.com/jgravelle/jcodemunch-mcp/issues/442),
but does not depend on either.

## Duplicate check

No existing request found across open and closed issues using `join`, `correlate`,
`correlation`, `call_uid`, `session_uid`, `request_id`, `trace`, `span`, `foreign key`,
and `tool_calls`.

rknighton:
Closes #456.

`ranking_events` and `tool_calls` now carry nullable `session_uid` and `call_uid`, so a ranking
event can be attributed to the call that produced it. Three files, 442 insertions, 5 deletions, 62
of them source.

```python
+            for _table in ("tool_calls", "ranking_events"):
+                self._add_column_if_missing(conn, _table, "session_uid", "TEXT")
+                self._add_column_if_missing(conn, _table, "call_uid", "TEXT")
```

```python
 async def call_tool(name: str, arguments: dict) -> list[TextContent] | CallToolResult:
+    call_token = begin_call_context()
+    try:
         return _enforce_response_cap(name, await _call_tool_impl(name, arguments))
+    finally:
+        end_call_context(call_token)
```

Both sinks read `self._session_uid` and `_CURRENT_CALL_UID.get()` at write time, so the six
`record_ranking_event` call sites and the single latency call site are unchanged.

## Why a ContextVar set at the dispatcher

The identity has to reach two sinks that no one call site owns: ranking
([`:913`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883a/src/jcodemunch_mcp/storage/token_tracker.py#L913))
and latency
([`:956`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883a/src/jcodemunch_mcp/storage/token_tracker.py#L956)),
written from different paths, with `record_latency` also reached directly by the front door.
Setting it once at the registered `call_tool`
([`:5280`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883a/src/jcodemunch_mcp/server.py#L5280))
gives every route one identity with no call-site and no signature change.

The dispatcher is re-entrant. `order` and executable `route` both re-enter `call_tool`, so one
client request can produce two `tool_calls` rows, and each entry needs its own `call_uid` to keep a
ranking event joined to exactly one latency row. Reset is by token rather than by setting `None`,
which would clear the outer entry's value and write `NULL` for it. That analysis sits at the set
site in `server.py`.

The columns are added the additive way `returned_count` was
([`:798`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883a/src/jcodemunch_mcp/storage/token_tracker.py#L798)),
not declared in `CREATE TABLE`, so the schema keeps one source of truth.

## The tests

The focused file collects 14 cases across nine named tests, in the order the issue names the
criteria. Existing upstream tests keep the telemetry-disabled coverage; this file adds the
issue-specific behaviour and the payload check.

- **Schema.** Both columns on both tables on a fresh database; legacy rows still `NULL` after a
  four-column migration; the probe runs once per resolved path per process.
- **Identity.** One dispatcher entry joins its ranking and latency rows; concurrent entries get
  distinct ids; re-entrant dispatch restores the outer context on both success and failure.
- **Boundaries.** A write outside any dispatched call stores `NULL` rather than inventing an
  identity, and the outbound payload keeps exactly its legacy keys.
- **The shipped path.** A real `server.call_tool("search_symbols", ...)` runs the whole chain: the
  dispatcher binds the identity, `_call_tool_impl` runs the tool through `asyncio.to_thread`, the
  tool writes its ranking row through the module singleton, and the latency row comes from
  `_call_tool_impl`'s own `finally`. Both rows carry the same non-null session and call id.

## Verification

- Focused suite: **14 passed**.
- **Blanking the read turns the suite red**, with the two sites gated separately rather than
  mutated together.
- Full suite under CI's own command on Windows 11 / Python 3.13.7: **7829 passed / 16 skipped**,
  against **7815 / 16** on the base: **+14**, no skip change. Both arms back to back from one
  checkout, coverage gate met at 79.67%.

  ```bash
  uv sync --locked --group dev --extra watch
  uv run pytest tests/ -v --tb=short --cov=src --cov-report=term-missing --cov-fail-under=74
  ```

  That is one leg. Your matrix covers Ubuntu and Windows across 3.10 to 3.13 and will report on
  the PR itself.
- `ruff check src/` clean, `compileall` clean.
- Rebased onto `2e3883a`. Every hunk header is unchanged: #473 substitutes two lines for two in
  `_perf_db_path` and shifts nothing under this branch.

## Existing installs

Existing databases migrate additively. `perf_telemetry_enabled` is `False` by default, so most
installs open no perf database at all; where it is on, the first write after upgrade adds the four
columns through the same `_add_column_if_missing` path `returned_count` uses. Both are nullable,
existing rows keep `NULL`, and nothing backfills them. No index, foreign key, retention or reader
change.

Two properties worth knowing before querying: `tool_calls` is trimmed to a rolling cap while
`ranking_events` is not, so the join is best-effort and wants `LEFT JOIN`; and `COUNT(DISTINCT
call_uid)` counts dispatcher entries rather than client requests, with the front-door row having no
matching ranking event. The `LEFT JOIN` contract sits at the migration site.

### DIFF
--- a/src/jcodemunch_mcp/server.py
+++ b/src/jcodemunch_mcp/server.py
@@ -5283,7 +5283,20 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent] | CallToolR
     Kept as the registered entry point (and the name the front door re-dispatches
     through) so every route into the dispatcher passes the cap.
     """
-    return _enforce_response_cap(name, await _call_tool_impl(name, arguments))
+    from .storage.token_tracker import begin_call_context, end_call_context
+
+    # **The dispatcher is re-entrant.** `order` and executable `route` both re-enter the
+    # registered `call_tool`, so one client request can produce two `tool_calls` rows. Giving
+    # each entry its own `call_uid` keeps a ranking event joined to exactly one latency row,
+    # which is why token-based reset matters: setting `None` in the `finally` would clear the
+    # outer entry's value and silently write `NULL` for it. The consequence to expect is that
+    # `COUNT(DISTINCT call_uid)` counts dispatcher entries rather than client requests, and
+    # the front-door row has no matching ranking event.
+    call_token = begin_call_context()
+    try:
+        return _enforce_response_cap(name, await _call_tool_impl(name, arguments))
+    finally:
+        end_call_context(call_token)
 
 
 async def _call_tool_impl(name: str, arguments: dict) -> list[TextContent] | CallToolResult:
--- a/src/jcodemunch_mcp/storage/token_tracker.py
+++ b/src/jcodemunch_mcp/storage/token_tracker.py
@@ -1385,8 +1385,8 @@ def record_tool_latency(
     _state.record_latency(tool_name, duration_ms, ok=ok, repo=repo, base_path=base_path)
 
 
-def begin_call_context(call_uid: Optional[str] = None) -> Token:
-    """Bind one dispatcher call identifier to the current execution context (#456).
+def begin_call_context() -> Token:
+    """Bind a fresh dispatcher call identifier to the current execution context (#456).
 
     ⚠ Reset with the returned token in a ``finally``. A skipped reset leaves the
     inner entry's identity visible to the outer one after it returns, so the join
@@ -1398,7 +1398,7 @@ def begin_call_context(call_uid: Optional[str] = None) -> Token:
     ``call_tool`` stays the single registered entry: the invariant is held by there
     being one door, not by this helper.
     """
-    return _CURRENT_CALL_UID.set(call_uid or uuid.uuid4().hex)
+    return _CURRENT_CALL_UID.set(uuid.uuid4().hex)
 
 
 def end_call_context(token: Token) -> None:
--- a/src/jcodemunch_mcp/storage/token_tracker.py
+++ b/src/jcodemunch_mcp/storage/token_tracker.py
@@ -36,6 +36,7 @@ import threading
 import time
 import uuid
 from collections import OrderedDict, deque
+from contextvars import ContextVar, Token
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional
@@ -43,6 +44,14 @@ from typing import Optional
 from .. import config as _config
 
 logger = logging.getLogger(__name__)
+
+# One identifier per entry into the registered call_tool. Set by the dispatcher,
+# read by the telemetry sinks at write time. Default None so writes outside any
+# dispatched call record NULL rather than inventing an identity.
+_CURRENT_CALL_UID: ContextVar[Optional[str]] = ContextVar(
+    "jcodemunch_current_call_uid",
+    default=None,
+)
 
 _SAVINGS_FILE = "_savings.json"
 
@@ -1360,6 +1369,16 @@ def record_tool_latency(
 ) -> None:
     """Record a tool-call duration for the current session (and optional perf db)."""
     _state.record_latency(tool_name, duration_ms, ok=ok, repo=repo, base_path=base_path)
+
+
+def begin_call_context(call_uid: Optional[str] = None) -> Token:
+    """Bind one dispatcher call identifier to the current execution context."""
+    return _CURRENT_CALL_UID.set(call_uid or uuid.uuid4().hex)
+
+
+def end_call_context(token: Token) -> None:
+    """Restore the execution context that preceded begin_call_context."""
+    _CURRENT_CALL_UID.reset(token)
 
 
 def latency_stats() -> dict:
--- a/src/jcodemunch_mcp/storage/token_tracker.py
+++ b/src/jcodemunch_mcp/storage/token_tracker.py
@@ -757,9 +757,7 @@ class _State:
                     tool      TEXT NOT NULL,
                     duration_ms REAL NOT NULL,
                     ok        INTEGER NOT NULL,
-                    repo      TEXT,
-                    session_uid TEXT,
-                    call_uid  TEXT
+                    repo      TEXT
                 )
             """)
             conn.execute("CREATE INDEX IF NOT EXISTS ix_tool_calls_tool ON tool_calls(tool)")
@@ -791,9 +789,7 @@ class _State:
                     confidence     REAL,
                     semantic_used  INTEGER NOT NULL,
                     identity_hit   INTEGER NOT NULL,
-                    repo_is_stale  INTEGER NOT NULL,
-                    session_uid    TEXT,
-                    call_uid       TEXT
+                    repo_is_stale  INTEGER NOT NULL
                 )
             """)
             conn.execute("CREATE INDEX IF NOT EXISTS ix_ranking_events_repo ON ranking_events(repo)")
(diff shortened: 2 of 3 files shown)


jgravelle (approved):
Merging. The ContextVar at the registered dispatcher is the right seam, and the re-entrancy analysis is the part that would have been easy to get wrong: order and executable route both re-enter, so resetting by token rather than by setting None is load-bearing, not style.

Two things I checked rather than took on trust, both because this is egress-adjacent. The outbound payload test asserts exact equality AND a negative on the two new keys, so a future field cannot drift into the anonymous record unnoticed -- that is a stronger guard than the criterion asked for, and it is the one I would have added. And the columns go through _add_column_if_missing the way returned_count did, so CREATE TABLE stays the single source of truth.

The end-to-end test earning its place is the detail worth naming: every other test in the file drives a fresh _State or stubs the impl, so without it the columns could stay, those tests stay green, and call_uid quietly goes NULL. Setting CODE_INDEX_PATH and storage_path both to the temp store is right too -- leaving the argument off sends the latency row to the developer's real store, which is the same class of leak as #411.

Second issue you have closed in two days, both inside the window.

PR #477, opened 2026-08-15
bench: measure the Counter front door, and stop quoting its saving

jgravelle:
## What this fixes

`benchmarks/route_recall/run_route_recall.py` claimed the Counter avoided
**~98%** of resident schema tokens. That figure had no run behind it.

Measured, it is **95.9%**.

## Why it survived

`tests/test_schema_budget.py` iterates `tool_profile`. `tool_profile` does not
apply to the Counter front door at all -- `tool_surface: "counter"` replaces the
catalog outright with six tools. So the project's largest context lever sat
outside its own guardrail, and the published number sat in a docstring where
nothing could fail when it drifted.

The direction is what makes it worth a PR rather than a typo fix. Overstating
the saving makes the recall trade `run_route_recall.py` exists to price look
cheaper than it is.

## What the numbers actually are

Measured with tiktoken `cl100k_base`, all arms, in `benchmarks/schema_baseline.json`:

| arm | tokens | vs `full_full` |
|---|---:|---:|
| `full_full` | 22,741 | baseline |
| `standard_full` | 21,431 | 5.8% |
| `core_compact` | 3,885 | 74.0% |
| `counter_full` | 939 | 95.9% |
| `counter_compact` | 939 | 95.9% |

Two findings beyond the fix:

- **`tool_profile: "standard"` is not a token lever.** 5.8%. Only `core` and
  `counter` move the number, and there is no gradient between them.
- **`compact_schemas` is a no-op under `counter`** -- measured, not assumed, so
  the identical pair above is a recorded fact rather than a reader's guess.
- **Descriptions are 36.2% of the `full` payload** and `compact_schemas` never
  touches them. `--breakdown` prints the names/descriptions/schemas split and
  the ten costliest tools. Any further cut has to come from there.

## Changes

- `benchmarks/harness/capture_schema_baseline.py` -- extended, not duplicated.
  A second script would have produced a third set of absolutes for the same
  question. Adds the two counter arms and a `--breakdown` mode.
- `benchmarks/schema_baseline.json` -- regenerated, two new keys.
- `benchmarks/route_recall/run_route_recall.py` -- `_schema_saving()` reads the
  baseline at runtime and prints it at the top of every run. A missing baseline
  prints "not measured"; it never falls back to a remembered figure.
- `tests/test_schema_budget.py` -- `test_counter_surface_within_baseline_tolerance`
  puts the counter arms under the same 5% ceiling; `test_counter_saving_is_read_not_typed`
  fails if any schema-saving percentage returns to the harness as a literal.
- `CLAUDE.md` -- names `schema_baseline.json` as canonical and marks
  `benchmarks/codex_surface/`'s 24,007/1,030 as a 2026-07 snapshot from a
  different payload shape, so the next reader does not read the gap as a
  regression.

## Verification

Both new tests confirmed **non-vacuous** against pre-fix `main`: 2 failed / 4
passed before, 6 passed after. The regex half was checked separately, because
the cheaper `_schema_saving in text` assertion fails first and would have
masked it.

`uv run ruff check src/` clean; the three touched files clean. 20 pre-existing
ruff errors elsewhere under `benchmarks/` are untouched.

## Notes for review

- **No `CHANGELOG.md` entry.** There is no `[Unreleased]` section, and #443 is
  an open fork PR -- our changelog edits are what conflict it.
- **Methodology caveat, stated in the harness docstring.** `cl100k_base` is a
  proxy for whatever tokenizer a client actually uses, so the absolute counts
  are approximate. The ratios are not: `o200k_base` and raw UTF-8 bytes
  reproduce every percentage above to within 0.3 points. Publish the ratios.
- **Scope.** These are the counts for this server's own tool list. A session
  behind the jmunch proxy carries the proxy's facade tools too, so real resident
  cost there is higher than `full_full`.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -286,6 +286,27 @@ the r/codex result than the raw number suggests, not a stronger one.**
 default `full`, 6 / 1,030 at `counter`) and needs no API credits; what it does
 not measure is what that costs in practice.
 
+⚠ **Those two numbers are a 2026-07 snapshot from THIS harness and are not the
+canonical figures.** `benchmarks/schema_baseline.json` is, written by
+`benchmarks/harness/capture_schema_baseline.py` and guarded by
+`tests/test_schema_budget.py`; it counts a different payload shape, so the two
+sets will never agree digit for digit and neither is wrong. Quote the baseline
+file. ⚠⚠ **Reconciled 2026-08-14: the Counter avoids 95.9%, not the ~98% that
+`run_route_recall.py` asserted for two months** — that literal is now computed
+from the baseline at runtime, with a test that fails if any schema-saving
+percentage returns to that file. **The gap existed because the budget guardrail
+only walked `tool_profile`, which does not apply to the front door at all**, so
+the single largest lever in the project had no test under it.
+
+⚠⚠ **The same run killed `tool_profile: "standard"` as a token lever: it drops 9
+of 91 tools and 5.7% of the payload.** Anyone selecting it as the safe middle
+setting gets nothing measurable. `core` (74.0%) and `counter` (95.9%) are the
+only two settings that move the number; there is no gradient between them, and
+the config surface currently implies there is. ⚠ Where the rest sits, from
+`--breakdown`: under `full`, tool DESCRIPTIONS are 36% of the payload and
+`compact_schemas` rewrites input schemas only, never descriptions. Schema
+compaction is near its floor; descriptions are untouched ground.
+
 ⚠ Design flaw recorded so nobody repeats it: summing per-invocation input across
 a RESUMED conversation counts accumulated context on every step, so the total is
 dominated by how much the agent read early on, which compounds.
--- a/benchmarks/harness/capture_schema_baseline.py
+++ b/benchmarks/harness/capture_schema_baseline.py
@@ -1,20 +1,34 @@
-"""Capture tools/list schema token counts at each profile x compact combo.
+"""Capture tools/list schema token counts at each surface x profile x compact combo.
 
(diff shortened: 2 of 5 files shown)


jgravelle:
Approving and merging. 11/11 green, and a trial merge onto current main (which has moved twice since this branched: #474 and the v1.108.280 release) applies cleanly and keeps both guard files green.

I ran the non-vacuity pass rather than trusting green, because this PR's whole thesis is that a guardrail existed and could not fire.

Reinstating the exact defect -- `~98% of resident schema tokens` back in the harness docstring -- turns `test_counter_saving_is_read_not_typed` red with the offending string quoted back. Dropping `counter_full` in the baseline to a value the front door exceeds turns `test_counter_surface_within_baseline_tolerance` red. Both fire.

The design choice worth naming is extending `capture_schema_baseline.py` instead of writing a second script. A second one would have produced a third set of absolutes for the same question, and this project already has the four-artifact `jcm_reference.json` drift on record as what that costs.

Two notes, neither blocking.

The transcription regex matches a percentage followed within 60 characters by schema/resident/token wording. `98%` written AFTER the noun -- "avoids most of the resident schema tokens -- 98%" -- would slip past. It catches the shape that actually occurred, which is the right bar for a ratchet; noting it so a future reader knows the boundary rather than assuming the guard is total.

`standard` measuring 5.8% is the finding with the longest tail. The config surface still reads as a gradient from core to standard to full, and it is not one: only `core` and `counter` move the number. That is a documentation change beyond this PR's scope, and it is the kind of thing that stays true and unsaid for months.

PR #479, opened 2026-08-15
fix(storage): resolve relative cache paths

### LINKED ISSUE #475: A second store at the same relative path is left uninitialised

Two stores named by the same relative `storage_path` from two working directories in one process
share a cache key, so the second one is opened as though it were already prepared: its database
stays empty and `index_folder` fails with `no such table: symbols`.

**Severity: low, but not opt-in.** It needs a relative `storage_path` and a working-directory change
in one process, which takes embedded or library use. Unlike #465 there is no `perf_telemetry_enabled`
gate, so this is the ordinary index path. Nothing indexed is corrupted, and both caches are
per-process, so a restart clears the cause and the next access repairs the file.

The two arms fail differently, and only one of them is loud. A **new** database at the colliding key
skips schema creation and fails loudly on first use with `no such table: symbols`. An **existing**
database at that key instead returns from the cache gate with the v4 to v17 chain skipped and **no
diagnostic at all** — the guard wraps schema creation and every migration together, so a cache hit
returns the connection before either can run. I am not claiming a silent wrong query result: several
absent-table consumers are defensive, and I did not turn the skipped migration into a verified
incorrect answer.

## Mechanism

Two process-wide caches key on the caller's spelling rather than on what it resolves to:

- `_VERIFIED_PATHS` holds `str(self.base_path)` and gates the `mkdir`
  ([`sqlite_store.py:791-794`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/sqlite_store.py#L791-L794),
  same three lines in
  [`index_store.py:393-396`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/index_store.py#L393-L396)).
- `_initialized_dbs` holds `db_key = str(db_path)` and gates schema creation **and the whole v4 to
  v17 migration chain**
  ([`:814-849`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/sqlite_store.py#L814-L849)).

`self.base_path` is `Path(base_path)` unresolved
([`:788`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/sqlite_store.py#L788))
and `_db_path` builds on it
([`:802`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/sqlite_store.py#L802)),
so one key names two directories once the CWD moves.

## Proof

At `2e3883a`. One source folder indexed twice under `storage_path="store"` from two working
directories, both store directories pre-created so this isolates `_initialized_dbs`:

```text
A: success=True symbol_count=1 error=None
A: local-sample-d72213bd.db -> 15 tables, 176128 bytes
B: success=False symbol_count=None error=Indexing failed: no such table: symbols
B: local-sample-d72213bd.db -> 0 tables, 0 bytes
_VERIFIED_PATHS  : ['store']
_initialized_dbs : ['store\\local-sample-d72213bd.db']
```

The migration gate fails the same way. Both stores start as current-schema fixtures marked
`index_version = 4` with the v17 `scip_*` tables dropped:

```text
after   A : index_version=17  scip_edges=True
after   B : index_version= 4  scip_edges=False
```

B stayed at 4 and nothing records that it was skipped; the load path only rejects a stored version
*above* the current one
([`:1556`](https://github.com/jgravelle/jcodemunch-mcp/blob/2e3883ab50884c550e8533a439b57d81f52f6060/src/jcodemunch_mcp/storage/sqlite_store.py#L1556)).
The fixture exercises the gate but is not a database v4 wrote.

With the pre-creation removed, the `_VERIFIED_PATHS` half alone gives
`OperationalError: unable to open database file` on B.

`PYTHONPATH=src python <script>`, standard library plus the package. **Click each triangle to expand
a script.**

<details>
<summary>1. Empty database, failed index (click to expand)</summary>

```python
"""One source repo, two stores, same relative spelling, CWD moved between them."""
import os, sqlite3, tempfile
from pathlib import Path

from jcodemunch_mcp.storage.sqlite_store import SQLiteIndexStore, _VERIFIED_PATHS

with tempfile.TemporaryDirectory(prefix="jcm-initdbs-") as tmp:
    root = Path(tmp).resolve()
    project = root / "sample"          # ONE source path, so the repo slug is stable
    project.mkdir()
    (project / "billing.py").write_text(
        "def calculate_invoice_total(items):\n    return sum(items)\n", encoding="utf-8"
    )
    work_a, work_b = root / "work_a", root / "work_b"
    (work_a / "store").mkdir(parents=True)
    (work_b / "store").mkdir(parents=True)

    cwd = os.getcwd()
    try:
        for label, workdir in (("A", work_a), ("B", work_b)):
            os.chdir(workdir)
            out = index_folder(path=str(project), use_ai_summaries=False,
                               storage_path="store")
            print(f"{label}: success={out.get('success')} "
                  f"symbol_count={out.get('symbol_count')} error={out.get('error')}")
            for db in sorted((workdir / "store").glob("*.db")):
                conn = sqlite3.connect(str(db))
                try:
                    tables = [r[0] for r in conn.execute(
                        "SELECT name FROM sqlite_master WHERE type='table'")]
                finally:
                    conn.close()
                print(f"{label}: {db.name} -> {len(tables)} tables, "
                      f"{db.stat().st_size} bytes")
    finally:
        os.chdir(cwd)

    print(f"_VERIFIED_PATHS  : {sorted(_VERIFIED_PATHS)}")
    print(f"_initialized_dbs : {sorted(SQLiteIndexStore._initialized_dbs)}")
```

</details>

<details>
<summary>2. Skipped migration chain (click to expand)</summary>

```python
"""Same relative spelling, two directories: the second database skips the migration chain.

Both stores start as v4 databases missing the v17 scip_* tables. The first
connect migrates its file to v17. The second connect finds the same key already
in ``SQLiteIndexStore._initialized_dbs`` and returns without running anything.
"""
import os, sqlite3, tempfile
from pathlib import Path

from jcodemunch_mcp.storage.sqlite_store import SQLiteIndexStore


def make_v4_db(path: Path) -> None:
    """A fully-schema'd database rewound to v4 with the v17 tables removed."""
    store = SQLiteIndexStore(base_path=str(path.parent))
    conn = store._connect(path)
    try:
        conn.executescript(
            "DROP TABLE IF EXISTS scip_edges;"
            "DROP TABLE IF EXISTS scip_unmapped;"
            "DROP TABLE IF EXISTS scip_meta;"
        )
        conn.execute(
            "INSERT OR REPLACE INTO meta (key, value) VALUES ('index_version', '4')"
        )
    finally:
        conn.close()
    SQLiteIndexStore._initialized_dbs.discard(str(path))


def state(path: Path) -> str:
    conn = sqlite3.connect(str(path))
    try:
        row = conn.execute(
            "SELECT value FROM meta WHERE key = 'index_version'"
        ).fetchone()
        tables = {r[0] for r in conn.execute(
            "SELECT name FROM sqlite_master WHERE type='table'"
        )}
    finally:
        conn.close()
    return f"index_version={row[0]:>2}  scip_edges={'scip_edges' in tables}"


with tempfile.TemporaryDirectory(prefix="jcm-migrate-") as tmp:
    root = Path(tmp).resolve()
    work_a, work_b = root / "work_a", root / "work_b"
    (work_a / "store").mkdir(parents=True)
    (work_b / "store").mkdir(parents=True)
    db_a = work_a / "store" / "local-sample.db"
    db_b = work_b / "store" / "local-sample.db"

    make_v4_db(db_a)
    make_v4_db(db_b)
    SQLiteIndexStore._initialized_dbs.clear()
    print(f"before  A : {state(db_a)}")
    print(f"before  B : {state(db_b)}")

    cwd = os.getcwd()
    try:
        for workdir in (work_a, work_b):
            os.chdir(workdir)
            store = SQLiteIndexStore(base_path="store")
            store._connect(store.base_path / "local-sample.db").close()
    finally:
        os.chdir(cwd)

    print(f"after   A : {state(db_a)}")
    print(f"after   B : {state(db_b)}")
    print(f"_initialized_dbs : {sorted(SQLiteIndexStore._initialized_dbs)}")
```

</details>

<details>
<summary>3. The mkdir half alone (click to expand)</summary>

```python
"""The _VERIFIED_PATHS half in isolation: the second directory is never created."""
import os, tempfile
from pathlib import Path

from jcodemunch_mcp.storage.sqlite_store import SQLiteIndexStore, _VERIFIED_PATHS

with tempfile.TemporaryDirectory(prefix="jcm-mkdir-") as tmp:
    root = Path(tmp).resolve()
    work_a, work_b = root / "work_a", root / "work_b"
    work_a.mkdir(); work_b.mkdir()

    cwd = os.getcwd()
    try:
        os.chdir(work_a)
        SQLiteIndexStore(base_path="store")
        print(f"A: store dir created = {(work_a / 'store').exists()}")

        os.chdir(work_b)
        store = SQLiteIndexStore(base_path="store")
        print(f"B: store dir created = {(work_b / 'store').exists()}")
        try:
            store._connect(store.base_path / "local-demo.db").close()
            print("B: connect ok")
        except Exception as exc:
            print(f"B: connect raised {type(exc).__name__}: {exc}")
    finally:
        os.chdir(cwd)

    print(f"_VERIFIED_PATHS : {sorted(_VERIFIED_PATHS)}")
```

</details>

## What else I checked

- **Does a restart make it worse?** No. I opened a leftover zero-byte `.db` in a new interpreter and
  got 15 tables, so the next process repairs it.
- **Can it reach a real server?** `git grep -n chdir 2e3883a -- src/` returns nothing; the only
  matches in the tree are seven files under `tests/`. Two servers launched from two directories have
  separate caches.
- **Is the default store exposed?** No. `Path.home() / ".code-index"` is already absolute.
- **Does anything catch an under-migrated store downstream?** Not that I found; the version check is
  one-sided. What the current code then does to a genuinely old file I did not establish.
- **Not tested:** hard links, junctions, case variants. This is about relative spellings, not
  filesystem aliasing in general.

## Expected vs actual

Actual: the key is the caller's spelling, so one key can name two directories.

Expected: the key identifies the directory. Three sites hold it, the two `_VERIFIED_PATHS` writes and
the `_initialized_dbs` check-and-add. Whether resolution belongs at the key or at `self.base_path` is
a real decision, since `base_path` is stored and returned elsewhere. I have not written the change or
any tests, so this is a reading of the constraint and not a patch behind it.

## Environment

- jCodeMunch version: `1.108.279` at `2e3883a`
- Python version: 3.13.7
- OS: Windows 11

mikemikimike:
## Summary

Fixes #475.

Relative storage paths were used verbatim as process-wide cache keys. If the working directory changed between two `SQLiteIndexStore` instances, the second store could reuse the first store's initialization state and skip directory creation or schema/migration setup.

## Changes

- Resolve configured storage paths before using them for `_VERIFIED_PATHS` and `_initialized_dbs` keys.
- Pass `IndexStore`'s resolved path to its SQLite backend.
- Add regression coverage for two relative stores across working-directory changes and for the resolved path handoff.

This preserves the existing behavior for absolute paths and default `~/.code-index` storage; it only makes relative paths stable for the lifetime of the process.

## Verification

- `PYTHONPATH=src python -m pytest tests/test_storage_path_resolution.py -q --basetemp .pytest-tmp` — 2 passed.
- `python -m ruff check src/jcodemunch_mcp/storage/sqlite_store.py src/jcodemunch_mcp/storage/index_store.py tests/test_storage_path_resolution.py` — passed.
- `git diff --cached --check` — passed.

The full suite was attempted with `PYTHONPATH=src python -m pytest tests -q`, after installing the declared runtime and development dependencies. Collection proceeded, but unrelated existing failures appeared across the suite; the run was stopped before completion. No build or benchmark command was run.


jgravelle:
Reviewed and verified. The fix is right and the tests are real.

Both keys were the caller's spelling, exactly as #475 describes, and resolving where the base path is built covers both stores. I checked non-vacuity per edit rather than per file: reverting either `.resolve()` alone turns `test_storage_path_resolution.py` red, so neither assertion is riding on the other.

Same defect class as v1.108.280, one layer up. That release resolved the perf-db path for this reason, because a cache keyed on a spelling is keyed on the caller's working directory. The index store was making that assumption about its own keys.

The `expanduser()` half is a second fix and worth naming separately. A `storage_path` written with a literal `~` (an MCP client config passing `CODE_INDEX_PATH: "~/foo"` never sees a shell) built a directory *named* `~` under the working directory here, while `process_registry.py` and `transcript_roots.py` expanded the same value against the real home. Anyone in that state has been running split across two locations. They now get the home one, and the tilde directory is not migrated. Correct behaviour, and it relocates an existing index, so it goes in the changelog rather than passing silently.

**Your branch turned four of our tests red and all four were already wrong.** Not your bug. Fixed on top in 503ae0b rather than sent back to you.

`patch("jcodemunch_mcp.tools.index_folder.Path.resolve", return_value=X)` replaces `resolve` on the `Path` *class*, so it answered for every path in the process, including the storage path `IndexStore` resolves on construction. Four tests get past the breadth guard, and each then created its index directory at the faked location. CI died in `mkdir` at `/workspaces/myrepo` and `\\server\share\`; a local Windows run left an empty `C:\work\project` outside any temp dir. Nothing in the suite could have reported that, because the writes went where no assertion was looking.

All 19 patch sites now go through `_resolve_only` in `tests/__init__.py`, with `autospec=True` so `self` reaches the side effect. The other fifteen were not correct either, only inert. Full suite on the merged tree: 7850 passed, 10 skipped, 0 failed, `ruff check src/` clean.

**One blocker, and it is not code: the CLA.** When I reviewed, this PR carried no `license/cla` status at all and CLA Assistant had not commented, which it did on other PRs here. Pushing 503ae0b woke it: the check is on the thread now and reads not signed. Signing takes about 30 seconds: https://cla-assistant.io/jgravelle/jcodemunch-mcp

**24 hours on that, to 2026-08-16.** If it does not clear, we implement the fix ourselves from #475 and our own v1.108.280, and you are credited in the CHANGELOG, the release notes and the close comment. The window decides whose commit it is. It does not decide whether you are credited, or whether this ships.

jgravelle:
Heads up: I pushed a merge of `main` onto this branch. The conflict was ours, not yours.

v1.108.282 went out today and its CHANGELOG entry landed in the same `[Unreleased]` block your #475 entry occupies, which put this PR into conflict. A conflicting fork PR has no merge ref, so it gets no CI at all until someone resolves it, and that would have taken your branch dark for a reason unrelated to your change. Resolving it on our side rather than sending you back for it.

The merge touches `CHANGELOG.md` only. Your entry stays on top under `[Unreleased]`; the release entry sits below it. Nothing in your code, tests, or the `_resolve_only` mock work was changed.

Your CLA signature is still the one thing outstanding. The status survived the push (still `pending` on the new head), so the link on the check should still work.

jgravelle (approved):
Verified locally: CLA status confirmed against the head SHA (state=success count=1), full matrix green.

The storage fix is the right shape - resolving where the base path is built means every consumer of the key gets it, same reasoning as v1.108.280's _perf_db_path.

The mock work is the more valuable half. Patching Path.resolve with a blanket return_value replaces resolve on the CLASS, so it answered for every path in the process - and the resulting directory writes landed where no assertion was looking, which is why the suite could never have reported it. autospec=True so self reaches the side effect is the detail that makes _resolve_only work at all. Converting the other fifteen inert sites was the right call; they were wrong too, just not yet reachable.

The expanduser() change and its unmigrated tilde directory are disclosed in the changelog rather than glossed. Thanks for both.

Issue #428, opened 2026-08-08
Rust const/static and Go const are declared in constant_patterns but never extracted: a file of only constants indexes as zero symbols


mussonking:
## What happened

I have a generated Rust file in one of my repos that is nothing but constants --
one `pub mod` per document type, 935 `pub const NAME: &str = "..."` inside them.
It's the contract layer of a data pipeline, so it's the single most useful file
in the repo to be able to search.

`index_folder` reports it in `no_symbols_files`. Zero symbols. Not "some
missing" -- zero.

I assumed it was the nested `pub mod` shape, so I checked an ordinary file:

```
apps/desktop/src/services/cra_dedup.rs
  functions: 17 indexed   (overlapping_core_lines_match, extract_core_lines, ...)
  types:      3 indexed
  constants:  0 indexed   (STATUS_DUPLICATE, PRIMARY_LINES, CORE_LINES, ...)
```

Same file, same parse. Functions come through, constants don't. So it isn't the
module nesting -- no Rust `const` is being indexed at all, anywhere.

## Cause

`RUST_SPEC` declares the node types:

```python
# parser/languages.py:442
constant_patterns=["const_item", "static_item"],
```

and the walker dispatches on them:

```python
# parser/extractor.py:541
if node.type in spec.constant_patterns and parent_symbol is None:
    const_symbol = _extract_constant(node, spec, source_bytes, filename, language)
```

but `_extract_constant` has no branch for either type. It handles `assignment`
(Python), `preproc_def` (C), `const_statement` (GDScript), `use_statement`
(Perl), `property_declaration` (Swift), `lexical_declaration` /
`variable_declaration` (JS/TS) -- and then falls off the end to `return None`.

Go has the same shape: `GO_SPEC` declares `constant_patterns=["const_declaration"]`
at languages.py:511, and there's no branch for that either.

Still present on 1.108.264 (I checked the wheel before writing this).

The part that cost me the most time is that there is no signal. A declared
pattern that is never implemented looks exactly like a language that has no
constants. `search_symbols` returns nothing and you conclude the symbol doesn't
exist, which is the one thing an index must never let you conclude wrongly. I
only caught it because I knew for a fact the file had 935 of them.

## Reproduce

```rust
// probe.rs
const TOP_LEVEL: &str = "a";
static ALSO_THIS: &str = "b";

pub mod inner {
    pub const NESTED: &str = "c";
}

pub fn visible() {}
```

[...]

Issue #382, opened 2026-07-26
Old tree sitter dependency?


kecsap:
Why does the software use this old tree sitter package?

"tree-sitter-language-pack>=0.7.0,<1.0.0"

Other code parser MCP tools happily use the newer versions >= 1.0.0

jgravelle:
Fair question, and the pin is deliberate rather than stale. I tested 1.13.3 against the full suite before answering, so here is what actually happens.

Three things break, in descending order of how much they matter.

**1.x fetches grammars over the network at first use. 0.x ships them in the wheel.**

This is the big one. The 1.x wheel contains a single native module and no bundled grammar bindings. `get_parser` resolves by downloading the grammar into a user-level cache. On a clean 1.13.3 install I watched that cache go from empty to 67 shared libraries as I walked our language list:

```
cache_dir: C:\Users\j\AppData\Local\tree-sitter-language-pack\v1.13.3\libs
downloaded count: 67
```

jCodeMunch is a read-only local code-exploration server. A first parse that reaches out to a remote manifest and writes executable objects to disk is a different security posture than the one the tool advertises, and it breaks airgapped and offline installs outright. We also have a standing rule that any new network or background behavior has to be disclosed in the README before it ships, so this would be a documented behavior change rather than a version bump. Not something to do as a dependency housekeeping commit.

**Three languages we support do not exist in 1.x.**

`autohotkey`, `ejs`, and `verse` all raise on 1.13.3:

```
Download error: Language 'autohotkey' not available for download.
```

Bumping the pin means silently dropping three languages.

**The Nim grammar was replaced with a different upstream, so our Nim extractor returns nothing.**

Same source, two versions:

```
0.13.0:  source_file > proc_declaration > identifier, parameter_declaration_list, ...
1.13.3:  module > stmt > routine > keyw, symbol > ident, paramList, ...
```

Different node vocabulary entirely, so our Nim symbol extraction yields zero symbols and `test_nim_parsing` fails. Fixable, but it is a rewrite of that extractor against a grammar we have not validated, not a pin edit.

Everything else is clean, which is worth saying plainly: **5812 passed, 12 skipped, 1 failed** on 1.13.3, and the one failure is Nim. The API surface we use from this package is exactly one symbol, `get_parser`, so there is no API incompatibility to speak of. If the download model were opt-in, or if we vendored the grammars ourselves, this would be an easy yes.

[...]

PR #485, opened 2026-08-17
Add MiniMax T2A text-to-speech backend for the voice and explainer pipelines

octo-patch:
Reason: The voice and narrated-explainer pipelines had no MiniMax T2A support, so neither regional endpoint, the speech-2.8 model family, the `model`/`text` request fields, nor the hex audio response could be used.

## What this changes

`gcm --voice` and `gcm explain` both synthesized speech by calling one hardcoded audio endpoint, with the model and voice as module constants in `groq/voice.py`. The request was built twice — once in `voice.speak`, once again in `explainer._render_tts`, which re-implemented its own client instead of reusing `_get_client`. There was no backend seam to extend, so this adds one.

New `src/jcodemunch_mcp/groq/tts.py`:

- Both regional T2A endpoints, `global_en` and `cn_zh`, in `MINIMAX_T2A_ENDPOINTS`.
- The full model family — `speech-2.8`, `speech-2.6`, `speech-02` and `speech-01`, each `-hd` and `-turbo` — with `speech-2.8-hd` as the default.
- `build_minimax_t2a_payload()` builds the request body. `model` and `text` are the two required fields; every optional field is omitted entirely when not supplied so the endpoint's own defaults apply rather than a copy of them drifting here. Payload keys are asserted to stay within the documented request-field set.
- `parse_minimax_t2a_response()` decodes the hex audio out of `data.audio`.
- `synthesize_minimax()` performs the synchronous `POST /v1/t2a_v2` operation with bearer auth, reusing the `httpx` already in the base dependencies — no new dependency.
- `MINIMAX_T2A_OPERATIONS` declares the async-create, async-query and WebSocket operations alongside the HTTP one.

`voice.speak` and `explainer._render_tts` now ask the backend first and fall back to the existing endpoint when nothing is configured, so default behaviour is unchanged. Selected with `JCODEMUNCH_TTS_PROVIDER=minimax`; auth uses `MINIMAX_API_KEY`, which is already in the recognised credential list, so no new credential env var is introduced.

## Three details worth review

**`output_format` is not the audio container format.** It selects how the response *carries* the audio (`hex` inlines it in `data.audio`, `url` returns a link that expires); the container lives in `audio_setting.format`, whose default is `mp3`. Both call sites read a WAV header — `speak` takes the sample rate, channel count and sample width from it, and `_render_tts` derives each slide's duration from the frame count — so treating `output_format` as the container would not have failed at the request. It would have surfaced as a `wave.Error` deep in playback, or as a video with wrong slide timing. The backend therefore always sends `audio_setting.format` explicitly and requests `hex`.

**An unknown region raises instead of falling back.** The two regions are separate deployments and a key issued for one is not valid on the other, so a typo'd region silently resolving to the default would read as an auth failure against a perfectly good key.

**A 200 can still carry a failure.** `base_resp.status_code` is checked before the audio is read, and `data` is nullable even on an otherwise successful envelope, so auth failures, rate limits and the invalid-character rejection surface as errors rather than as zero bytes of audio.

## Tests

`tests/test_groq_minimax_tts.py`, 40 tests: endpoint and region resolution, model-family validation, request-body construction, response parsing including the envelope-failure and null-`data` cases, and backend selection. Three of them cover the call-site wiring — one asserts the default client is never contacted once the backend answers, and `_render_tts` had **zero** coverage before this, which is how a duplicated request builder went unnoticed.

## Checks run

- `python3 -m pytest tests/test_groq_minimax_tts.py tests/test_groq_voice.py tests/test_groq_explainer.py` — 56 passed, 2 failed.
- `python3 -m pytest tests/test_claude_md_rotation.py tests/test_credentials.py tests/test_groq_minimax_tts.py tests/test_groq_voice.py` — 64 passed.
- `python3 -m ruff check src/` — all checks passed.

The 2 failures are both in `test_groq_explainer.py::TestGatherRepoInfo` and are pre-existing in my environment, not caused by this change: they fail identically on the unmodified tree because `tree_sitter_language_pack` is not installed here. I could not run the full suite for the same reason. No version bump and no CLAUDE.md "Current State" rotation, so the rotation gate stays green; the changelog entry went under `[Unreleased]`.


jgravelle:
Thanks for this — the full matrix is green, and the three details you flagged for review are the right three. The `output_format` one in particular would have been expensive to find later: both call sites parse a WAV header, so treating it as the container would have surfaced as a `wave.Error` in playback or as wrong slide timing, not as a request failure.

One blocker: the CLA isn't signed yet. CLA Assistant has prompted on this PR and signing takes about half a minute.

Being explicit about the clock and what happens when it runs out, because a deadline with an unstated consequence reads as a threat and this isn't one:

**You have 24 hours from this comment — until 2026-08-18 11:00 UTC.** Sign the CLA in that window and this merges as your commit. If it isn't signed by then, we implement the same change ourselves and credit you in the CHANGELOG, the release notes and the close comment. So the window decides whose commit it is. It never decides whether you're credited, and it never decides whether the feature ships.

If you want longer, say so on this thread and we'll hold it — CONTRIBUTING.md invites that by name, and the clock exists to stop work going quiet rather than to catch anyone out.

A substantive review follows separately; the clock is only about the signature.

jgravelle:
Substantive review, as promised. The shape is right: `synthesize()` returning `None` to mean "nothing configured" keeps the default path byte-for-byte, `httpx` is already a base dependency so nothing new is pulled in, and `MINIMAX_API_KEY` is already in the recognised credential list so no new credential surface appears. 40 tests. The de-duplication of the request build between `voice.speak` and `explainer._render_tts` is worth having on its own.

Three things before this can merge.

**1. README disclosure, and this is the blocking one.** This adds two new outbound hosts (`api.minimax.io`, `api.minimaxi.com`). The standing rule here is that every new network behaviour is README-disclosed before shipping — that rule exists because an undisclosed persistent service is what got this package quarantined on PyPI, and it is also why the `tree-sitter-language-pack` 1.x pin stays (#382: its `get_parser` downloads grammars at runtime). This PR documents the backend in GROQ.md and CLAUDE.md but does not touch README.md.

The egress here is opt-in and user-invoked rather than background, so it may not belong under README's "Background behavior, fully disclosed" heading. Either way the call should be explicit rather than made by omission. Add the line, or say on this thread why it does not apply and we will settle it here.

**2. `MINIMAX_T2A_OPERATIONS` declares three operations that nothing implements.** `textToAudioAsyncCreate`, `textToAudioAsyncQuery` and `textToAudioWebSocket` are in the dict; only `textToAudioHttp` is wired up.

We shipped a whole release on this exact hazard last week (1.108.281, #428): four languages declared `constant_patterns` with no branch implementing them, so `search_symbols` returning nothing read as an answer rather than as a gap. A declared capability with no implementation is indistinguishable from one that exists and returned nothing. Your module docstring does name the limitation, which is more than #428 had — but the constant is what a caller reads, and it does not carry the caveat.

Preference: keep only the implemented operation in the dict and put the other three in a comment, or name the constant so the status is in the name. Either removes the ambiguity at the point of use.

**3. A missing `base_resp` reads as success.** `parse_minimax_t2a_response` does `base_resp.get("status_code", MINIMAX_BASE_RESP_OK)`, so an envelope with no `base_resp` at all is treated as OK.

The convention in this codebase runs the other way: unknown must never resolve to the confident answer. `FreshnessProbe.classify()` returns `unknown` for any comparison it could not make, and `embedding_store.has_any()` is tri-state precisely so "could not establish" is never `False`. It is mild here because the `audio_hex` check catches a genuinely empty response a few lines later, so the failure mode is a worse error message rather than a wrong answer. Still the wrong default.

Minor, no action needed: the `unknown = set(payload) - set(MINIMAX_T2A_REQUEST_FIELDS)` check cannot fire today, since every key in `payload` is set by the function itself from a fixed set. It is a reasonable drift guard for later edits — worth a short comment saying that, so a later reader does not mistake it for validation of caller input.

None of this is large. The CLA remains the only hard blocker, on the clock posted above.

octo-patch:
Thanks for the review. I made all three changes as requested and pushed a new commit.

1. **README disclosure.** README.md now names both new outbound hosts (`api.minimax.io` for `global_en`, `api.minimaxi.com` for `cn_zh`) in the security and privacy section, stated explicitly as opt-in and user-invoked rather than background, and noting that neither host is contacted unless `JCODEMUNCH_TTS_PROVIDER=minimax` is set. I made the call rather than leaving it to omission, and kept it out of the background-behavior enumeration since the request only happens on an explicit `gcm` invocation.

2. **Declared vs implemented operations.** Split into `MINIMAX_T2A_IMPLEMENTED_OPERATIONS`, which now contains only `textToAudioHttp`, and `MINIMAX_T2A_UNIMPLEMENTED_OPERATIONS` for the async-create, async-query and WebSocket entries. The status is now in the name at the point of use, the endpoint paths stay discoverable, and a test asserts the two sets are disjoint so nothing can dispatch to an unimplemented entry.

3. **Missing `base_resp` no longer reads as success.** `parse_minimax_t2a_response` raises when `base_resp` is absent or carries no `status_code`, instead of defaulting to OK, so an outcome that cannot be established is never resolved to the confident answer. Two tests cover it: no envelope at all, and an envelope with only a `status_msg`. I also added the comment marking the request-field check as a drift guard for later edits rather than validation of caller input.

Checks: `ruff check src/` clean, and `pytest tests/test_groq_minimax_tts.py tests/test_groq_voice.py` green at 43 + 7. Two `test_groq_explainer.py` cases fail in my sandbox on a missing `tree_sitter_language_pack` install; they fail identically on the previous commit, so they are environmental rather than from this change.

_(Edited to correct a test count: the voice suite is 7 tests, not 16 as first posted.)_

jgravelle:
Merged `main` in and resolved the CHANGELOG conflict on this branch. Head is `54ccf44`. Your section stays under `## [Unreleased]`; the released `1.108.283` block sits below it.

The conflict came from our release landing in the same `[Unreleased]` block your entry occupies. A conflicting fork PR has no `refs/pull/N/merge`, so it gets no CI at all — which is why zero checks had run on your latest head. That is cleared now.

All three changes check out. I verified each in the diff rather than reading the summary:

- README names both hosts with the opt-in and user-invoked framing, and keeping it out of the background-behaviour enumeration is the right call — you stated the reason, which is what I asked for.
- `MINIMAX_T2A_IMPLEMENTED_OPERATIONS` holds only `textToAudioHttp`, with the rest in the unimplemented set.
- `parse_minimax_t2a_response` raises both when `base_resp` is absent and when it carries no `status_code`, with a test for each.

Your `test_groq_explainer.py` diagnosis was right too — both cases pass here with `tree-sitter-language-pack` installed, so they were environmental as you said.

**One question, asked directly: are you affiliated with MiniMax?**

Why I'm asking. This account has roughly 3,000 pull requests across GitHub, nearly all of them MiniMax integrations. The standing rule here for adding a named vendor is that a *user* asks for it — that is exactly how MiniMax became a summarizer provider in this repo, via [#184](https://github.com/jgravelle/jcodemunch-mcp/issues/184). No one has asked for TTS.

There is no wrong answer and nothing turns on it being one or the other by itself. A comparable PR here earlier today opened by stating the author's affiliation up front, and it cost them nothing. I would rather have it on the record than infer it from a search.

**The clock is unchanged: 2026-08-18 11:00 UTC**, with the default exactly as posted above. The CLA is still the only hard blocker and still unsigned.

jgravelle:
Closing this. Not on quality, and the quality is worth stating precisely because it is not the reason.

The `output_format` finding is real: it selects how the response *carries* the audio rather than the container, and both call sites parse a WAV header, so getting it wrong would have surfaced as a `wave.Error` deep in playback instead of a request failure. You addressed a three-point review within hours, split the implemented and unimplemented operation sets so status lives at the point of use, made `parse_minimax_t2a_response` refuse to resolve an unestablished outcome to success, and corrected your own test count unprompted. 43 tests, green on a full run here. Collapsing the request builder duplicated between `voice.speak` and `explainer._render_tts` is a genuine improvement independent of the backend. None of that wasted my review time.

**The reason is demand.** Adding a named vendor here requires a user to ask for it. That is exactly how MiniMax became a summarizer provider in this repo — [#184](https://github.com/jgravelle/jcodemunch-mcp/issues/184), filed by someone using the tool. Nobody has asked for text-to-speech; searching the tracker, the only mention of TTS is this pull request.

**The second half is the profile.** This account carries 3,089 pull requests, 2,242 of them with "minimax" in the title alone, at roughly nineteen a day since March, while listing `company: Independent Developer`. I asked the affiliation question directly above and I will leave it there rather than answer it for you. What I will say is that the pattern and that field do not agree, and a named-vendor addition is precisely where that matters.

**Two things stated rather than left to inference.**

This was never going to merge unsigned. The CLA is absolute and nothing on this thread suggested otherwise.

And I am retracting half of what I posted earlier. That comment said that if the CLA went unsigned we would implement the same change ourselves, and that the window "never decides whether the feature ships." **The authorship-and-credit half stands** — the `output_format` distinction is your finding, and it gets credited if a TTS seam is ever built here. **The feature half I am withdrawing.** It was written before I understood the shape of this, and whether jCodeMunch grows a MiniMax TTS backend is a product decision that belongs on user demand rather than on a PR arriving. Saying so openly beats letting the promise quietly lapse.

If a jCodeMunch user turns up wanting TTS through MiniMax, this reopens on their ask.

Issue #372, opened 2026-07-21
Indexer descends into nested worktrees under `.claude/worktrees/`, polluting the parent index


bitfliq:
**Version:** 1.108.150

Claude Code creates git worktrees inside the repo at `<repo>/.claude/worktrees/<name>` (the convention the tool documents via `worktree_base_path`). When indexing or running `watch-all` on the repo root, the directory walk appears to descend into those worktree checkouts and pull their files into the parent repo's index.

## What I saw

- **Index pollution + cap truncation.** Each worktree is a near-duplicate checkout, so per-language file counts scale by roughly `(worktrees + 1)`. With a handful of worktrees this pushes the parent past `max_folder_files` (default 2000), and real parent files get silently dropped — the only signal is a one-line warning.
- **Identity collision under a git identity.** Running `watch-claude` alongside `watch-all` fails because each nested worktree resolves (by path containment) to the parent's `owner/name` slot, so both try to own the same index: `Index '<owner>/<repo>' already exists at '.../.claude/worktrees/...'. Indexing '<repo>' would overwrite it.` During this the index dropped to 0 symbols.

## Likely cause

The walk prunes only the hardcoded skip-dir list plus `.gitignore`; it doesn't treat a nested `.git` file (a linked worktree) as a boundary, and neither `.claude` nor `worktrees` is in the default skip list. Interestingly the identity layer already knows about worktrees (it handles a `.git` file as a working-tree marker) — the walker just doesn't use that knowledge.

## Suggested fixes (in rough order of preference)

1. **Skip nested working trees in the walk.** When descending, stop at any subdirectory that is a distinct working tree — e.g. a `.git` file/dir whose `git rev-parse --git-common-dir` differs from the root being indexed. This is the general fix and handles worktrees wherever they live.
2. **Add `.claude/worktrees` to the default skip list.** Cheap and covers the Claude Code convention specifically, though it misses worktrees configured elsewhere.
3. **Have `watch-all` reuse the same nested-worktree exclusion** so it coexists with `watch-claude` out of the box regardless of identity mode.

## Workaround

Setting `extra_ignore_patterns: [".claude/worktrees/"]` keeps the parent index clean and, since patterns are matched relative to each index's own root, doesn't stop `watch-claude` from indexing the worktrees separately.

jgravelle:
[...]

Issue #303, opened 2026-05-17
resolve_repo can time out before returning canonical worktree candidates in large agent-worktree environments


rknighton:
Thanks for the work in #277. Canonical worktree reuse is the right direction, and this report is a narrower edge case rather than a claim that the #277 behavior is broadly broken.

## Environment

- Windows
- jcodemunch-mcp 1.108.13
- git_root_identity enabled
- About 130 git worktrees
- About 40 JCodeMunch index DBs
- Mostly temporary agent-created worktrees for the same logical repository

This is an atypical and arguably suboptimal environment. Most users will not have 100+ worktrees for one repo. But agent/automation systems can accumulate this shape naturally, and the failure mode is severe: agents lose access to JCodeMunch and fall back to broad local reads.

## Related Issue

Related: #277, “Worktree-aware canonical index reuse for Codex, VS Code, and generic MCP clients.”

## Observed Behavior

- jcodemunch_guide succeeds immediately.
- resolve_repo on the canonical checkout succeeds quickly.
- resolve_repo on an unindexed agent worktree path times out after the MCP client timeout.

Example path shape:

    C:\Users\<user>\.agent\worktrees\<id>\<repo>

## Expected Behavior

resolve_repo should quickly recognize the worktree and return canonical_candidates for the already-indexed canonical checkout.

Example response shape:

    {
      "found": true,
      "indexed": false,
      "repo": "local/<worktree-derived-id>",
      "hint": "this is a Git worktree of an already-indexed repo...",
      "canonical_candidates": [
        {
          "repo": "<canonical-repo-id>",
          "source_root": "<stable-checkout-path>",
          "rationale": "shared --git-common-dir"
        }
      ]
    }

## Actual Behavior

resolve_repo times out before returning. In agent workflows, that turns a recoverable lookup miss into a tool-availability failure: the agent never sees the canonical candidate and may fall back to broad local filesystem reads.

## Root Cause Hypothesis

resolve_repo appears to run expensive identity resolution for the unindexed worktree path before reaching canonical candidate discovery. At this scale, that can exceed the MCP timeout. Per-index git subprocess calls for common-dir comparison may also contribute.

## Local Patch That Fixed It

A local patch made resolve_repo fast again by:

1. Loading repo metadata once.
2. Exact-matching indexed source_root first.
3. Using cheap local path identity for unindexed paths.
4. Canonical-matching worktrees by git common-dir.
[...]

PR #497, opened 2026-08-18
fix(resolve_repo): stop containment resolving across a nested repo boundary (#492)

### LINKED ISSUE #492: `resolve_repo` matches on `source_root` containment before any repository-identity check, so a path inside a nested git repository resolves to the enclosing parent index

`resolve_repo` returns the enclosing parent index, as `indexed: true`, for a path
that belongs to a different git repository, without ever establishing that the
path and the returned repository are the same repository.

## Reproduction

A parent repository with a second, independent git repository checked out inside
it. Case 1 is the failure. Case 2 is the same mis-resolution with the read
happening to succeed. Case 3 is the control.

```python
"""resolve_repo returns a parent index for a path in a DIFFERENT git repository."""
import json, os, shutil, subprocess, sys, tempfile
sys.path.insert(0, "src")

from jcodemunch_mcp.tools.resolve_repo import resolve_repo
from jcodemunch_mcp.tools.get_file_content import get_file_content

def git(cwd, *a):
    return subprocess.run(["git", *a], cwd=cwd, check=True,
                          capture_output=True, text=True).stdout.strip()

def make_repo(path, files, gitignore=None):
    os.makedirs(path, exist_ok=True)
    git(path, "init", "-q")
    git(path, "config", "user.email", "repro@example.invalid")
    git(path, "config", "user.name", "repro")
    for n, t in files.items():
        with open(os.path.join(path, n), "w", encoding="utf-8") as fh:
            fh.write(t)
    if gitignore:
        with open(os.path.join(path, ".gitignore"), "w", encoding="utf-8") as fh:
            fh.write(gitignore)
    git(path, "add", "-A"); git(path, "commit", "-q", "-m", "init")

def run(label, gitignore_nested, index_nested, read_path="nested-repo/alpha.py"):
    work = tempfile.mkdtemp(prefix="jcm-resolve-")
    store = os.path.join(work, "index")
    parent = os.path.join(work, "parent")
    nested = os.path.join(parent, "nested-repo")
    os.makedirs(store)
    make_repo(parent, {"marker.py": "def marker():\n    return 1\n"},
              gitignore="nested-repo/\n" if gitignore_nested else None)
    make_repo(nested, {"alpha.py": "def alpha():\n    return 1\n"})
    index_folder(path=parent, identity_mode="local", use_ai_summaries=False,
                 storage_path=store)
    if index_nested:
        index_folder(path=nested, identity_mode="local", use_ai_summaries=False,
                     storage_path=store)

    target = os.path.join(nested, "alpha.py")
    r = resolve_repo(path=target, storage_path=store)
    print(f"\n--- {label} ---")
    print(f"  repo returned          : {r.get('repo')}")
    print(f"  indexed / loadable     : {r.get('indexed')} / {r.get('loadable')}")
    print(f"  match_path             : {(r.get('_meta') or {}).get('match_path')}")
    print(f"  'canonical_candidates' in response: {'canonical_candidates' in r}")

    got = get_file_content(repo=r["repo"], file_path=read_path,
                           storage_path=store)
    print(f"  read of {read_path!r} succeeded: {bool(got.get('content'))}")
    verdict = (got.get("_meta") or {}).get("verdict")
    if verdict is not None:
        print(f"  verdict.state          : {verdict.get('state')}")
    shutil.rmtree(work, ignore_errors=True)

run("Case 1  nested repo gitignored by the parent", True, False)
run("Case 2  nested repo NOT gitignored (parent absorbs its files)", False, False)
run("Case 3  CONTROL: nested repo has its own index", True, True, read_path="alpha.py")
```

Output at `b85ef61` (the `local/...` suffixes derive from the temporary path and
differ per run; nothing else does):

```text
--- Case 1  nested repo gitignored by the parent ---
  repo returned          : local/parent-d2002c7d
  indexed / loadable     : True / True
  match_path             : source_root_containment
  'canonical_candidates' in response: False
  read of 'nested-repo/alpha.py' succeeded: False
  verdict.state          : absent

--- Case 2  nested repo NOT gitignored (parent absorbs its files) ---
  repo returned          : local/parent-025508d0
  indexed / loadable     : True / True
  match_path             : source_root_containment
  'canonical_candidates' in response: False
  read of 'nested-repo/alpha.py' succeeded: True
  verdict.state          : ok

--- Case 3  CONTROL: nested repo has its own index ---
  repo returned          : local/nested-repo-4730b2f2
  indexed / loadable     : True / True
  match_path             : exact_source_root
  'canonical_candidates' in response: False
  read of 'alpha.py' succeeded: True
  verdict.state          : ok
```

Cases 1 and 2 resolve identically and wrongly. The difference is only whether the
mistake is visible: in case 2 the parent's walk happened to absorb the nested
repo's files, so the read succeeds and nothing looks amiss. Case 3 shows the
resolver does the right thing as soon as an exact index exists, which controls for
the path and the repository both being valid.

## Expected vs actual

**Expected.** A path belonging to a nested git repository with no index of its own
either resolves to an index that represents that repository, or reports the
repository as unindexed.

**Actual.** The parent index is returned with `found: true`, `indexed: true`,
`loadable: true` and `match_path: source_root_containment`. The nested repository
is never identified as a distinct repository, so the caller binds to a corpus
built from a different checkout with a different history.

Worth being precise about the failed read in case 1, because it is not itself a
bug: `nested-repo/` is gitignored by the parent, and a gitignore exclusion is a
case where the corpus is being defined and a search over it can still prove
absence — the position recorded at
[index_folder.py#L100-L111](https://github.com/jgravelle/jcodemunch-mcp/blob/b85ef61ca767ff681fdc74e7f124e461e4829464/src/jcodemunch_mcp/tools/index_folder.py#L100-L111).
The `absent` verdict is correct for the corpus that was searched. The defect is
that the corpus searched was the wrong one, and that decision was made before any
identity check ran. Case 2 makes the point without involving absence semantics at
all: the same wrong repository is returned there too.

## Mechanism

| Site | Lines |
|---|---|
| Containment scan, deepest `source_root` wins | [resolve_repo.py#L247-L283](https://github.com/jgravelle/jcodemunch-mcp/blob/b85ef61ca767ff681fdc74e7f124e461e4829464/src/jcodemunch_mcp/tools/resolve_repo.py#L247-L283) |
| Canonical worktree discovery, reached only afterwards | [resolve_repo.py#L285-L292](https://github.com/jgravelle/jcodemunch-mcp/blob/b85ef61ca767ff681fdc74e7f124e461e4829464/src/jcodemunch_mcp/tools/resolve_repo.py#L285-L292) |
| `_build_indexed_response` | [resolve_repo.py#L136-L158](https://github.com/jgravelle/jcodemunch-mcp/blob/b85ef61ca767ff681fdc74e7f124e461e4829464/src/jcodemunch_mcp/tools/resolve_repo.py#L136-L158) |

Fast path 1 returns on `Path.is_relative_to(source_root)` alone. Nothing in that
branch consults git identity, and `_build_indexed_response` returns `None` only
when the repo id is malformed or `index_present` is false — its `p_resolved`
parameter is never read. `indexed` is set from `status.loadable`, which is a
property of the index, not of the requested path.

The ordering is deliberate: fast path 1 exists to avoid the `resolve_index_identity`
walk that could hang, per the issue cited as `jcm#303` in the comment. The gap is
that it has no cheap correctness guard, only a speed guarantee.

A guard that would fit the existing design is a check for an intervening `.git`
between the requested path and the matched `source_root`. That is a filesystem
stat, not a subprocess, so it keeps fast path 1 fast. It must exclude submodules,
whose `.git` file points at `.git/modules/<name>` and which are deliberately
indexed into the parent — the distinction already drawn in `is_linked_worktree`
([git_root.py#L226-L247](https://github.com/jgravelle/jcodemunch-mcp/blob/b85ef61ca767ff681fdc74e7f124e461e4829464/src/jcodemunch_mcp/storage/git_root.py#L226-L247)).

Two things I checked that do not help here. `canonical_candidates` is not a
fallback for this case: `_find_canonical_candidates` matches on a shared git
common-dir, and an independent nested clone shares none, so it returns `[]`. And
the key is absent from the response entirely rather than present and empty.

## Environment

- jCodeMunch: commit `b85ef61ca767ff681fdc74e7f124e461e4829464`, which declares version `1.108.284`
- Run from the checkout with `sys.path.insert(0, "src")`, not an installed wheel
- Python 3.13.7
- Windows 11 Pro 26200
- Repositories under test: parent 2 files, nested 1 file, created by the harness above (throwaway, deleted at exit)

## Duplicate check

Searched all 283 issues, open and closed, over title and body, for `resolve_repo`,
`source_root_containment`, `containment`, `nested repo`, and `nested git`.

No duplicate. The two nearest neighbours are a different boundary:

- [Indexer descends into nested worktrees under `.claude/worktrees/`, polluting the parent index](https://github.com/jgravelle/jcodemunch-mcp/issues/372) is indexing-side. Its fix prunes linked worktrees via `is_linked_worktree`, which requires `.git` to be a file; an ordinary nested clone has a `.git` directory and is not pruned by it.
- [resolve_repo can time out before returning canonical worktree candidates in large agent-worktree environments](https://github.com/jgravelle/jcodemunch-mcp/issues/303) is the antecedent, and my own report. The containment fast path was added to avoid that timeout; this is the case that ordering admits.

## Acceptance criteria

- [ ] For a file inside a nested independent git repository with no index of its own, `resolve_repo` does not return the enclosing parent index with `indexed: true`.
- [ ] That holds whether or not the nested repository's files were absorbed into the parent corpus, so case 2 above changes result along with case 1.
- [ ] When the nested repository has its own index, `resolve_repo` still returns it via `exact_source_root`.
- [ ] A path inside a **submodule** still resolves to the parent index, since submodule content is indexed into the parent and #372 fixed worktree exclusion specifically without changing submodule behaviour.
- [ ] A file that is inside the parent repository but absent from its corpus (gitignored, oversize, skipped directory) still resolves to the parent. Being outside the corpus and belonging to another repository are different conditions, and only the second changes which repository is returned.
- [ ] An ordinary path that genuinely belongs to the indexed parent still resolves by containment with no added subprocess call.

## Suggested metadata

Labels: `bug`, `resolver`.

Priority: medium. It needs a nested independent repository inside an indexed
parent. The mis-resolution happens whenever that layout exists, but it only
becomes visible when the nested repository's files are outside the parent corpus,
which is why it can sit unnoticed.

jgravelle:
Closes #492.

> Stacked on #496 (`fix/493-index-file-head-advance`) so the CHANGELOG entries do not conflict. The base retargets to `main` automatically when that merges; review only the second commit.

## What was wrong

Fast path 1 matched on `source_root` containment alone, so a path inside an **independent git repository nested in an indexed parent** came back as the parent index with `indexed: true` and `match_path: source_root_containment`. Nothing in that branch consulted git identity. The caller was bound to a corpus built from a different checkout with a different history.

**Whether it looks wrong depends on something irrelevant to the defect** — your case 1 versus case 2 is the cleanest statement of it. Gitignored, the read fails and returns `absent`, correct for the corpus that was searched and indistinguishable from a normal empty result. Absorbed into the parent walk, the same wrong repository is returned and the read *succeeds* with `state: ok`. Both change here.

## The fix

A `.git` stat between the requested path and the matched `source_root` — the guard you proposed.

- **No subprocess.** Fast path 1 exists to avoid the `resolve_index_identity` walk that can hang (#303), so a correctness guard on it must not reintroduce a process spawn. A test monkeypatches `subprocess.run` to raise and asserts the ordinary containment path still resolves.
- **Classify by where `.git` points, not by whether it is a file.** `.git/worktrees/<name>` versus `.git/modules/<name>` is #372's distinction. **Submodules still resolve to the parent** (criterion 4), since their content is indexed into it. Linked worktrees are also left alone — fast path 2 already answers for them via `canonical_candidates`, and diverting them here would change #303's answer. `git clone --separate-git-dir` leaves a `.git` file pointing at neither and counts as independent; a file/directory test would have read it as a submodule.
- **A file outside the parent's corpus still resolves to the parent** (criterion 5). Gitignored, oversize or skipped is not the same condition as belonging to another repository.

## Verification

Your three-case reproduction against this branch:

```text
--- Case 1  nested repo gitignored by the parent ---
  repo returned          : local/nested-repo-d1a166da
  indexed / loadable     : False / None
  match_path             : not_indexed

--- Case 2  nested repo NOT gitignored (parent absorbs its files) ---
  repo returned          : local/nested-repo-d81734b0
  indexed / loadable     : False / None
  match_path             : not_indexed

--- Case 3  CONTROL: nested repo has its own index ---
  repo returned          : local/nested-repo-5f04d30e
  indexed / loadable     : True / True
  match_path             : exact_source_root
```

Both failing cases now identify the nested repository and report it unindexed, rather than returning a different one as indexed. Case 3 is unchanged.

`tests/test_resolve_repo_nested_repo_boundary.py`, 11 tests. **Seven red against `b85ef61`.** The remaining four are criteria 4–6 and pass on both sides by design. Submodules and linked worktrees are tested against real git layouts, not fabricated `.git` markers.

Suite: **7923 passed, 17 skipped, 0 failed**, ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,51 @@
 
 ## [Unreleased]
 
+### `resolve_repo` no longer answers a repository question with a filesystem fact (#492, @rknighton)
+
+Fast path 1 matched on `source_root` containment alone, so a path inside an
+**independent git repository nested in an indexed parent** came back as the
+parent index with `found: true`, `indexed: true`, `loadable: true` and
+`match_path: source_root_containment`. Nothing in that branch consulted git
+identity. The caller was bound to a corpus built from a different checkout with
+a different history.
+
+⚠⚠ **Whether it looks wrong depends on something irrelevant to the defect.**
+When the nested repository is gitignored by the parent, the read fails and
+returns `absent` — correct for the corpus that was searched, and the reason it
+reads as a normal empty result. When the parent's walk absorbed the nested
+files, the same wrong repository is returned and **the read succeeds with
+`state: ok`**. The mis-resolution is identical; only the symptom differs. Both
+cases change here.
+
+The guard is a `.git` stat between the requested path and the matched
+`source_root`. ⚠ **A subprocess would have been the wrong fix at the right
+place**: fast path 1 exists precisely to avoid the `resolve_index_identity`
+walk that can hang in large agent-worktree environments (#303), so a
+correctness guard on it must not reintroduce a process spawn. A test asserts no
+`subprocess.run` on the ordinary containment path.
+
+⚠ **Classify by where `.git` POINTS, not by whether it is a file.** A `.git`
+directory is an independent repository; a `.git` file carries a `gitdir:`
+pointer, and `.git/worktrees/<name>` versus `.git/modules/<name>` is exactly the
+distinction #372 drew. **Submodules deliberately still resolve to the parent** —
+their content is indexed into it, and #372 excluded linked worktrees without
+changing that. Linked worktrees are also left alone: fast path 2 already answers
+for them via `canonical_candidates`, and diverting them here would change #303's
+answer. `git clone --separate-git-dir` leaves a `.git` file pointing at neither,
+and counts as independent — a file/directory test would have read it as a
+submodule.
+
+⚠ **A file outside the parent's corpus still resolves to the parent.** Being
+gitignored, oversize or in a skipped directory is not the same condition as
+belonging to another repository, and only the second changes which repository is
+returned.
+
+⚠ `tests/test_resolve_repo_nested_repo_boundary.py` (11). Seven are red against
+`b85ef61`; the remaining four are the boundary tests above and pass on both
+sides by design.
+
+
 ### A single-file refresh no longer certifies a corpus it never re-read (#493, @rknighton)
 
 `index_file` wrote live HEAD as the repository's stored SHA. `repo_is_stale` is
--- a/src/jcodemunch_mcp/tools/resolve_repo.py
+++ b/src/jcodemunch_mcp/tools/resolve_repo.py
@@ -13,6 +13,76 @@ from ..storage.git_root import resolve_index_identity
 logger = logging.getLogger(__name__)
 
 
+def _dotgit_kind(directory: Path) -> Optional[str]:
+    """Classify a ``.git`` entry: "repo", "submodule", "worktree", or None.
+
+    A ``.git`` DIRECTORY is an independent repository. A ``.git`` FILE carries a
+    ``gitdir:`` pointer, and where it points is the whole distinction #372 drew:
+    ``.git/worktrees/<name>`` is a linked worktree, ``.git/modules/<name>`` is a
+    submodule. Anything else pointing outside both is a ``--separate-git-dir``
+    clone, which is independent.
+    """
+    dotgit = directory / ".git"
+    try:
+        if dotgit.is_dir():
+            return "repo"
+        if not dotgit.is_file():
+            return None
+        text = dotgit.read_text(encoding="utf-8", errors="ignore").strip()
+    except OSError:
+        return None
+    if not text.startswith("gitdir:"):
+        return None
+    target = Path(text[len("gitdir:"):].strip())
+    if not target.is_absolute():
+        target = directory / target
+    parent_name = target.parent.name
+    if parent_name == "worktrees":
+        return "worktree"
+    if parent_name == "modules":
+        return "submodule"
+    return "repo"
+
+
+def _independent_repo_between(p_resolved: Path, source_root: Path) -> Optional[Path]:
+    """Root of an independent git repository between a path and a source root.
+
+    Returns the deepest such root, or None when the path really does belong to
+    ``source_root``'s repository.
+
+    ⚠⚠ Containment is a statement about the FILESYSTEM; the caller wants one
+    about the REPOSITORY. A nested independent clone inside an indexed parent
+    satisfies the first and fails the second, and before #493's sibling #492 the
+    resolver returned the parent as ``indexed: true`` for it, binding the caller
+    to a corpus from a different checkout with a different history.
+
+    ⚠ Submodules deliberately do NOT count. Their content IS indexed into the
+    parent, which is the boundary #372 drew when it excluded linked worktrees
+    without changing submodule behaviour. Linked worktrees are left alone here
+    too: fast path 2 already answers for them via ``canonical_candidates``, and
+    diverting them into this branch would change #303's answer.
+
+    ⚠ Stats only, no subprocess — fast path 1 exists to avoid the
+    ``resolve_index_identity`` walk that can hang (#303), so a correctness guard
+    on it must not reintroduce a process spawn.
+    """
+    try:
+        candidates = [p_resolved, *p_resolved.parents]
+    except (OSError, ValueError):
+        return None
+    for directory in candidates:
+        if directory == source_root:
+            return None  # reached the indexed root without crossing a boundary
+        try:
+            if not directory.is_relative_to(source_root):
+                return None
+        except (OSError, ValueError, AttributeError):
+            return None
+        if _dotgit_kind(directory) == "repo":
+            return directory
+    return None
+
+
 def _compute_repo_id(folder_path: Path, store: Optional[IndexStore] = None) -> str:
     """Compute the repo ID that index_folder would use for a directory path."""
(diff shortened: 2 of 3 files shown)


jgravelle:
Superseded by #498 — same branch, same commit ('230bad6'), retargeted at `main`.

Closed by mechanism, not by choice: merging #496 with `--delete-branch` removed this PR's base branch, and GitHub closes a stacked PR when its base disappears rather than retargeting it. A closed PR's base cannot be changed and it cannot be reopened while the base is gone, so #498 is the recovery. Nothing about the change itself moved.

PR #520, opened 2026-08-20
test(license): the version suffix and the LICENSE version imply each other

jgravelle:
Taking @marcelruhf's improvement back from the sibling repos, as promised on jdocmunch-mcp#122.

`test_the_version_suffix_tracks_the_license_file` (mine, #518) asserted that the identifier carries a `-X.Y` suffix **unconditionally**. That is only correct while this repo's LICENSE happens to state a version. He ported the ratchet into jdocmunch-mcp and jdatamunch-mcp, whose LICENSE files carry no `Version` line, and wrote the general form instead: a version line and a suffix **imply each other, in both directions**.

The half only his version catches is the one nobody would think to test — an identifier claiming a licence version the text does not state. Proven by temporarily editing LICENSE:

- `Version 1.1` → `Version 1.2`, identifier unchanged: **fails** (the case mine also caught)
- `Version` line removed, identifier still `-1.1`: **fails** (the case mine missed)

LICENSE restored byte-for-byte after both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## [Unreleased]
 
+### The license-identifier ratchet asserted this repo's accident (@marcelruhf)
+
+`test_the_version_suffix_tracks_the_license_file` required the identifier to
+carry a `-X.Y` suffix unconditionally, which is only correct while this LICENSE
+happens to state a version. @marcelruhf ported the ratchet into jdocmunch-mcp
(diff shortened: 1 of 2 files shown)


marcelruhf (approved):
The assertion body matches what we shipped on jdatamunch-mcp #4 and jdocmunch-mcp #122: a Version line and a suffix imply each other, both ways. No concerns.

PR #532, opened 2026-08-21
A baseline gets as many guesses as the system it is the floor for

jgravelle:
Three findings from digging into whether `route@1` is the right objective. The
short answer is that **`@1` versus `@3` was the wrong axis**, and the reading I
had been leaning toward — that `@3` is the number that clears the bar — is
refuted by the corpus's own data.

## 1. The `@3` floor was mis-specified in route's favour

`run_emitted_task.py` reported route's `@3` recall against the best constant
**single** action, and a comment in the script argued this was fair because "the
floor is RANK-INVARIANT… route@3-vs-floor is the one that matters." **A baseline
gets as many guesses as the system it is the floor for.**

Against the best constant 3-**set**, same 40 cases:

| | route | best constant 3-set | delta |
|---|---|---|---|
| strict@3 | 62.5% | 92.5% | **−30.0** |
| exact@3 | 70.0% | 100% | **−30.0** |
| family@3 | 70.0% | 100% | **−30.0** |

`strict@3` moves from **+17.5** against the 1-set floor to **−30.0** against the
k-matched one. **`@3` deepens the emitted-task result rather than rescuing it.**
Both floors are emitted by the harness now (`blind_floor_kset`,
`vs_kset_floor_pts`), computed rather than hand-derived.

⚠⚠ **The emitted corpus cannot discriminate a router from a fixed list.** Its
best constant 3-set scores **100% exact** — 40 cases, 6 distinct primary labels,
87.5% in one family; the holdout is 20 cases and 3 labels. That is a property of
the sample, not a verdict on route, and it is asserted so it stays visible.

⚠ The human harness reported **no floor at all**; its 1-set number existed only
in the *other* script's docstring. It now reports k-matched floors: `@1` 71.2% vs
5.1% (**+66.1**), `@3` 86.4% vs 13.6% (**+72.8**), over 59 queries with 69
distinct targets. **Route routes decisively on human phrasing at both k. What it
does not do is route on agent wording.**

## 2. `results.json` had drifted, and `CLAUDE.md` cited it

Committed: `route@1 69.5 / @3 88.1`. Fresh run, same harness, same corpus:
`71.2 / 86.4`. Two queries moved — one gained rank 1, one fell out of the top 3 —
because tool descriptions changed under a frozen artifact. `catalog_actions` was
unchanged at 91, so nothing about size gave it away.

`CLAUDE.md` cites `route@1 69.5%` as the evidence that moratorium condition 1 is
met. The verdict is unchanged — 71.2 still clears the 60% bar — **which is
exactly why it went unnoticed for two weeks.** Maintenance Practice 4 forbids
hand-typing a benchmark number; a number read out of a stale artifact is that
defect one level up.

`tests/test_route_recall_artifacts_are_fresh.py` re-runs both harnesses (~1s
each) and fails when a committed artifact disagrees. **Non-vacuity is against the
real defect**: restoring the actual stale `results.json` from `HEAD` turns it red.

⚠ A red there on a PR that only touched tool descriptions is the signal working.
Descriptions are the router's input, and the moratorium conditions are stated
over these exact numbers.

## 3. The open question, restated as one binary

Strip the degeneracy away and the emitted corpus is a single decision:

- **87.5%** of golds are `search_text` or `search_symbols`
- split **18 / 17**
- majority-class baseline **51.4%**
- route, on cases where it lands in the pair at all: **12/23 = 52.2%** — chance

One rule, `/find|locate|where is|look up|search for|definition of/ →
search_symbols`, takes 21 of 35 rank-1 picks because agent-emitted tasks open
with "find".

**So the objective is not `route@1` versus `route@3` over 91 actions. It is
`P(correct | gold ∈ {search_text, search_symbols})`.** Anyone proposing to
optimise route should say which of those two numbers they intend to move.

⚠ This also explains why the identifier-shape and verb hypotheses died. Both
failed on **coverage** at 5–15%, but the decision needing an answer is one binary
required **100%** of the time. A predicate reaching 15% cannot move it — purity
was never the issue.

**H3 is named in `ROADMAP.md` and deliberately not started:** `search_symbols`
matches names, `search_text` matches content, so the discriminator is whether the
sought thing *is* an indexed symbol name — absent from the query string
(consistent with both refutations), cheap from the index, **coverage 100% by
construction**. It needs the 157 unused corpus rows sampled for *pair balance*
rather than uniformly, and the predicate declared before labels are read.

## Scope

Benchmarks, tests and docs only. No source change, no version bump, **and the
moratorium does not move** — this makes the disclosure stronger, not weaker.
Suite **8109 passed, 17 skipped** (8105 + exactly the 4 new tests), `ruff check`
clean over `src/` and `benchmarks/route_recall/`.


jgravelle:
**Amending the H3 claim in the description above.** It said H3 "needs the 157 unused corpus rows sampled for pair balance". That instruction is wrong, and I checked before starting rather than after.

**H3 is not runnable on this corpus, and the blocker kills the whole outside-the-string hypothesis class.**

The rows are `{case_id, candidate_rank, prompt_text, gold_primary, gold_alts, emitted_task}` — **no repository field** — and the tasks are heterogeneous by design: "this project", "our mod", "the capture button handler". Of the 35 pair-labelled cases, **4 name a resolvable repo.** There is nothing to index, so there is nothing to probe; pointing one shared index at all 35 would score accidental matches rather than the property.

That generalises past H3. The joint H1/H2 finding points at a signal from *outside* the query string — repo state, a first-pass retrieval, prior turns — and every one of those needs the context this corpus does not carry. H4 and H5 of the same shape hit the identical wall, after paying the setup cost.

⚠ **The 157 unused rows do not fix it.** Same generator, repo-less for the same reason. They are an asset only for query-*string* hypotheses — the class already refuted twice. The standing note read as "the next hypothesis is ready, 157 rows are waiting"; it was not, and that reading is now corrected in `ROADMAP.md` and `CLAUDE.md`.

**What readiness costs**, priced so the decision gets made rather than discovered: a corpus where each case is bound to a real repository at a pinned commit, tasks generated against that repository, and gold labels assigned by someone who can see it. That is corpus construction, not an afternoon.

Until it exists, `P(correct | gold ∈ {search_text, search_symbols})` is a well-posed objective with **no instrument that can measure a repo-grounded answer to it.**

Pushed as `c20692c`. Docs only on top of the k-set floor work — no source change, no version bump, moratorium unmoved.

PR #533, opened 2026-08-21
The grounded pilot: H3 is refuted, and the corpus project is cancelled

jgravelle:
The small grounded pilot, built to decide whether the full repo-grounded corpus
was worth starting. **It decided: no.**

Three commits, in this order, and the order is the point — `git log` for
`benchmarks/route_binary_pilot/` shows the predicate and protocol landing before
a single case existed. An assertion of pre-registration in prose is not evidence;
a commit sequence is.

## Result

60 cases, balanced 30/30, bound to `expressjs/express`, `fastapi/fastapi` and
`gin-gonic/gin` at the SHAs published in `benchmarks/tasks.json`.

| | accuracy | vs 50% floor | Wilson 95% | p vs chance |
|---|---|---|---|---|
| full vocabulary | **53.3%** | +3.3 | [40.9, 65.4] | **0.699** |
| own name ablated | **50.0%** | 0.0 | [37.7, 62.3] | **1.000** |

Indistinguishable from a coin, with the interval spanning the floor in both
conditions. Leakage existed — 12 of 30 class-S tasks matched their own target's
name — and bought nothing.

## The mechanism is the part worth keeping

| | gold `search_symbols` | gold `search_text` |
|---|---|---|
| accuracy | **100%** | **6.7%** |

**The predicate answered `search_symbols` on 58 of 60 tasks.** It scores 100% on
one class for exactly the reason it scores 6.7% on the other: it never says
anything else. A constant classifier wearing a probe's clothing.

Why — in a real repository the symbol vocabulary absorbs ordinary English:

| repo | symbols | matchable name parts |
|---|---|---|
| expressjs/express | 200 | 164 |
| gin-gonic/gin | 1,179 | 1,347 |
| fastapi/fastapi | 6,841 | **4,303** |

Fourteen of sixteen common English words tested are fastapi symbol-name parts:
`message`, `from`, `path`, `string`, `file`, `text`, `status`, `value`, `error`,
`name`, `body`, `type`, `data`, `request`.

## ⚠⚠ Coverage was never the property, and H3 was argued for on that ground

| hypothesis | fires on | fails because |
|---|---|---|
| H1 identifier shape | ~15% | decides too few cases |
| H2 imperative verb | ~5% | decides too few cases |
| H3 vocabulary probe | **~97%** | decides them all the same way |

I argued H3 was the first of its family worth running *because* its coverage was
100% by construction. That was **necessary and not sufficient** — the same shape
as moratorium conditions 1 and 2 being met without clearing the freeze. The
property all three lack is **separation**: a predicate must fire *differently* on
the two classes, and firing often is not firing differently.

**Screen the next hypothesis on separation before anyone counts its coverage.**

## The decision the pilot was built to make

`PROTOCOL.md` registered the asymmetry in advance: a negative is decisive, a
positive would have certified nothing. **The corpus project is cancelled.**
Building cases bound to real repositories with labels assigned by someone who can
see them would have cost a project and hit the same wall — because the wall is
not the corpus. It is that vocabulary membership does not separate these two
classes in any repository large enough to matter.

Cost of finding that out: an afternoon and three clones.

## Not ruled out, and deliberately not run

A probe keyed on retrieval **outcome** rather than vocabulary **membership**:
does `search_symbols` actually outrank `search_text` for this query against this
index? That compares two scores instead of testing set membership, and index size
does not trivially defeat it.

⚠ **These 60 cases are spent.** Running H4 against them would be a fitting pass
wearing an experiment's clothes. New hypothesis, new cases, same registration
discipline.

## Disclosure

One mid-build correction: the first generated corpus filled class T with
`it('should …')` test titles and would have been a test-title detector. Test
paths are excluded now. **The fix landed after inspecting the cases and before
the predicate was run once** — no prediction was computed against the discarded
version, and `predicate.py` is untouched since its registration commit.

## Scope

Benchmarks and docs only. No source change, no version bump, **and the moratorium
does not move.** Suite **8109 passed, 17 skipped**, `ruff check` clean over `src/`
and `benchmarks/`.


jgravelle:
**Answering "do we merge this if H3 is refuted?" — yes, and here is what it needed first.**

The negative result *is* the deliverable. This project already treats them that way: H1's and H2's refutations are cited constantly, and the codex_surface negative is kept as a standing warning. A refutation living only in a closed PR is the "where a fact lives decides whether it is used" failure — the next session proposes H3 again.

But merging it unguarded would have reintroduced two defects fixed elsewhere today.

**1. A spent corpus is an attractive nuisance.** `RESULT.md` says "must not be run on this corpus" — that is prose, and prose warnings get walked around. `predicate.py` is now pinned by digest: editing it while `cases.json` exists is a fitting pass, and in a diff it looks like ordinary work. ⚠⚠ The test **cannot judge whether a given edit is fitting and does not try — it makes the judgement happen**, the same design as the LICENSE digest pin. Hashed over line-ending-normalised text, because hashing raw bytes pins a property of the *checkout*; that exact mistake went red on every Ubuntu leg and green on every Windows one when the LICENSE pin shipped.

**2. Frozen artifacts go stale.** That is this morning's defect — `benchmarks/route_recall/results.json` drifted while `CLAUDE.md` cited its stale figure for two weeks. The pilot cannot have a freshness gate: re-running it needs three external checkouts and three local indexes, none of which exist in CI. So the guard runs the other way — every number quotable from `RESULT.md` must match `results.json`, the verdict must still be stated, and **the at-chance conclusion must still follow from the artifact**. That last is asserted as the *property* (interval spans the floor, p not significant) rather than the literals, so a genuinely separating re-run fails it and forces a re-reading instead of quietly inheriting the old conclusion.

⚠ **The gate found two mismatches on its first run**: `RESULT.md` rounded `0.6989` to `0.699` and never used the word "cancelled". The prose was corrected, not the gate.

Non-vacuity, per falsification:

| falsification | red |
|---|---|
| edit `predicate.py` | digest test |
| soften REFUTED → "inconclusive" | refutation test |
| artifact claims significance | number-match **and** at-chance tests |

Pushed as `54e95d3`. Suite **8114 passed, 17 skipped** — 8109 plus exactly these 5 — ruff clean over `src/`, `tests/` and `benchmarks/`.

jgravelle:
**Status note so eleven green checks are not read as done.**

This PR is `MERGEABLE/BLOCKED`: 11 checks green, 0 failing, and `license/cla` **absent** rather than failing.

Diagnosed rather than assumed, because two causes present identically:

- `commits/<head>/status` returns `state=pending count=0` on **both** heads of this branch — the original `31720d7` and the current `54e95d3`.
- There is **no CLAassistant comment on this PR at all** (only the health-radar bot).

Zero statuses on the *first* head plus no bot comment is the "never fired" shape, not the "our push erased it" shape. The bot is working generally — it posted `license/cla = SUCCESS` on #532 earlier today — so this is specific to this PR, and two pushes have not woken it.

⚠ **Absent counts as NOT SIGNED and is never a thing to merge past.** Now that `license/cla` is a required check on `main`, that correctly reads as `BLOCKED` rather than as a missing row — which is the fail-closed behaviour policy 3d was enabled for, showing up for the first time.

The remedy is the bot's own, and it is one click: https://cla-assistant.io/check/jgravelle/jcodemunch-mcp?pullRequest=533

Leaving it blocked until then. Nothing in the branch is waiting on anything else — suite 8114/17/0, ruff clean, and the moratorium does not move either way.

Issue #421, opened 2026-08-07
`receipt` scans only `~/.claude/projects`, so sessions under a custom `CLAUDE_CONFIG_DIR` profile are never counted (measured: 12 of 348 calls seen)


MotoMato85:
## Environment

- jcodemunch-mcp 1.108.251 (latest on PyPI at the time of writing), Windows 11 Pro 10.0.26200, Python 3.14
- Claude Code with two additional profiles selected via the `CLAUDE_CONFIG_DIR` environment variable, so transcripts land in three places: `~/.claude/projects` (default), `~/.claude-headroom-code-on/projects`, and `~/.claude-headroom-code-off/projects`
- Index store at the default `~/.code-index` (no `CODE_INDEX_PATH` set)
- Also affects jMunch Console's Savings panel, which shells out to `jcodemunch-mcp receipt` for its "jCode tool calls" tile and per-tool table

## Bug

`jcodemunch-mcp receipt` reports far fewer jCode calls than were actually made. Claude Code writes each session's transcript under the active profile's config directory (`CLAUDE_CONFIG_DIR`, default `~/.claude`), but `receipt` scans a hardcoded `~/.claude/projects`, so every session run under any other profile is never counted. The only override, `--projects-root`, takes a single path, so no invocation can cover two profiles at once, and a machine with two profiles open simultaneously can never get one correct total.

Measured on this machine for the August 1 to 7 window, same `receipt` binary, only the root changed:

| transcript root | calls | savings_tokens |
| --- | ---: | ---: |
| `~/.claude/projects` (what `receipt` scans) | 12 | 32,646 |
| `~/.claude-headroom-code-on/projects` | 223 | 745,339 |
| `~/.claude-headroom-code-off/projects` | 113 | 349,280 |

So `receipt` reports 12 of the 348 calls actually on disk, about 3%. jMunch Console's "jCode tool calls (This Month)" tile shows that same 12 no matter how much jCodeMunch is used, because the console invokes `receipt` without `--projects-root` (its per-tool breakdown comes from the same scan and misses the same calls). The meter-driven tokens tiles keep updating correctly, which makes the frozen call count look like a console bug. It is not. The console faithfully renders what `receipt` returns.

## Isolation

[...]

Issue #406, opened 2026-08-03
check_references misses call sites in the defining file, and the TaskCompleted hook reads a key it never returns


MotoMato85:
Severity: High: `check_references` — the documented "is this identifier used anywhere" / dead-code check — returns `is_referenced: false` for symbols that are demonstrably called, whenever the call sites live in the same file as the definition.

## Summary

Two independent defects in the same chain.

**(1) `_check_single` skips the whole defining file.** The content scan builds `defining_files` from every file containing a symbol with the matching name, then `continue`s past those files entirely. The intent — "finding the name in the defining file is not a reference" — is right, but the implementation discards the file rather than the definition. Every same-file call site is lost. Single-file modules are hit hardest: a helper defined and used in one module reports as unreferenced.

**(2) `run_taskcomplete` reads `total_references`, which `check_references` never returns.** `ref_result.get("total_references", 0) == 0` is therefore always true, so the TaskCompleted diagnostic labels every candidate symbol it inspects as unreferenced.

(2) is the same defect class as #338 — a caller reading a shape the callee does not produce. That fix (commit `7d0e996`, 2026-06-19) moved `check_edit_safe` / `check_delete_safe` onto the batch form. `cli/hooks.py` was last modified 2026-06-17 and was not part of that sweep. Neither singular nor batch mode exposes a `total_references` key.

## Environment

- `jcodemunch-mcp 1.108.224` (latest release; both sites verified unchanged on `main`)
- Windows 11, Python 3.13, `uv tool install`
- Reproduced through MCP tool calls and directly against the library

## Reproduction

Minimal — one file, six lines:

```python
# mod.py
def helper(x):
    return x + 1


def main():
    return helper(1)
```

```text
index_folder(path="<dir containing mod.py>")
check_references(repo="<repo>", identifier="helper", search_content=true)
search_text(repo="<repo>", query="helper")
```

## Observed vs expected

Observed:

```text
check_references('helper') -> is_referenced: false, import_count: 0, content_count: 0
search_text('helper')      -> 2 matches
                              mod.py L1: def helper(x):
                              mod.py L6:     return helper(1)
```

Expected: `is_referenced: true`, with `mod.py` L6 reported as a content reference. L1 is the definition and should stay excluded.

Second case, from a real 1837-symbol index — `validateCatchRecord`, defined at `simulator/lib/sync-format.js:201-239`:

```text
ground truth (all occurrences, all in one file):
[...]

Issue #334, opened 2026-06-17
PreCompact CLI hook cannot see the live MCP session journal


mmashwani:
Severity: Medium (long-running Claude Code and Codex sessions can lose the intended code-session recovery context at compaction time because the registered PreCompact hook runs as an external CLI process and emits an empty session snapshot)

## Summary

The `hook-precompact` CLI path is registered for Claude Code and Codex as the mechanism that should inject a compact jCodeMunch session snapshot before context compaction.

In a live MCP session, `get_session_snapshot` correctly reports the files and searches already touched by the agent. But the external CLI hook runs in a separate process, imports `get_session_snapshot()`, and reads a fresh process-local `SessionJournal`. The result is a valid `systemMessage`, but it usually contains `Files explored: 0 | Searches: 0`.

That makes the hook look healthy from a command/exit-code perspective while failing the thing it is meant to preserve.

## Environment

- `jcodemunch-mcp`: 1.108.55
- `jdocmunch-mcp`: 1.84.0
- Clients considered: Claude Code and Codex
- Hook shape: external command registered as `jcodemunch-mcp hook-precompact`

## Reproduction

1. Start a client session with jCodeMunch connected.
2. Use jCodeMunch tools enough to create live journal state, for example read several symbols/files and run a few searches.
3. Call the live MCP tool:

   ```text
   get_session_snapshot
   ```

4. In a terminal, run the same external command a client hook runs:

   ```bash
   printf '{}' | jcodemunch-mcp hook-precompact
   ```

5. Compare the live MCP snapshot with the external CLI hook output.

I also created a small probe script that demonstrates the process boundary directly:

- External `jcodemunch-mcp hook-precompact`: empty snapshot.
- In-process `run_precompact()` after seeding `SessionJournal`: non-empty snapshot.

## Observed vs expected

Observed:

```json
{
  "systemMessage": "## Session Snapshot (jCodemunch)\n**Duration:** 0s | **Files explored:** 0 | **Searches:** 0"
}
```

In the same investigation, the live MCP `get_session_snapshot` reported several focus files and recent searches.

Expected:

The external PreCompact hook should emit the same useful session state the live MCP server sees, or clearly report that no live session state was available and fall back to a different compact orientation payload.

## Root cause (verified in source, v1.108.55)

The CLI hook builds the snapshot inside the short-lived hook process:

[...]

PR #535, opened 2026-08-24
fix(hooks): ten confirmed defects in the Claude Code hook steering layer

marcelruhf:
## Symptom

Claude Code sessions "forget" to call jCodeMunch and fall back to local search. `Bash(grep/rg/find)`, `Glob`, raw `Read`. That is exactly what the enforcement hooks exist to stop. The hook layer was broken in more than one place, and the breaks did not share a cause. Every fix here reproduced its defect against the unfixed tree first.

### Features that were silently dead

**TaskCompleted diagnostics never ran in any real deployment.** The hook runs in its own process, so the in-process journal it read was always empty and the handler returned without output. #334 fixed this class for PreCompact. This handler never got the bridge. It now reads the persisted live journal, bounded at 240 minutes so the report cannot present a dead session's edits as this task's, and uses the batch `check_references` form from the #406 sweep.

**Three loops read `owner`/`name` keys that `list_repos()` never returns.** Entries carry `repo`. The SubagentStart briefing, the PreCompact/SessionStart landmarks, and TaskCompleted repo scoping skipped every repo, always. Test mocks with the wrong entry shape hid it. The mocks now use the real `{"repo": "owner/name"}` shape. After that loop actually ran, landmarks still skipped later repos: one `seen` set held both repo ids and symbol ids, so the first repo with changed symbols made the next repo look already visited. Repo ids and symbol ids now have their own sets.

**The PostToolUse reindex spawned a bare-name `jcodemunch-mcp`.** Under the hook shell's minimal PATH, that name resolves nowhere on pipx and `pip --user` installs. Those are the installs `init` already resolves with an absolute path. Spawns now reuse this process's own invocation and fall back to `python -m`. The spawn refuses a relative `argv[0]`, because the hook's cwd is the checked-out repo and a file named `jcodemunch-mcp` there must never be what runs. Copilot's hooks.json gets absolute paths. `install`, `uninstall`, and `status` match rules by subcommand, and a re-install upgrades a leftover bare-name Copilot rule in place.

**A symlink in the checkout path disabled steering for that repo.** `index_folder` stores roots resolved. The hook compared with `abspath` only, so on macOS the `/tmp` to `/private/tmp` link alone silenced both the nudge and the strict deny. All root comparisons now go through `realpath`.

**Re-running `init` never upgraded a stale rule.** A pre-1.108.47 install kept matcher `"Read"` forever while install reported success. `config --check` printed the expected matcher instead of the installed one, hiding the drift. The merge now converges matcher and command to the shipped rule, and leaves the matcher alone on rules that carry a user-authored hook. The check prints the installed matcher and warns on drift.

**PreCompact wrote its snapshot into a field Claude Code discards.** The hooks docs say Claude Code drops a PreCompact hook's `systemMessage` and `continue` fields. `init` only wires this hook into Claude Code, so the emission reached nobody in any supported install, and still paid a full snapshot build on every compaction. Since v1.108.255 (#420), `SessionStart(compact)` injects the snapshot as `additionalContext`, which reaches the model. This PR removes the PreCompact emission. `systemMessage` is a user-display field wherever a harness honors it. The snapshot exists to re-orient the model after compaction. Porting that field would still show a human a report and miss the model. The hook stays registered. `run_precompact` is one line of transcript-root registration (#421). PreCompact is the only moment a hook gets before compaction, and it costs one spawn.

### Coverage that steered the wrong way

**Bash and Glob never fired the hook.** The matcher was `Read|Grep`, so the usual native search routes never hit it. Strict mode then funneled a denied Grep into unmonitored Bash. The matcher is now `Read|Grep|Glob|Bash`. The hook intercepts only command lines that open with a search command, after stripping leading `FOO=1` assignments and one subshell paren. A grep after a pipe filters output jcm cannot serve, so it passes untouched. Strict mode denies pure searches only. Never `find`, because `find … -delete` opens with the same word. Never pipelines or compound commands. Never commands that name paths outside every indexed root.

**The advisory Read nudge fired even outside indexed repos**, and recommended tools that would error there. It now checks indexed-root overlap first, as strict mode always did.

**The SubagentStart catalog ignored `tool_surface`.** A first-ever install writes `tool_surface="counter"`. Under that setting the briefing named 41 tools the subagent's client cannot call, and never mentioned `order`/`menu`/`route`. The briefing now matches the tools the client can actually call, and loads only the repo or repos that contain the subagent's `cwd`. Brief-everything stays the fallback.

**Hostile input used to crash the hook.** A garbage `JCODEMUNCH_HOOK_MIN_SIZE` no longer crashes every hook at import. It parses to the default. Non-dict stdin, null `tool_input`, NUL-byte paths, list-shaped Copilot `toolArgs`, and malformed live-journal entries no longer raise. Landmarks now refuse a bad journal row the same way TaskCompleted already did.

## Split and cost

`cli/hooks.py` was 978 lines on main. These fixes pushed it past 1,100, so it is now the `cli/hooks/` package: `_common`, `steering`, `reindex`, `snapshot`, `landmarks`, `taskcomplete`, `briefing`. The facade re-exports the `run_*` entry points, so `server.py` dispatch is unchanged. The facade imports in ~5 ms warm on the machine that measured it.

The steering gate reads roots via `IndexStore.list_source_roots()`, one read-only meta row per `.db`. That replaces `list_repos()`, whose per-repo `SELECT COUNT(*)` scans used to run in front of every intercepted call.

The TaskCompleted and landmark loops probe file membership read-only with `SELECT 1 … LIMIT 1` before hydrating an index. Any doubt hydrates.

Transcript-root registration takes an exact-string fast path in the steady state instead of realpath-resolving up to 64 stored roots per hook event.

Tool-surface resolution is one function, `counter.resolve_tool_surface`, a pure module shared by the server and the hooks. Validation used to live only in the server.

## Behavior that changed on purpose

Matching `Bash` puts a hook process, ~0.4 s warm on the machine that measured it, in front of every Bash call in every mode. The mode check happens inside the hook. A fast hook entry that skips the server import is the recorded follow-up. That is what drops the 0.4 s.

Features that used to be silent now print something. Briefing content, landmarks, post-task diagnostics.

`install` and `init` now report "converged … to the shipped rule". `config --check` prints the installed matcher and a drift warning.

A consumer that rendered PreCompact's `systemMessage` loses that report. Of the other harnesses this project configures, two would show the field: Codex in its TUI, and Gemini CLI as an INFO line after `gemini hooks migrate --from-claude`. Neither reads the hooks `init` installs. Both look in their own config files. Cursor is the one harness that auto-loads `~/.claude/settings.json`, PreCompact included, and it drops `systemMessage` from hook output. Where the field renders at all, it renders to the user, not into model context. SessionStart delivers the snapshot now.

The reindex child is now this install's own binary. A setup that shadowed PATH to redirect the reindex loses that redirection.

## Known limits

Two concurrent sessions sharing one `CODE_INDEX_PATH` can attribute each other's edits in the advisory diagnostics. The live journal has no session binding, and the MCP server never learns the client session id, so fixing that is a protocol change, not a hook change. A `git worktree` checkout is a different directory and stays outside indexed roots. The realpath fix covers symlinked paths only. The ~0.4 s stays until the fast hook entry lands.

## Verification

Ubuntu CI on this head is 8222 passed, 24 skipped, 0 failed, same totals on 3.10, 3.11, 3.12, and 3.13. Windows is 8229 passed, 17 skipped on 3.10 through 3.13. Both collect 8246 tests. A local 8112 / 134 run collected the same 8246 with optional packages missing. `ruff check src/` is clean.

`tests/test_hook_steering_fixes.py` pins each defect as an outcome. One test is end-to-end with no mocks: a real index, a symlink-alias cwd, real store reads. Another fails if diagnostics silently drop one of the session's symbols. The roots listing and the membership probe have real-store tests. One more fails if any briefed catalog name is missing from `server._catalog_names()`, so the briefing list cannot rot the way the briefing that ignored `tool_surface` did. The Copilot install test now asserts by subcommand, not a bare-name substring, because the absolute-path command on Windows is `jcodemunch-mcp.EXE` and the old assert could never match it. Some older tests had encoded the defects: wrong mock shapes, assertions on the discarded channel, byte-preserving dedup. Those tests now assert outcomes.


jgravelle (approved):
Reviewed the whole diff. Approving.

The defect class this closes is the one that is hardest to notice: features that ran, returned nothing, and reported success. The `owner`/`name` versus `repo` key mismatch across three loops is the clearest case, and the detail that the only producer of the wrong shape was a test mock is the part worth keeping. So is declining to add a fallback for it.

A few things I checked rather than took on faith:

- The extension set is byte-identical to the one on main, so the wider matcher does not also widen what counts as a code file.
- `_path_overlaps` covers the ancestor direction, so grepping a parent of an indexed repo still intercepts.
- `_repo_contains_any` fails open on every error path. A wrong False would silently kill the caller; a wrong True costs one load.
- The `os.path.isabs` guard in `_self_invocation`. The hook cwd is the checked-out repo, so this is the difference between re-invoking ourselves and executing whatever a repo happens to name `jcodemunch-mcp`.
- `resolve_tool_surface` changes one behaviour beyond consolidation: a whitespace-only env value now falls through to config instead of resolving to unrecognized. That is the better answer and the docstring says so.

## The one number I would like to correct

The PR reports the Bash matcher costing ~0.4 s warm. On Windows 11, Python 3.10, this repo, it is **4.0 s**, measured three times, with the same result for a command the matcher does not intercept:

```
non-search Bash: 4.095s / 4.053s
search Bash:     4.154s
```

The split is the useful part:

```
import jcodemunch_mcp.server        0.744s
main() before dispatch              3.018s
python -c pass                      0.119s
```

So roughly three quarters of it is `main()` building an argparse tree of 45 subparsers that a hook invocation never reads. That is ours and it predates this PR. It was simply free until now, because the matcher excluded Bash and the hook never spawned for one.

That does not change the merge decision. It does change what the follow-up is: not a fast hook entry that skips the server import, which is worth 0.74 s, but an argv branch ahead of argparse, which is worth the other 3.0 s. The hooks package imports in 0.200 s standalone, the same as the single module did before the split, so the ceiling here is roughly 20x rather than 5x. I am implementing it now and it will ship in the same release as this.

Thanks for reproducing each defect against the unfixed tree first, and for saying plainly which older tests had encoded the defects. That is the part that usually goes unsaid.

jgravelle:
Approved above. The one thing standing between this and `main` is that the PR is still a draft, so the merge is refused at the API.

Mark it ready for review by **2026-08-25 11:00 UTC** and it merges as your commits.

If that time passes, we implement the same change ourselves and you are credited in the CHANGELOG, the release notes and the close comment. The window decides whose commits these are. It does not decide whether the work ships, and it does not decide credit.

Separately, and not a condition of the merge: the argparse cost I measured is ours to fix, and I am doing that now regardless of which way this goes.

jgravelle:
Correction to my review, because it changes what the follow-up is.

I attributed the startup cost to the argparse tree. That was wrong. The 45 subparsers cost **7 ms**. The cost is `verify_package_integrity()`, which runs on every CLI invocation and used `packages_distributions()` — a map of every distribution on `sys.path`, built to answer a question about one of them. 3.35 s here, uncached, per invocation, and on this box it then returned nothing because the code runs from source.

```
packages_distributions()         3.350s
distribution("jcodemunch-mcp")   0.005s
main() argparse tree             0.007s
import jcodemunch_mcp.server     0.744s
```

Fixed in #537: cheapest-first ordering, the full map reached only when the official distribution is absent or did not provide the running module. `hook-pretooluse` end-to-end goes 4.0 s to 0.94 s on this box.

So your recorded follow-up is still the right one and it is now the whole remaining term — the ~0.74 s server import, reached through the `jcodemunch_mcp.server:main` console entry point. I have not touched that; it is yours if you want it, and the timebox above is unaffected either way.

Worth noting the split you did costs nothing measurable: `cli/hooks/` as a package imports in 0.200 s, the same as the single module did.

jgravelle:
Merged. Two housekeeping notes, neither of which changes anything about the work.

**I marked this ready for review on your behalf.** It had been a draft since it was opened, with no WIP marker in the title or body and no draft toggle in its history, so it read as a flag that never got flipped rather than a hold you had put on it. If you were not finished, say so and we will treat what is on `main` as a checkpoint rather than a verdict — nothing here is hard to revise.

**Your CLA status was missing for about an hour and it was not your doing.** When you merged `main` at 11:49, the new head arrived with no `license/cla` status, and with that check required the PR read `BLOCKED`. Nothing was erased and your signature was never in question: `license/cla` is a legacy commit status rather than a check run, so it is posted to one specific SHA and a new commit simply starts with none. The old head still reads `success`. The webhook received your push and answered `200 OK` without posting, so we redelivered that same event from our side and the status appeared in about 25 seconds. You did not need to do anything, and if it happens again on another PR, it is ours to clear.

On the substance: the merge conflict resolution was clean. `CHANGELOG.md` came through `+40/-0`, meaning you kept both `[Unreleased]` blocks instead of overwriting one, which is the usual casualty. The full matrix then re-ran green against current `main` rather than against the base you forked from, so this is verified where it counts.

The startup cost we discussed is fixed in #537, already on `main` and included in your merge. It was not argparse. `verify_package_integrity()` called `packages_distributions()` on every invocation, which maps every distribution on `sys.path` to answer a question about one of them — 3.35 s here, and on this machine it then returned nothing. Cheapest-first ordering takes a `hook-pretooluse` spawn from 4.0 s to 0.94 s.

That leaves the ~0.74 s server import as the entire remaining cost of the Bash matcher, reached through the `jcodemunch_mcp.server:main` console entry point. Still yours if you want it, and no clock on it.

Thanks for this one. Ten defects, each reproduced against the unfixed tree first, and you said plainly which older tests had been encoding the defects rather than quietly rewriting them. That last part is the one that usually goes unmentioned.

jgravelle:
Shipped in [v1.108.294](https://github.com/jgravelle/jcodemunch-mcp/releases/tag/v1.108.294), live on PyPI.

Your work leads the release. The CHANGELOG section is credited to you and states that the whole of it is yours including the write-up, and the release notes open on the ten defects rather than on the performance fix that came out of reviewing them. I reordered the sections so the generated summary would carry your name rather than mine.

Verification, for the record: local 8234 passed / 17 skipped / 0 failed, with the 3.13 reproduce identical. Totals reconcile against 1.108.293 — 8251 against 8180, which is 5 from the integrity-check tests and 66 from yours. Your CI reported 8246 collected, and 8251 minus those 5 is 8246 exactly, so two independent measurements agree.

The ~0.74 s server import is untouched and still yours, with no clock on it. Nothing about this release changes that.

PR #547, opened 2026-08-25
docs: fix both broken pointers in CLAUDE.md

jgravelle:
`docs/` is deliberately local — the `.gitignore` comment reads "Local docs notes", and history shows `b4d2aa8 chore: remove internal planning docs from feature branch`. That convention is left intact.

This one spec is different from its siblings. The `PARSER_GENERATION` comment (#546) cites it as **the thing that ends the manual-bump obligation**:

> Until that exists, ANY parser change that alters which symbols exist must bump this line in the same commit.

A contributor bound by that rule has to be able to read the document that says when it lifts. So exactly one file is tracked; everything else in `docs/` stays local.

## The gotcha worth recording

The pattern is `docs/*`, **not** `docs/`. Git cannot re-include a file whose parent *directory* is excluded, so `!docs/prd-extraction-fingerprint.md` under the directory form is **silently inert** — the first attempt here read as correct and changed nothing. `git check-ignore -v` was also misleading: it printed the negation line while still exiting 0.

Verified the way that can't lie:

```
$ git ls-files --others --exclude-standard docs/
docs/prd-extraction-fingerprint.md
```

One file, and `prd-parity-map.md` / `prd-evidence-receipts.md` confirmed still ignored.

## Still broken, not fixed here

`CLAUDE.md` also references `docs/prd-framework-routes-endpoint-impact.md`, which remains untracked and is therefore an unfollowable pointer for anyone but the maintainer. Left alone deliberately — no rule is bound to it, so the same argument does not apply, and widening the exception on my own judgment is how a narrow carve-out becomes a general one.


jgravelle:
Second commit added: the other broken pointer.

`docs/prd-framework-routes-endpoint-impact.md` is worse than untracked — it is **not on disk and never existed in git history**. CLAUDE.md was citing a document that has never existed anywhere. The surrounding claim is real and kept (get_endpoint_impact is the first slice of the framework-routes design, FastAPI prefix / Spring class-mapping is the follow-on); only the pointer goes, with a note saying why so the next reader does not go looking.

Found by sweeping **every** file path in the tracked docs rather than fixing the one that was reported — the standing lesson that the reported list is usually not the list. Two other results, both deliberately left alone:

- `benchmarks/token_baselines/vX.json` — a template placeholder, not a broken path. The directory exists and holds `v1.108.163.json`.
- CHANGELOG hits (`src/main.ts`, `src/riskGutter.ts`, `docs/legacy/.../File123.php` …) — historical entries citing example paths from *other* projects. CHANGELOG is authoritative history, not a live pointer surface, and rewriting it to chase paths would be the wrong repair.

⚠ Note for anyone tempted to add a ratchet here: the replacement text still *names* the file, in order to explain that it does not exist. A naive scanner flags it identically to the citation it replaced, so such a guard would have to distinguish 'cited as a pointer' from 'named as absent' — otherwise it is a guard with a false positive on the very fix it is meant to protect.

Issue #414, opened 2026-08-05
Symbol names corrupted in 16 language extractors - tree-sitter byte offsets used to slice a decoded `str` (any non-ASCII character earlier in the file shifts every following symbol name)


MotoMato85:
## Environment

- OS: Windows 11 (10.0.26200.8973)
- jcodemunch-mcp: 1.108.241, installed via `uvx jcodemunch-mcp`. Every line number below refers to `jcodemunch_mcp/parser/extractor.py` as published in `jcodemunch_mcp-1.108.241-py3-none-any.whl` on PyPI: 431252 bytes, md5 `72e83ea63c50fbab91a79fbf0f049495`. The file analysed here was compared byte for byte against a fresh download of that wheel.
- Python: 3.14.6
- tree-sitter: 0.26.0
- tree-sitter-language-pack: 0.13.0
- mcp (Python SDK): 1.29.0
- Client: Claude Code, stdio transport
- Backend: sqlite

The `mcp` SDK version is incidental: the repro at the bottom imports `parse_file` directly and never starts a server. The same failure was first seen under mcp 1.23.3 and still reproduces unchanged under 1.29.0.

MCP server config used:

```json
{ "mcpServers": { "jcodemunch": { "command": "uvx", "args": ["jcodemunch-mcp"] } } }
```

Note: this is not transport-, client-, or OS-specific. It is a pure parsing bug, reproducible in a plain Python script with no MCP server involved (see Repro steps).

## Bug

16 of the per-language symbol extractors in `jcodemunch_mcp/parser/extractor.py` slice a **decoded `str`** using tree-sitter's **byte** offsets. The shape below appears verbatim as a `_text` helper in 9 of the 16; the other 7 perform the same slice inline or through a differently named helper:

```python
source = source_bytes.decode("utf-8", errors="replace")

def _text(node) -> str:
    return source[node.start_byte:node.end_byte]
```

`node.start_byte` / `node.end_byte` are byte offsets into the UTF-8 source. Indexing a `str` with them is correct only while the file is pure ASCII. Each non-ASCII character appearing **earlier** in the file pushes the extraction window forward by the number of extra bytes it encodes to: one for U+0080..U+07FF, two for U+0800..U+FFFF, three above that. The window therefore lands on an unrelated run of source text further down the file. Its width stays `end_byte - start_byte`, so for an ASCII identifier it still spans exactly the right number of characters. That is why the corrupted names read as plausible fragments rather than obvious garbage.

[...]

PR #548, opened 2026-08-25
feat(parser): Racket language support, with its ceiling measured not asserted

otherjoel:
Adds Racket (`.rkt` / `.rktl` / `.rktd`) as a first-class language, and ships a harness that measures how far the static parser actually gets rather than asserting it.

## Why a custom walker

The tree-sitter Racket grammar is **fully homoiconic** — there are no named `define` or `struct` nodes; every form is `list` → `symbol`, and `(…)` and `[…]` share the node type. So `_parse_with_spec` is unusable and this follows
the three Lisps already here (`_parse_clojure_symbols`, `_parse_commonlisp_symbols`, `_parse_elisp_symbols`), dispatching on the head symbol with an empty `LanguageSpec`.

Extracts `define` (procedure vs value discriminated by header shape or a lambda-valued RHS), the `define/*` method forms, `define-syntax` / `define-syntax-rule`, the `struct` family, `define-type` / `define-signature` / `define-generics`, `define-values`, and `module` / `module+` / `module*` plus `(define C (class …))` as nested scopes with `::`-qualified names. It is also the first Lisp in the tree with **import edges** (`require`, including `only-in` / `rename-in` / `prefix-in` / `for-syntax`) and **call references**. `;;` and `#| |#` blocks above a form become the docstring.

## The measurement, and why it is the point of the PR

`benchmarks/racket_fidelity/` compares the extractor against Racket's own expander. `syntax-original?` separates names a human typed from names a macro introduced, so only the first group counts toward coverage — counting macro output as a miss would make the gap look several times worse than it is. `module->exports` answers the separate consumer-facing question and reports `procedure?` on the instantiated value.

Over 211 files of the collects tree (Racket 9.2):

| Bucket | | |
|---|---|---|
| `extra` | **0** | nothing we emit is a name Racket does not know |
| `wrong_span` | **0** | every definition falls inside the bytes we return |
| `missing` | 485 | **86.2%** of human-written module bindings found; **152 of 211 files completely clean** |
| `callable_unknowable` | 212 | a ceiling, not a bar |

The asymmetry is deliberate: an incomplete index makes an agent read the file, a wrong one makes it repeat a falsehood.

## Limitations

- **A macro that DEFINES is invisible.** `racket/list.rkt` writes `(define-lgetter second 2)` twelve times; `racket/fasl.rkt` generates 49 constants from one macro. Real, exported, human-typed names no static parser
  can reach. This is the bulk of the 485.
- **`rename-out` is not represented and cannot be** — the index has an import graph and no export graph. `(require (rename-in "m.rkt" [f g]))` records `f`, the definition-site name, matching the reduction `_clean_names` already applies to `import {a as b}` and Gleam's `X as Y`.
- **`callable_unknowable`**: `(define curry (make-curry #f))` is callable and no syntactic test can know it, so it reports as `constant`. Zeroing that needs an evaluator.
- **`.scrbl` is deliberately unsupported, on a measurement** — a Scribble file parses with `has_error: False` and yields garbage (every prose word becomes a top-level symbol; `@defproc[(greet …)]` extracts nothing). A green parse with an empty result and no error signal is worse than no support.

## Three guards the grammar makes invisible

The harness found all three:

1. `quote` / `quasiquote` / `syntax` / `sexp_comment` are **named wrapper nodes holding a real `list`**, so without a skip `#;(define x 1)` — how Racketeers disable code — extracts as a live symbol.
2. The head symbol is **not** lowercased. The neighbouring Common Lisp extractor lowercases because CL's reader upcases; Racket is case-sensitive, and copying that would make `(Define x 1)` a definition.
3. **Descent is an allow-list.** A `define` inside a `when` body or a macro invocation's clause is an internal definition; descending into unrecognised forms claimed importable bindings that do not exist (measured on `racket/private/dict.rkt` and `racket/set.rkt`).

Three times the **oracle** turned out to be wrong rather than the parser — submodule names, a name defined twice in one file, and class members — each fixed in the harness rather than by suppressing true output. Class-member fidelity is explicitly **not** measured by the harness and says so; unit tests cover it.

## Notes for review

- `tests/test_racket_fidelity.py` gates the two must-be-zero buckets against a frozen copy of the expander's answer, so they run **without Racket installed**. The full sweep stays a benchmark (Racket is a dev prerequisite, not a package dependency).
- `_IMPORT_PATTERNS` / `_DEF_PATTERNS` entries in `plan_refactoring.py` are required — `TestLanguageCoverage` iterates `LANGUAGE_REGISTRY`.
- A `racket` case was added to `tests/test_byte_offset_symbol_names.py::CASES` and verified **non-vacuous** by reintroducing the #414 idiom and confirming it goes red.
- **No `PARSER_GENERATION` bump.** That counter is for files whose content is unchanged *and already in the index*; `.rkt` files are absent from every existing index, so `detect_changes_with_mtimes` reports them as new (`sqlite_store.py:2214`).

### Pre-existing issue found, not fixed here

A config carrying an explicit `languages` list — which `jcodemunch-mcp init` writes — will not enable a newly added language, and `config --upgrade` only injects entirely missing **keys**, so it does not repair the list. This affects
every language ever added, not just Racket; fresh configs activate Racket automatically. Noted in the CHANGELOG. 

## Testing

`8255 passed, 134 skipped` — 8390 total, up from 8332 on this base, all new tests passing. `uv run ruff check src/` clean.

The single failure, `tests/test_suggest_corrections.py::test_freshness_correction_no_patch`, is **pre-existing** and reproduces on a clean tree at this base.

End-to-end: indexing the Racket collects tree (960 files) yields 4,573 Racket symbols — 3,136 function / 1,047 constant / 388 class — with submodule qualification (`base::reader`) and 891 docstrings.


jgravelle (approved):
Reviewed and merging. Three things I checked specifically, and all three hold.

**The fidelity harness earns the title.** `extra: 0` and `wrong_span: 0` measured against Racket's own expander across 211 files and 3,526 definitions, with those two buckets gated in CI off frozen oracle data so the check runs on machines with no Racket installed. Splitting "we missed something" from "we said something false" and giving them different bars is the right decomposition, and it is the half most language PRs skip.

`syntax-original?` to separate human-typed names from macro-introduced ones is what makes the coverage number mean anything, and `module->exports` for the post-`rename-out` view answers a genuinely different question rather than the same one twice. Naming what the harness does NOT measure — class members as a category error, `methods_unscored` kept visible, `files_oracle_failed` never dropped from the denominator — is the part I would have had to ask for.

**The extractor follows the existing custom-parser precedent** rather than inventing a shape. Dispatch on head-symbol text with `is_named` filtering is the correct answer to a homoiconic grammar, and it matches `_parse_clojure_symbols`. `struct` -> `class` is justified against `VALID_KINDS` and the Common Lisp `defstruct` mapping instead of asserted. Treating the `define/public` family as `method` unconditionally, so an unrecognised `class*` shape still kinds correctly, is a good call.

**No `PARSER_GENERATION` bump needed, and I checked rather than assumed.** That counter forces a re-parse of files already in an index; `.rkt` was previously `wrong_extension` and absent from every index, so Racket files arrive through discovery on the next index rather than needing existing rows re-read. Coverage change, not an extraction change to stored symbols.

Two non-blocking nits I will fix in the release commit rather than send you another round-trip for: `HOMOIENIC` should be `HOMOICONIC`, in both `extractor.py` and `languages.py`.

`tests/fixtures/racket/REGENERATE.md` saying "Regenerating is not a way to make a red test green" is the sentence that makes the frozen oracle trustworthy. Thanks for that.

jgravelle:
Approved above — I tried to merge and GitHub stopped me because this is still a draft, which is your call to make, not mine. Mark it ready whenever you are, and it goes in.

One thing you should know rather than discover: **v1.108.298 ships in the next few minutes and its release commit touches `CHANGELOG.md`, so it will put this branch into conflict.** Our rule is that a mergeable contributor PR goes in before any changelog-touching work of ours, and I would have honoured that here — the draft flag is the only reason I am not.

**That conflict is ours, and we will resolve it.** You do not need to do anything about it, and you should not read a red mergeability box as a problem with your work. If it is in the way before you mark this ready, say so and we will push the resolution to `feat/racket-language-support` ourselves.

Nothing about the timing reflects on the PR. The review is done and the verdict is yes.

PR #549, opened 2026-08-25
Racket: index the bindings that `struct` forms generate

otherjoel:
## Racket: index the bindings that `struct` forms generate

An improvement to #548.

### Why these were left out originally

PR #548, which introduced support for Racket, did not attempt to index bindings introduced by macros because doing so accurately would require fully expanding the code. This PR revisits the special case of bindings introduced by `struct` and `define-struct` because they are so common and follow known rules.

### What this does

Derives the generated bindings and points them at the struct form, so `get_symbol_source("posn-x")` returns the `struct` that produces it. That is the answer DrRacket gives for "where does this come from".

| Form | Now indexed |
|---|---|
| `(struct posn (x y))` | `posn`, `posn?`, `posn-x`, `posn-y` |
| `(define-struct posn (x y))` | the above, plus `make-posn` |
| `(struct posn (x y) #:mutable)` | the above, plus `set-posn-x!`, `set-posn-y!` |
| `(struct posn (x [y #:mutable]))` | `set-posn-y!` only |
| `(struct posn (x y [z #:auto]))` | `posn-z` alongside the rest |
| `(struct derived base (c))` | `derived`, `derived?`, `derived-c` — **not** `derived-a` |
| `(struct posn (x y) #:constructor-name make-a-posn)` | `make-a-posn`, and **not** `make-posn` |
| `(struct posn (x y) #:name posn-type)` | `posn-type`, as a type rather than something callable |
| `(serializable-struct posn (x y))` | same set as `struct` |

Two rules are easy to get backwards. **Inherited fields keep the supertype's accessors** — `(struct derived base (c))` binds `derived-c` and not `derived-a` — so only the struct's own field list may be used; and because the supertype occupies the slot before that list, the field list is the *first* list child rather than a fixed index. That is also what makes `(serializable-struct/versions posn 1 (x y) ())` resolve to `(x y)` rather than the trailing `()`. Separately, **`#:constructor-name` replaces the default constructor where `#:extra-constructor-name` adds one.**

`struct:posn` is deliberately not emitted. It is a struct-type descriptor almost nobody calls directly, and one more symbol matching every query for the struct is ranking noise.

### Struct options were checked individually

Seventeen variants were compared against Racket's expander and all match exactly: `#:transparent`, `#:prefab`, `#:authentic`, `#:sealed`, `#:inspector`, `#:reflection-name`, `#:guard`, `#:property`, `#:methods`, `#:auto` with `#:auto-value`, the constructor- and name-keywords, inheritance, zero-field structs, and several combinations.

They are safe for a structural reason rather than by luck: options always follow the field list. Two would defeat a sloppier rule. `#:guard (lambda (a b n) ...)` and `#:property prop:procedure (lambda (s) 1)` take a `(lambda ...)` argument, which is another list node, so a field-list rule that took the *last* list, or a fixed index, would read the guard as the fields. And `#:methods gen:custom-write [(define (write-proc ...) ...)]` contains real `define` forms; those are class members rather than module-level bindings, and they do not leak. There is a test for each.

### Two bugs this also fixes

`serializable-struct` and `serializable-struct/versions` were absent from the form table entirely, so they produced **no symbol at all** — not even the struct name. In a real Racket project I tested against, that was 21 struct forms producing nothing.

`#:name` and `#:extra-name` both bind their argument as an additional struct-type transformer, and it was being dropped, so `(struct posn (x y) #:extra-name posn-type)` lost `posn-type`. It is now emitted as a `type` rather than a `function`, because it names the struct type and is not callable — claiming otherwise would be the same class of error as labelling a procedure a constant.

### The fabrication the benchmark caught

The `#:constructor-name` rule shipped wrong in my first attempt. I emitted `make-<name>` for every `define-struct`, which invented `make-base-object/c` for `(define-struct base-object/c (...) #:constructor-name NEVER_CALL_THIS)` in `racket/private/object-c.rkt`.

That was the only fabricated name across 211 files, and `benchmarks/racket_fidelity/` caught it in the `extra` bucket — the one that must stay at zero, because a name in the index that Racket does not bind is something an agent will trust and chase. Reading the code would not have found it. Every rule above was subsequently checked against `expand` output for that exact variant rather than against the documentation.

### Measurement

Against merged `main`, using `benchmarks/racket_fidelity/`:

| | merged main | this PR |
|---|---|---|
| symbols indexed (211-file corpus) | 3,060 | **3,438** |
| names Racket does not bind (`extra`) | 0 | **0** |
| definitions outside their reported line range | 0 | **0** |
| human-written definitions found | 86.2% | 86.5% |

The coverage figure barely moves, and that is expected: struct accessors are macro-generated, so the expander classifies them outside the human-written set that percentage measures. The real effect is the symbol count and what becomes findable.

End to end, the Racket collects tree goes from 4,573 to 5,680 symbols and a real Racket project from 1,754 to 1,935. In that project `bot-stepper-id` went from no match at all to an exact hit at confidence 0.797.

`benchmarks/racket_fidelity/results.json` is regenerated in this diff so the committed artifact matches the code that produced it, per the repo's rule against hand-typed benchmark numbers.

### Testing

8,423 tests, up 27 — one per row of the variant table and one per struct option, including a regression test naming the `#:constructor-name` fabrication. `uv run ruff check src/` clean. The single failure, `tests/test_suggest_corrections.py::test_freshness_correction_no_patch`, is pre-existing on `main` and unrelated to this change.


jgravelle (approved):
Approved. The `#:constructor-name` write-up is the part that decides this: the rule was wrong, `extra` caught it on one name in 211 files, and the fix is a keyword scan rather than a wider net. That is the harness from #548 doing the job it was built for, on its author.

Checked the two rules that are easy to get backwards, both against the diff and against the grammar:

- **Field list is the first `list` child after the name.** This is what makes `(struct derived base (c))` and `(serializable-struct/versions posn 1 (x y) ())` both resolve, and it is what stops `#:guard (lambda (a b n) ...)` and `#:property prop:procedure (lambda (s) 1)` from being read as fields. A last-list or fixed-index rule fails all four; there is a test for each.
- **`#:name` / `#:extra-name` emitted as `type`, not `function`.** Right call — a struct-type transformer is not callable, and labelling it one is the same class of claim as the fabricated `make-base-object/c`.

`serializable-struct` and `serializable-struct/versions` producing no symbol at all, not even the struct name, is the quieter of the two fixes and probably the bigger one for anyone using `racket/serialize`.

Numbers reconcile against `results.json`: 3,060 -> 3,438 symbols, `missing` 485 -> 475, `extra` and `wrong_span` both still 0, coverage 86.2 -> 86.5. All 12 checks green on the merge ref.

One thing to fix on our side after merge, not a change request: **the coverage figure has three mirrors and this diff moves one of them.** `results.json` is regenerated, but `LANGUAGE_SUPPORT.md` still reads `86.2%` and `benchmarks/racket_fidelity/README.md` still reads `485` / `152 of the 211` / `"86.2%"` (the derived figures go to 475 / 153; the "10 worst account for 311" line happens to still hold). Nothing binds those to `results.json`, so the drift is silent — which is our missing ratchet, not your omission. We will sync both and add the guard.

Merging.

jgravelle:
Merged. The three-mirror sync is done in 72ef602 — `LANGUAGE_SUPPORT.md` and `benchmarks/racket_fidelity/README.md` now read 86.5 / 475 / 153, and `tests/test_racket_fidelity_artifacts.py` derives all six figures from `results.json` so the next regeneration cannot leave a prose copy behind. Five of its assertions fail against the pre-fix tree; the three that pass are the figures your change genuinely did not move.

Thanks for both of these — the fidelity harness from #548 is already paying for itself.

Issue #553, opened 2026-08-26
search_ast returns zero matches for ALL languages — encoder schema declares 'results' but tool returns 'matches' (schema/producer key mismatch)


RascoApps:
## Summary

`search_ast` returns zero matches for **every language and every preset/category**, not just C#. The compact encoder schema (`encoding/schemas/search_ast.py`) declares table key `results` and scalar `result_count`, but the tool (`tools/search_ast.py`) returns `matches` and `total_matches`. The encoder's `response.get("results", [])` finds nothing and silently produces an empty table.

This is the same class of bug as the `search_text` #246 fix (flat columns declared, nested/different shape returned). The "fail closed" guard at `schema_driven.py:194` only catches "rows exist but all columns empty" — it does **not** catch "rows under a different key entirely."

## Reproduction

```
search_ast(repo="<any indexed repo>", pattern="god_function", language="csharp")
→ #MUNCH/1 tool=search_ast enc=sa1
  repo=... pattern=god_function __tables=a:results:file|line|match_type|snippet|symbol_id|symbol_name
  (empty — no rows)
```

Same result for `category="all"`, `pattern="todo_fixme"`, `pattern="lines:80+"` (custom DSL), and all languages (Python, TypeScript, C#). The tool itself works correctly — calling `search_ast()` directly in Python returns real matches. The data is lost in the encoder.

## Root cause

`tools/search_ast.py` returns (lines 1196-1208):

```python
result = {
    "repo": f"{owner}/{name}",
    "total_matches": len(all_matches),
    "severity_counts": {...},
    "matches": all_matches,          # ← key is "matches"
    "truncated": ...,
    "_meta": {
        "files_scanned": ...,        # ← key is "files_scanned"
        ...
    },
}
```

`encoding/schemas/search_ast.py` declares:

```python
_TABLES = [sd.TableSpec(key="results", ...)]     # ← expects "results"
_SCALARS = ("result_count", ...)                  # ← expects "result_count"
_META = ("timing_ms", "files_searched", ...)      # ← expects "files_searched"
```

The encoder does `response.get("results", [])` → `[]` → empty table. The scalar `result_count` is also dropped (not in response). The `_meta.files_searched` is dropped (tool emits `files_scanned`).

Additionally, the declared columns don't match the match dict keys:

| Schema column | Actual match dict key |
|---|---|
| `match_type` | `pattern` |
| `symbol_id` | `enclosing_symbol` |
[...]

PR #554, opened 2026-08-26
Racket: let a project declare its own defining forms

otherjoel:
### This is a speculative PR

It came out of investigating why Racket support misses so much on a DSL-heavy project. A few choices here may be a bigger departure from project norms than you want, so please close it freely if so — it seemed worth building to find out.

**It adds the first language-scoped config key.** Every existing key is global in meaning. I considered a generic `definition_forms: { racket: … }` map and a separate `.jcodemunch.racket.jsonc` sidecar; reasoning for landing on a flat Racket-prefixed key is under *Why Racket-only* below.

**Its correctness rests on a declaration jCodeMunch cannot verify**, which is a departure from the rest of the Racket work, where every rule is checked against Racket's expander. Two ways of inferring it automatically were built and measured first, and both fabricate names; the numbers are under *Why declared rather than detected*. Worth noting the failure mode is mild — a project's own DSL is usually a handful of obvious forms its authors know cold, and omitting the key leaves things exactly as they are today rather than worse.

**It also documents `.jcodemunch.jsonc`**, which is arguably a second change. It has been read since v1.108.197 and appears in no user-facing document, and the feature is unusable without it, so I bundled them — happy to split if you would rather.

Everything else checks out against CONTRIBUTING and the maintenance practices: the catalog moratorium does not apply (no new action), the one `except` logs, `benchmarks/racket_fidelity/results.json` is unchanged, tests do not touch real config, and both quality gates are unaffected.

---

## Racket: let a project declare its own defining forms

Racket projects routinely define their own defining forms with `define-syntax`. A project built on a DSL might write `(defstep (check-admin) ...)` or `(defstudy consent ...)` hundreds of times, and none of those names reach the index, because `(defstep (check-admin) ...)` is textually indistinguishable from a function call.

This adds `racket_definition_forms`, settable per-project in `.jcodemunch.jsonc` or globally:

```jsonc
{
  "racket_definition_forms": {
    "defstep":  "function",
    "defstudy": "constant",
    "defvar":   "constant",
    "define-schema": "class"
  }
}
```

Each entry maps a form to what it binds — `function`, `constant`, `class` or `type`.

**Where the name sits is read from the source rather than declared.** `(defstep (check-admin) ...)` takes the head of the parameter list; `(defstudy consent ...)` takes the bare symbol. That is not merely less to type: a single form is not consistent. Measured on one project, `defstep` appears 44 times in the first shape and once in the second, so a declared position would have missed the odd one out. Inferring it found 63 symbols a fixed declaration did not.

Measured on that project: **1,935 → 2,302 indexed symbols**, with `consent`, `check-admin`, `current-matrix` and `admin` going from unfindable to indexed at their real locations.

### Why declared rather than detected

The obvious move is to guess. I tried twice and measured both against Racket's expander:

| Heuristic | Real names recovered | Names fabricated |
|---|---|---|
| any `def*`-prefixed head is a definition | 140 | **225** |
| …restricted to macros the repo itself defines | 127 | **168** |

`(default d ...)` and `(definify map ...)` are function calls that happen to start with `def`. `(define-logger enter!)` binds `log-enter!-debug`, not `enter!`. Neither can be rescued, because what a macro binds is a property of its expansion, not its call site.

So the only sound source is the user, which makes this an **assertion the project cannot verify** — a wrong declaration indexes a name Racket does not bind, and `benchmarks/racket_fidelity/` cannot catch it, since the harness only knows forms it can see expanded. The README and config template both say so.

### Safety properties

- **Inert by default.** The default is `{}`, and an unconfigured project parses byte-identically — verified by re-running the full fidelity corpus for the same 3,438 symbols and the same zero `extra` / zero `wrong_span`. `results.json` is unchanged in this diff as a result.
- **Declarations cannot shadow real Racket syntax.** Matched after every built-in, so declaring `define` or `struct` gets the built-in handling. Tested, because the first cut placed the check before the struct branch and silently did shadow it.
- **A typo costs one form, not a file.** Entries are skipped individually, `kind` is restricted to four values, and non-string values are `isinstance`-checked first, since `in frozenset` on a dict raises.

### Why Racket-only

Clojure, Elixir and Common Lisp have the same blindness and this codebase serves all three. But none has been measured, and a generic `definition_forms` map would be a general promise backed by one data point. A separate `.jcodemunch.racket.jsonc` was the other candidate; it would need its own loader, cache, precedence rules and validation, which is more new machinery than a key in a file that already exists. If a second language earns one, `<lang>_definition_forms` appears beside this, and unification becomes a decision with evidence behind it.

### Also: `.jcodemunch.jsonc` is now documented

The per-project overlay has been read since v1.108.197 and appeared in no user-facing document. README now covers what it is, what overlay semantics mean (declared keys win, omitted keys fall through to global then default), and the Racket key above.

### Testing

8,475 tests, up 12. `uv run ruff check src/` clean, and the generated config template still validates. The single failure, `tests/test_suggest_corrections.py::test_freshness_correction_no_patch`, is pre-existing on `main`.


jgravelle:
Reviewing this before merge turned up one regression, and it is in what the feature displaced rather than in the feature.

The declared-forms block replaced the `Nothing matched` fall-through guard instead of sitting in front of it. `_RACKET_SPLICING_HEADS` is now defined at `extractor.py:10984` and referenced nowhere, and the `return` that stopped the walker descending into unrecognised forms is gone — so `_walk` now recurses into every form it does not recognise.

Measured on identical input, `origin/main` against this branch:

```racket
(define (real-fn x) (+ x 1))
(some-unknown-macro
  (define hidden-internal 42)
  (define (hidden-fn y) y))
```

```
origin/main   ['real-fn']
this branch   ['real-fn', 'hidden-fn', 'hidden-internal']
```

Those two are internal definitions inside a macro invocation. Emitting them claims importable bindings Racket does not create — the fabrication class `benchmarks/racket_fidelity/` holds at zero, and the same shape as the `make-base-object/c` case your harness caught during #549.

Still present in `b471cd3` after your rebase, so the force-push carried it forward.

## Why nothing caught it

All eight CI jobs were green, and so was our full local suite — 8,482 tests. Every assertion over that code path is positive: they check that a splicing head **is** descended into. `test_racket_language.py:207` asserts `t-helper` is found inside `(module+ test ...)`, and that still passes with the guard deleted, because descending into everything includes descending into `module+`. Nothing asserted the absence, so removing the guard was invisible to every signal either of us reads.

That is the part I would keep from this even after the code is fixed. A guard covered only by positive assertions can be deleted without a single test noticing.

## What I have ready

Restoring the guard after the declared-forms block, plus two tests — one for each direction, so a future early-return that suppresses real symbols fails the second one. The absence test fails against the pre-restore tree and passes after. Suite green on your rebased head.

Your call on who commits it. Happy to push it onto this branch if you would rather not context-switch, or leave it to you if you would rather it be yours — you are clearly mid-flight on this and I do not want to force-push over you again. Either way it is a small diff and the feature itself is unaffected.

One thing that was ours, not yours: the conflict that hit this branch earlier came from our own CHANGELOG edits in `4fa46a1` (#553), which is why it went `CONFLICTING` and lost CI for a while. You have already resolved it in the rebase, so nothing is outstanding there.

Also worth flagging so it does not read as a new failure: whichever of us pushes next, `license/cla` will show as **absent** on the new head. It is a per-SHA legacy status that does not follow a new commit, and the gate cannot tell "not reported" from "not signed", so it fails closed to `BLOCKED`. Your signature is stored per account and is not in question — it needs a status posted against the current head, which I can trigger from our side.

otherjoel:
Confirmed, and thank you — this is my regression and you're right about the mechanism. `_RACKET_SPLICING_HEADS` was left unreferenced when I moved the declared-forms block; the splice took the `Nothing matched` return with it.

It's worse than your minimal case shows. On the fidelity corpus:

| | before | with the guard gone |
|---|---|---|
| `extra` | 0 | **5** |
| `wrong_span` | 0 | **26** |

The five are `cmp/c`, `elem/c`, `equal-key/c`, `kind/c`, `lazy?` in `racket/set.rkt` — the exact names #548 added the guard to suppress, so this had silently reverted that fix.

And `results.json` was wrong in the diff. I regenerated it *before* the displacing edit, so the committed artifact asserted `extra: 0` while the code produced 5. That's the real process failure on my side: I treated the harness as a one-time check rather than a gate on the final tree, and it would have caught this in ninety seconds.

**Your point about positive assertions is the part I'd keep too**, and I verified it rather than taking it on faith — deleting the guard again, 86 of 87 Racket tests still pass. Only an absence test sees it.

Fixed in `a9d31cc`, which I've pushed rather than taking you up on the offer, since it was mine to fix and I had the context loaded. It does a little more than you outlined:

- Guard restored after the declared-forms block, with a comment recording how it was lost.
- Both directions pinned, as you suggested: `test_unrecognised_forms_are_not_descended` fails without the guard, `test_splicing_forms_are_still_descended` fails if a future early return over-suppresses, since `begin` genuinely splices.
- `results.json` regenerated; it now reproduces from the committed tree.
- **The CI-safe path had the same hole.** No frozen fixture wrapped a `define` in an unrecognised form, so `test_racket_fidelity.py` couldn't see it either, with or without Racket installed. `guards.rkt` now carries that shape — verified against the expander that `macro-body-local` and `macro-body-fn` are not module bindings — and `test_no_fabricated_symbols[guards.rkt]` now fails on a Racket-free machine when the guard is removed. That's the gap that let it reach you.

Suite 8,495, lint clean, harness back to `extra: 0` / `wrong_span: 0`.

Noted on the CHANGELOG conflict and on `license/cla` being per-SHA — thanks for flagging both.

jgravelle:
Merged in `3735d80`. Verified your fix independently before merging rather than taking the description on trust — guard live at `extractor.py:11493`, unknown-macro bodies no longer leak, `module+` and `begin` still splice, and both new tests fail against the reintroduced defect. Suite green here at 8,495 total, matching your count.

Two things you found that I did not, and both are worth more than the fix:

**The CI-safe path had the same hole.** My patch would have closed the defect and left `test_racket_fidelity.py` blind, because no frozen fixture wrapped a `define` in an unrecognised form. `guards.rkt` closes the route this took to reach me, not just the defect. I confirmed `test_no_fabricated_symbols[guards.rkt]` fails with the guard removed on a machine with no Racket installed, which is the property that matters.

**I under-reported the severity.** I measured a two-line synthetic case; you measured the corpus — `extra` 0 → 5, `wrong_span` 0 → 26 — and identified the five names in `racket/set.rkt` as the exact ones #548's guard was added to suppress. That reframes it: not a new defect, a silent revert of a shipped fix. Worth correcting in the record.

The `results.json` point is the one I would put in the project's standing notes. A committed artifact asserting `extra: 0` while the tree produces 5 is the same failure we hit in 1.108.298, where eight artifacts mirrored one benchmark run and five were stale while both sync tests passed. Regenerating before the last edit is exactly how it happens. You caught yours in the same PR; ours took twenty-two days.

Thanks for taking the fix rather than handing it back — and for checking the positive-assertion claim instead of accepting it. 86 of 87 passing with the guard deleted is a better statement of the problem than anything I wrote.

Issue #550, opened 2026-08-25
`from . import <sibling>` resolves to `__init__.py`, so 20 live files in this repo are reported dead


rknighton:
## The one thing that is wrong

`from . import receipts` in `evidence/producers.py` is a dependency on
`evidence/receipts.py`. JCodeMunch records it as a dependency on
`evidence/__init__.py` instead, and the edge to `receipts.py` is never built.

The shipped source uses that form 49 times. Re-indexing this repository with
only the import specifier changed, nothing else, drops
`find_dead_code(granularity="file")` from 42 dead files to 22. The other 20,
`runtime_identity.py` and nine `parser/context/` providers among them, are
reported dead only because of this.

## Reproduction

```powershell
uv venv --python 3.13 .venv
uv pip install --python .venv\Scripts\python.exe jcodemunch-mcp==1.108.298
.venv\Scripts\python.exe repro_sibling_import.py
```

```python
"""`from . import <sibling>` resolves to __init__.py, and what that costs.

Part 1 isolates the resolver behavior against a dotted-import control.
Part 2 re-indexes jcodemunch_mcp's own installed source twice, changing only
the import specifier, and diffs find_dead_code.

The rewritten copy is indexed, never imported or executed.
"""

from __future__ import annotations

import platform
import re
import shutil
import tempfile
from importlib.metadata import version
from pathlib import Path

import jcodemunch_mcp
from jcodemunch_mcp.parser.imports import _extract_python_imports, resolve_specifier
from jcodemunch_mcp.tools.find_dead_code import find_dead_code
from jcodemunch_mcp.tools.index_folder import index_folder

BARE_DOT = re.compile(r"^(\s*)from \. import ([A-Za-z_][A-Za-z0-9_]*)(.*)$", re.M)
FILES = {
    "src/demo/__init__.py",
    "src/demo/broken_a.py",
    "src/demo/broken_b.py",
    "src/demo/other.py",
}


def part1() -> None:
    print("Part 1: resolver, with a dotted-import control")
    for spec in (".", ".broken_b"):
        got = resolve_specifier(spec, "src/demo/broken_a.py", FILES)
        print(f"  resolve_specifier({spec!r:12}) -> {got!r}")
    print()
    print("  extractor keeps the sibling name but resolve_specifier never sees it:")
    print(f"    from . import broken_b        -> {_extract_python_imports('from . import broken_b')}")
[...]

PR #556, opened 2026-08-27
Racket: read the #lang line first, resolve collection paths, and close the fidelity gaps the review found

otherjoel:
## Summary

Follow-up to #548/#549/#554. I reviewed the Racket support against three ground truths — Racket's expander (the existing fidelity harness), Racket's reader (`read-syntax` with each `#lang`'s own reader, for langs the expander cannot instantiate), and five real package-layout projects on disk — and fixed what the review found. Twelve commits, one finding per commit, each with its own tests and CHANGELOG entry, ordered by severity: things the index said that were false first, reachable misses second, noise last. The series can be reviewed commit by commit, and if any of it should be its own PR, the commits split without rework.

**One deliberate departure from the written rule, stated up front:** `PARSER_GENERATION` is **not** bumped, although every change here alters what unchanged `.rkt` content yields and the `#lang` gate removes fabricated symbols that a 1.108.297–.301 index still holds. The counter is one integer for the whole tree, so a bump re-parses every language for everybody, and Racket support was three days old with — as far as I know — one user. Commit 12 substitutes a narrower mechanism: every local index is now stamped with `racket_config_digest` at save (`""` when unconfigured), an index with **no** such meta key is one that predates the stamp, and a local index holding Racket files with no key re-parses once (`rebuild_reason: "racket_index_predates_gate"`). That reaches exactly the indexes that need it and, unlike a skipped bump, stays repairable at any later date — an absent key is detectable forever, a stamp equal to the constant is not. If you prefer the global rule, restoring the bump is one line beside the constant, and the note there says so.

## Headline numbers

| Measurement | Before | After |
|---|---|---|
| Fidelity corpus (211 files, Racket v9.2 collects): `missing` | 475 (86.5%) | **362 (89.7%)** |
| Fidelity corpus: `extra` / `wrong_span` | 0 / 0 (with a named exemption for `define-generics`) | **0 / 0, exemption removed** |
| Files with nothing missing | 153 of 211 | **171 of 211** |
| `#lang conscript`, 207 files, reader oracle (with `racket_langs: {"conscript": "at-exp"}`) | 39% found, **~100 fabricated** | **0 missing, 0 wrong_span** |
| `#lang punct`, 94 files | 1 symbol (correct), but a `(define …)` in a Markdown code sample would index | **0 symbols, text-searchable** |
| `#lang racket/base`, 712 of my own files, reader oracle | 0 parse errors, 0 missing, 0 wrong_span | unchanged |
| splitflap (`foo-lib/` + `(define collection "foo")` layout): resolved `require` edges | **0 of 70** | 13; 7 of 17 files gain an importer |
| congame: resolved `require` edges | 304 | **624** |

## Commits

| # | Commit | Finding | Fix |
|---|---|---|---|
| 1 | `#lang` gate | tree-sitter-racket parses S-expressions; a `#lang` names a reader. Every `.rkt` was parsed as `racket/base`: `#lang punct` (Markdown) indexed code samples, `#lang conscript` (at-exp) lost 61% of definitions to four prose characters (`;` `"` `#` `\|`) and fabricated ~100 internal defines that error recovery re-parented to module level (`list → ERROR → program`). | Three tiers decided from the `#lang` line before the grammar runs: `sexp` (walk), `at-exp` (blank `{…}` bodies byte-for-byte, walk the paren skeleton), `text` (no symbols). Unlisted langs are `text`; `racket_langs` in config promotes a project's own lang. `ERROR` nodes are skipped in both directions and the file is named at WARNING. |
| 2 | Docstring adjacency | A trailing comment on the previous form's line, and a file-header block separated by a blank line, became the next form's docstring — prose served as fact. | The chain must end on the line directly above the form and stops at a trailing comment. |
| 3 | Call references | Parameters, `let`/`for` clause heads (only 7 of the `for/*` forms were listed), `match` patterns, struct field lists and `provide` specs were recorded as calls of the enclosing function — or, for a struct, of whichever synthesised accessor was emitted last (`posn-y calls=['x','a','values']`). These feed `get_call_hierarchy`, blast radius and `get_untested_symbols`. | Headers skipped whole; every `for*` variant matched by prefix; clause and pattern forms walk only their values; struct family and `provide`/`require` not descended; `send`/`new` record the method/class. |
| 4 | Knowable kinds | `(define/contract f (-> …) (lambda …))` and Typed `(define f : T (lambda …))` were `constant` — the value is not at `children[2]` for either. `define-syntaxes` contradicted the "a macro is a function" rule two blocks away. | Value located by shape, annotation in the signature, `match-lambda`/`thunk` join the lambda heads. |
| 5 | Struct and generics | `define-generics` emitted the bare stem — a name Racket does not bind — and none of the methods; the harness forgave it by name in `_oracle_knows`, so `extra: 0` was carrying a fabrication. `(define-struct (child parent) fields)` — the old supertype header, 130 uses in 36 collects files, 283 in 66 pkgs files — yielded nothing at all. | Emit `gen:<name>`, `<name>?`, `<name>/c` and each method; exemption removed, still 0 `extra`. List header handled; `#:type-name` binds a type. |
| 6 | Binding-form tables | `begin-encourage-inline` was not a splicing head (hid `sqr`, `sgn`, `conjugate`, all of math-predicates.rkt); `define-sequence-syntax` hid `range` and 19 names in for.rkt; `define-syntax-parse-rule` — the *current* name of `define-simple-macro`, which was listed — was not; `define-syntax-class`, `define-syntax-parameter`, `define-logger`, `define-inline`, `define-check`, `define-unit` all `(no symbols)`. 113 of the 475 misses were table entries, not macro output. | Listed under the kind each header implies; `define-logger` synthesises `app-logger` and `log-app-<level>` the way struct accessors are. |
| 7 | Require wrappers | `(for-syntax racket/base "helpers.rkt")` kept the first path and dropped the rest (166 multi-path wrappers in pkgs); `(for-meta 1 "m.rkt")` recorded the phase level `1` as a module path; `(submod "other.rkt" sub)` was dropped as if intra-file; `\brequire\b` matched `require-syntax`. | The reader returns every (path, names) pair a sub-form carries. |
| 8 | Collection paths | A collection path names a directory `info.rkt` declares, not a repo path: `(require foo/bar)` means `foo-lib/bar.rkt`. The resolver only ever tried `foo/bar.rkt`, which exists when the repo *is* a collects root (the harness corpus) and in no package. `LANGUAGE_SUPPORT.md` said collection paths resolve. | `build_racket_collection_map` reads every `info.rkt` (`'multi`, and several dirs declaring one name — congame-cli/core/doc all say `"congame"`); `augment_racket_collection_edges` adds a root-relative edge beside the original at `CodeIndex` construction, the #550 shape, so the 26 `resolve_specifier` call sites are untouched. |
| 9 | `module+` and annotations | Repeated `(module+ test …)` blocks emitted two `class` symbols with the same id, and `symbols.id` is a PRIMARY KEY. `(: …)` annotations were a single last-seen slot; the infix `(: g : A -> B)` rendered `: :`. | One symbol per submodule; annotations keyed by name. |
| 10 | Config-change re-parse | `racket_definition_forms` (#554) applied to nothing on an existing index: it changes what the parser emits for identical bytes, and the incremental path skips identical bytes. Measured: `check-admin ABSENT` after adding the key and re-indexing, present only after a full reindex. The shared parse cache had the same hole. | Digest of both keys stamped on the index; mismatch → one full re-parse (`rebuild_reason: "racket_config_changed"`); digest in the parse-cache key. Six end-to-end tests through `index_folder`. |
| 11 | Fixtures and artifacts | — | `forms.rkt` (one instance of every form that yielded nothing, the wrong kind or an unbound name) and `atexp.rkt` (asserts the raw bytes *still* fail the grammar) join the CI-safe fidelity gate with frozen expander answers; `results.json`, harness README, `LANGUAGE_SUPPORT.md`, README regenerated; `test_racket_fidelity_artifacts.py` binds the figures. |
| 12 | Unstamped-index re-parse | See the note at the top. | `racket_reparse_reason()` replaces `racket_config_changed()`; an absent stamp on a local index holding Racket files re-parses once. |

## Tests

`ruff check src/` clean. Full suite on the branch: **8,518 passed / 136 skipped / 1 failed**. The one failure, `tests/test_suggest_corrections.py::test_freshness_correction_no_patch`, is pre-existing and environmental — it fails identically on `main` in a clean worktree on this machine, because a jCodeMunch watch service (`us.gravelle.jcodemunch-watch`) is running here and the tool truthfully says so instead of recommending `watch-install`. It reads the developer's real machine state, which is the Practice-8 shape; worth its own issue, not this PR.

Every new test is an absence test where it can be: the fabrication class is pinned structurally (the recovery test asserts the `ERROR` ancestry exists before asserting the name is absent), the harness exemption is gone rather than widened, and the collection-map tests go through `resolve_specifier` for every added edge because an edge nothing downstream can resolve is indistinguishable from no edge.

## Deliberately not in this PR

- `.rktd` still indexes as a zero-symbol source file, which `find_dead_code` reports dead (it is data, never `require`d).
- `callable_unknowable` still counts plain `struct` constructors and `(module name …)` wrappers; the harness README now says so instead of calling them unknowable.
- A partial parse is a WARNING log line naming the file, not a per-file coverage signal.
- `CLAUDE.md` untouched — the Current State rotation is yours.

## What comes next

The at-exp handling here is a byte-level pre-pass that blanks `{…}` text bodies before tree-sitter sees them; it is measured and honest about being a trick, but nothing in the stack actually reads at-expressions, and the grammar we ship (`6cdh/tree-sitter-racket`, precompiled in `tree-sitter-language-pack`) cannot be taught to. The next PR will attempt a hand-rolled Racket reader in Python — the default reader plus at-exp, handing the walker nodes with the surface it already consumes — which would replace the pre-pass and remove the error-recovery class entirely: a reader fails at a position and resumes at the next top-level form, so a stray `)` costs one form rather than the rest of the file, and nothing is ever re-parented under an `ERROR` node. Its acceptance bar is the harness in this PR: `extra` and `wrong_span` stay at zero, conscript stays at zero missing *without* the blanking, and the stray-paren case recovers the definitions after it. Anything in the review that a reader would remedy on its own — an at-exp heuristic for unknown langs, resuming after a parse error, the known blind spots of the blanking (`@cmd["a }"]{…}`, `|{ … }|` delimiters) — was left out of this PR on purpose rather than built twice.

Racket v9.2 [cs] was used for the expander and reader measurements; the CI-safe gate needs no Racket install.


jgravelle (approved):
Merging. Trial-merged onto current \`main\` and ran the suite there rather than reading the branch result: **8,706 passed / 13 skipped / 0 failed**, \`ruff check src/\` clean. Your one reported failure (\`test_freshness_correction_no_patch\`) passes here, which supports your read that it is your machine's watch service and not the PR — it is the Practice 8 shape and worth its own issue.

Twelve commits, one finding each, ordered by severity, every one measured against a ground truth rather than against our own output. The `define-generics` exemption being **removed** rather than widened is the part I want to call out: `extra: 0` was carrying a fabrication, and an exemption that hides one is worse than a red bucket.

On the `PARSER_GENERATION` departure — you argued it well and the mechanism is better than the rule it declines. "An absent key is detectable forever, a stamp equal to the constant is not" is exactly the failure that cost us the `refresh` campaign in .302, and a per-project digest reaches config changes a global counter never can.

One thing you could not have known when you wrote it: **this release bumps `PARSER_GENERATION` 6 -> 7 anyway**, for an unrelated Rust fix that landed hours before yours. So the cost you were avoiding is already being paid, every index re-parses regardless, and your Racket changes reach existing indexes through the bump as well as through your stamp. That takes nothing away from the stamp — it stays the only mechanism that catches a `racket_langs` or `racket_definition_forms` edit, which is what commit 10 is actually for. I will amend the one clause on `racket_config_digest` that now reads as if no bump were happening, in a follow-up commit of ours, not yours.

PR #564, opened 2026-08-30
Racket: a reader in Python, measured against Racket's reader

otherjoel:
## Racket: a reader in Python, measured against Racket's reader

This is the follow-on #556 promised: `.rkt` files are read by a Racket reader written in Python (`src/jcodemunch_mcp/parser/racket_reader.py`) instead of tree-sitter-racket, and the brace-blanking pre-pass is deleted. The reason is the one #556's description gave for deferring it: a `#lang` line selects a *reader*, and a grammar cannot follow it. at-exp text bodies were prose to Racket and tokens to the grammar, and the grammar's error recovery re-parented internal definitions to module level, which is the one thing an index must never do.

Seven commits, one finding each, each with its test and CHANGELOG entry. As with #556, every number below is read off a harness run against Racket itself, not off our own output.

### What the reader is

It implements the default reader (the "Reader" chapter of the Racket reference — every `#` dispatch, `|sym bol|` and `\ ` escapes, here strings, the full number grammar per radix, `#;` / `#| |#` / shebang comments, infix dots) and the at-exp extension exactly as `scribble/reader` implements it, read off that source: `@` as a non-terminating dispatch character, the dispatch order (`@;` → quote prefixes → `{`/`|…{` body → `[` datums → `|…|` escape → command datum, with `|` terminating the command), balanced braces, alternate delimiters with mirrored closers, `@|expr …|` and `@"string"` escapes, `@;` in both spellings, and the command readtable inside escapes. The command character is a parameter.

It produces the tree shape tree-sitter-racket produced — same type names, comments as named children, `dot` nodes, quote wrappers, `#(…)` as `vector` around `list` — so the walker in `extractor.py` is **unchanged**. An at-form `@cmd[d]{t}` is a `list` from the `@` to the closing brace, because that is what `read` produces (`(cmd d "t")`), and a command-only form (`@foo`, `@(expr)`) is just the command node. Both are what `read-syntax` reports, position for position.

`@` is never inferred from the text. In the default reader it is a symbol constituent and `(define @foo 1)` binds `@foo`; the `#lang` tier from #556 is what switches at-exp mode on, unchanged.

### How it was measured

**A new harness, `benchmarks/racket_fidelity/run_reader_fidelity.py`, compares the tree against `read-syntax` node for node.** `reader_oracle.rkt` reads each file with the reader its own `#lang` selects and emits every syntax object as (type, byte start, byte span); line counting is off, so `syntax-position` counts bytes and nothing converts. Three buckets gate at 0: `only_racket`, `only_ours`, `our_only_error`.

| Corpus | Files | Nodes compared | at-forms | only_racket | only_ours | our_only_error |
|---|---|---|---|---|---|---|
| The whole `collects` tree + every `#lang at-exp` file in the distribution (committed `reader_corpus.json`) | 725 (45 at-exp) | 761,009 | 3,622 | 0 | 0 | 0 |
| `#lang conscript` projects on my machine, lang promoted to at-exp | 152 | 53,269 | 3,283 | 0 | 0 | 0 |
| `#lang pollen/mode racket/base` files on my machine | 7 | 5,977 | 51 | 0 | 0 | 0 |

What the harness deliberately does **not** compare is written in the results file and the README: strings *inside* an at-exp body (the at-exp reader splits and merges body text in ways the walker never reads; the form's own span is compared), comments, the contents of `#hash`/`#s` literals (Racket does not position their keys), and the datum after a `#reader <module>` form, which is read by *that* module's reader — 9 in the corpus, counted.

**The harness found three defects in itself before any in the reader**, which I mention because it is the pattern #556 warned about: the oracle spliced a dotted tail's syntax object into its parent (`(a . ,b)` showed a bare `,` symbol), it emitted a sized vector's repeated element three times (`#3(a)` fills by repetition, the same object), and its `#lang` detection had the block-comment gap fixed in the first commit. An oracle is code too.

**A differential against tree-sitter itself** over 2,089 files agreed on all 1.8M nodes except where Racket sided with the reader: `#cs` is a case-fold switch, not a symbol; `#<< eos` (a terminator with a leading space) is a here string; `#\SPACE` is a character.

**The expander harness from #556 is byte-for-byte the committed run** after the swap — 211 files, `extra` 0, `wrong_span` 0, `missing` 362, 89.7% — which is what a behaviour-preserving swap should show, and why the reader was measured on its own first. On 208 conscript files with no blanking: 0 missing, 0 wrong spans.

**Speed:** 17.3 MB of Racket in 1.71 s against tree-sitter's 0.61 s — 2.8× slower, about 10 MB/s, well under the walker's own cost.

**Gated in CI without Racket:** a second frozen oracle, `tests/fixtures/racket_reader_oracle.json`, pins every node Racket's reader produces for the fixtures, the way `racket_oracle.json` pins the expander. Two fixtures are new: `reader.rkt` (one instance of every default-reader form) and `atexp-syntax.rkt` (the Scribble documentation's own `@`-syntax examples, quoted so the file still expands). Both are gated by both oracles.

### What changes for a user

- **Read errors cost the broken form, not the rest of the file.** The reader marks the form's span `ERROR` and resumes at the next column-0 form (Racket rejects the whole file; tree-sitter lost everything after the first error). An extra `)` closes a form early and would leave its remaining internal definitions to be read as top-level forms — exactly the fabrication tree-sitter's recovery produced — so on an unexpected closer every indented top-level form since the last column-0 form is folded back into the `ERROR`. The warning names the file and line now. `test_definitions_after_a_stray_close_paren_are_missed_not_fabricated` had pinned the loss as intended behaviour; it asserts the definitions are found.
- **`racket_langs` may declare a lang's command character**: `"mylang": {"tier": "at-exp", "command_char": "◊"}`, because `make-at-readtable` takes `#:command-char` and Pollen uses `◊`. One non-whitespace character; a malformed entry costs that entry only; it rides in the existing config digest.
- **`#lang pollen/mode racket/base` is at-exp code with `◊`**, built in — it was filed under the document lang `pollen` by the prefix rule. **`#lang rosette` is S-expressions** (its reader is `syntax/module-reader`). Both found by measuring; both were documents before.
- **A `#| |#` block comment above `#lang` no longer hides the lang from the gate.** `openssl/mzssl.rkt` opens with one. Harmless there; not harmless for a document lang behind a licence header, which would have been read as S-expressions.
- **`require` edges come from the reader too.** `imports.py` had its own comment-stripper and form reader — a second, weaker Racket reader — and found `(require` by regex. Measured over 2,489 files: 131 edges the regex produced and the reader does not, all inside `#;` comments, `#'` macro templates, quasiquoted `eval` payloads and here-strings — none a dependency of the file; none missed the other way. `extract_imports` takes `repo` now so a project's own at-exp lang is read the way the walker reads it. **One measured loss, stated plainly:** a document-tier file contributes no edges any more — the regex used to guess at them, and across the distribution and my projects the guess found 60 edges in 17 of 211 document files (`2d`, `scribble/lp`, `beeswax/template`, `rash` …). A reader we do not have cannot say where the Racket in a document is; declaring the lang in `racket_langs` gets its edges back.

### No `PARSER_GENERATION` bump, again

The reader changes what unchanged `.rkt` bytes yield, and the incremental path never re-reads unchanged content. Same answer as #556: `READER_GENERATION` is stamped on every local index holding Racket files beside the config digest, under the same rule (absent means tree-sitter parsed it), and a mismatch forces one full re-parse with its own reason, `racket_reader_changed`. It reaches exactly the indexes that hold `.rkt` files. The shared parse-cache key carries the generation too, because that key carries `INDEX_VERSION` rather than `PARSER_GENERATION` and would otherwise have served a tree-sitter-era parse of identical bytes; a test bumps the constant and asserts the key moves.

### Left out on purpose

- **Scribble (`.scrbl`) support.** The reader's text-at-top-level entry point exists (`read_racket(..., inside=True)`, tested), so the reader half of it is done. What remains is a symbol-model decision — what a `@defproc` or a `@section` *is* in the index, since Markdown headings are not indexed either — plus an extension mapping and one walker table. That is a design to put in front of you, not something to smuggle into a reader PR. The `part` tree from `(dynamic-require path 'doc)` would be its oracle.
- **A comment as the second child of a `define` types the binding `constant`** (`(define x ; doc\n (lambda …))`): the walker reads `kids[2]`, and tree-sitter put comments there as named children, so the reader mirrors it. Unchanged here because the swap was to be behaviour-preserving; one line in `_racket_named` would fix it, with its own measurement.
- **Symbol escapes in binding names** (`(define |foo bar| …)`, `(define \@email …)`) are indexed by their raw spelling. Present in the corpus only inside quoted data, so the expander harness could not see it; the reader makes decoding cheap when it matters.
- **`#lang 2d`**, `scribble/lp`, `beeswax/template`, `rash` and the other document-tier langs in the loss above — each is a reader we do not have.

### Tests

Full suite on this machine, same job both times, one env-dependent test deselected on both (`test_freshness_correction_no_patch`, which reads the developer's real machine state): `main` **8,693 passed, 136 skipped**; this branch **8,843 passed, 136 skipped, 0 failed** — +150, identical skips. `uv run ruff check src/` clean.

Every earlier fidelity claim (#556's harness, its frozen fixtures, its gate tests) still holds. The six Racket test files go from 248 tests on `main` to 398 here, `test_racket_reader.py` alone being 127; every other test file is untouched, so the whole-suite delta is +150. As before, the results files are committed and are the run they say they are. Footprint: 28 files, +2,816 / −285.

https://claude.ai/code/session_015UvRTtv53rU3E2vY15yrsz


jgravelle:
Merged. Ships in 1.108.310.

Two things from the review worth writing down.

**Your `.gitattributes` commit earned itself on its first real encounter, in a case you could not have hit.** The trial merge onto `main` came back **6 red on Windows** while all 8 of your jobs were green. Cause: `.gitattributes` and the fixtures it governs arrived in the same merge, so an existing clone with `core.autocrlf=true` wrote CRLF and moved every byte position the frozen oracle records. A fresh clone — which is what CI does — cannot see it. `test_fixtures_are_checked_out_byte_for_byte` failed first and named the cause, so it was one read instead of an afternoon staring at position diffs. `rm tests/fixtures/racket/*.rkt && git checkout -- tests/fixtures/racket/` clears it; anyone on Windows pulling this merge into an existing checkout will want that.

**Declining the `PARSER_GENERATION` bump was the right call and better than what I had drafted.** I came in expecting to ask for one. `READER_GENERATION` reaches exactly the indexes holding `.rkt` files, where a global bump re-parses every language for everybody to fix a Racket-only change — the #556 argument applied consistently. The catch underneath it is the one I would have missed: the shared parse cache keys on `INDEX_VERSION`, not `PARSER_GENERATION`, so without `:rr<gen>` in the Racket key a file parsed by tree-sitter last week keeps being served after the swap. The test that bumps the constant and asserts the key moves is the part that keeps it true.

Also noted: the fidelity harness counts with `collections.Counter`. Our Rust harness keyed bare names into a set and graded a 37.9% name-collision rate as a perfect run for six days. Yours cannot make that mistake.

The 131-vs-0 measurement on `require` edges is the kind of number that settles an argument before it starts.

PR #570, opened 2026-09-02
fix: tolerate stripped blast-radius cache metadata

rknighton:
## Summary

- tolerate cached `get_blast_radius` results whose `_meta` envelope was stripped by the dispatcher
- add a direct cache-hit regression and a repeated dispatcher-call regression
- preserve the existing tool signature and response contract

## Root cause

`server.call_tool` can strip `_meta` from the same dictionary retained by the result cache when `meta_fields=[]`. The next identical `get_blast_radius` cache hit indexed `cached["_meta"]` and raised `KeyError: '_meta'`.

## Local validation

Exact head: `6ce53a04239f132ed2efcdfad717d5feb1ec5ec1`

- `PYTHONPATH=src py -3 -m pytest tests/test_v1_108_56.py tests/test_server.py tests/test_result_cache.py -q`: 103 passed
- `PYTHONPATH=src py -3 -m pytest tests -q`: 9,133 passed, 14 skipped, 3 failed
- the same three failures reproduce on untouched upstream `main` at `9fb1568c46e81d5c053c53c7589567674fc1290d`: one configured `JCODEMUNCH_PERF_TELEMETRY` warning assertion and two Windows symlink privilege failures
- `py -3 -m ruff check --ignore F401 src/jcodemunch_mcp/tools/get_blast_radius.py tests/test_v1_108_56.py tests/test_server.py`: passed
- `git diff --check`: passed
- cache-envelope microbenchmark, 10,000,000 loops best of 7: direct lookup 37.9 ns; tolerant lookup 45.8 ns; delta 7.9 ns per cache hit

Hosted CI was not queried. The local evidence above is the complete validation record for this submission.

## Risk

JCodeMunch post-edit analysis rated the change medium risk at 0.4203 because `get_blast_radius` is a large, highly connected MCP tool. The patch itself changes one cache-envelope lookup and preserves the signature.

### DIFF
--- a/src/jcodemunch_mcp/tools/get_blast_radius.py
+++ b/src/jcodemunch_mcp/tools/get_blast_radius.py
@@ -375,7 +375,7 @@ def get_blast_radius(
     cached = result_cache_get("get_blast_radius", repo_key, specific_key)
     if cached is not None:
         result = dict(cached)
-        result["_meta"] = {**cached["_meta"],
+        result["_meta"] = {**cached.get("_meta", {}),
                            "timing_ms": round((time.perf_counter() - start) * 1000, 1),
                            "cache_hit": True}
         return _attach_scip_to_blast(result, store, owner, name)
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -603,6 +603,54 @@ async def test_meta_fields_empty_list_removes_meta_envelope():
 
 
 @pytest.mark.asyncio
+async def test_meta_fields_empty_list_preserves_blast_radius_cache(
+    tmp_path, monkeypatch,
+):
+    """Stripping response metadata must not corrupt the cached tool result."""
+    from jcodemunch_mcp import config as config_module
+    from jcodemunch_mcp.tools.index_folder import index_folder
+
+    (tmp_path / "owner.py").write_text(
+        "def target_symbol():\n    return 1\n"
+    )
+    (tmp_path / "consumer.py").write_text(
+        "from owner import target_symbol\n\n"
+        "def consume():\n    return target_symbol()\n"
+    )
+    store_path = str(tmp_path / "idx")
+    repo = index_folder(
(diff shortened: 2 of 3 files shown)


jgravelle:
Thanks — this is a real one and the diff is right. `cached["_meta"]` raises `KeyError` on a cache hit whenever `meta_fields: []` stripped `_meta` before it was cached, which is invisible on any machine carrying a non-default config. All 12 checks pass, and the regression test sits at the dispatcher, where the stripping actually happens.

**The CLA gate is not about your signature.** You've signed: #473 and #474 are merged, both authored as `33261235+rknighton@users.noreply.github.com`. This PR's commit `3c23f07` is authored `Agent <agent@example.com>`, which resolves to no GitHub account — and that is what CLA Assistant is reporting:

> **Agent** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA.

So a recheck won't clear it. Nothing about your agreement changed, and the commit still names an author nobody can attribute. We won't post the status ourselves either: it's an assertion about a legal agreement, and a maintainer-written one is indistinguishable from the genuine article afterwards.

Amending the author fixes it and re-provokes the check as a `synchronize`:

```
git commit --amend --author="rknighton <33261235+rknighton@users.noreply.github.com>" --no-edit
git push --force
```

If that hasn't happened by **2026-09-03 11:30 UTC**, we'll implement the same fix ourselves and credit you in the CHANGELOG, the release notes and the closing comment. The window decides whose commit it is — never whether the fix ships, and never whether you're credited.

rknighton:
Filed #572 with the root cause. Short version: the result cache hands out the object
the dispatcher then edits, which is why `find_references` has this same crash, and why
two quieter failures come from the same writes.

Nothing there blocks this PR. Merge it whenever suits, or close it in favour of the
cache fix if you would rather have one change than two. I have no preference on which,
and either way the guard in this PR ends up unreachable rather than wrong.

PR #571, opened 2026-09-02
markdown: index .md/.markdown headings and fenced code blocks (v1.108.316)

devtomnl:
## Summary

`.md` / `.markdown` files were dropped at discovery as `wrong_extension`. They now index as a first-class language via the bundled tree-sitter `markdown` grammar, following the Svelte precedent for registration (extension map + `MARKDOWN_SPEC` / `LANGUAGE_REGISTRY`, auto-flows into the MCP `language` enum and config validation).

**`heading` symbols carry the document outline.** ATX (`#`..`######`) and setext (underlined) headings, parented by heading level — an h2 nests under the nearest preceding lower-level heading. A heading's byte range covers its whole section (so `get_symbol_source` returns section text, not just the heading line); setext headings sit flat in the grammar, so their extent is cut at the next sibling setext heading. Names are emphasis-stripped (`## **Usage**` → `Usage`), empty headings become `(untitled)`, and same-file duplicates are disambiguated inline (`~2`) so parent ids never dangle. The `docstring` carries the section's first paragraph (skipping siblings *before* the heading — flat setext sections put the previous heading's paragraph there).

**`code_block` symbols** surface fenced code blocks (` ``` ` / `~~~`) named by their info-string language (`python`; `code` when absent), parented under the enclosing heading.

## Behavior notes

- Frontmatter (`minus_metadata`) is skipped; a file with no headings or fences returns no symbols and stays indexed for text search only (the SASS precedent).
- Doc text produces **no import edges** — code snippets inside docs do not leak into the import graph.
- `.mdx` is deliberately not routed: the language pack has no JSX-aware markdown grammar.
- `find_dead_code` skips doc extensions (`.md`, `.markdown`): docs are consumed by humans and agents, never imported by code — without the guard every indexed `.md` reported `zero_importers` at confidence 1.0 with all of its headings listed as dead symbols.

## Drive-by fix

`VALID_KINDS` gains `heading` and `code_block`, and also **`field`** — the `search_symbols` `kind=` gate validates against that set, and `field` has been emitted by the Python parser since the dataclass-fields change without ever being in it, so `search_symbols(kind="field")` errored. The `kind` enum in the `search_symbols` input schema now derives from `VALID_KINDS` instead of a hardcoded list that had already drifted from it (a raw-stdio MCP probe caught that: the runtime accepted `heading`, the wire schema rejected it).

## Testing

- `tests/test_markdown.py` (18): hierarchy, inline duplicate disambiguation, emphasis stripping, empty headings, setext extents, tilde fences, no-info fences, byte ranges/content hashes, frontmatter skip, no import leakage, file summary, discovery routing (`.mdx` stays excluded).
- Intent-preserving updates to `test_tools`, `test_suggest_queries`, `test_v1_108_56` (they codified the old "markdown ignored" behavior) and `test_server` (kind-enum assertion now reads `VALID_KINDS`).
- Full suite green locally modulo the `tiktoken`-tokenizer download, which this network cannot reach (63 fails / 22 setup errors, all `openaipublic.blob.core.windows.net` ConnectionError at fixture setup — the same run shows 0 non-environmental failures); `ruff check src/` clean.

**NO INDEX_VERSION bump** — additive; previously-skipped files just gain symbols on re-index. No new tool: `core_compact` unchanged (the language enum is demoted under compact schemas), full surfaces gain one enum value inside the 5% drift guard.


jgravelle:
Closing this on scope, and I want to be exact about which part: **markdown section extraction belongs to jdocmunch-mcp, not here.** The suite splits along a boundary that predates the PR — jcodemunch owns symbol extraction (functions, classes, types) and file-level summaries; jdocmunch owns prose doc extraction and section search across doc files. `heading` symbols carrying section byte ranges and section-first-paragraph docstrings are the second thing.

That's not a hypothetical overlap. jdocmunch already indexes this repository: 62 `.md` documents, 2,570 sections, with roles assigned. Merging this would give two servers two models of the same markdown, and the first question after that is which one is right when they disagree.

**The quality is not the reason, and I don't want that misread.** This is one of the more careful PRs this repo has received. You hit all five version pin sites, rotated Current State correctly, respected the `NO INDEX_VERSION bump` call, and when the new enum pushed `core_compact` to 4,003 against a hard 4,000 ceiling you fixed it through the existing `_COMPACT_DEMOTE_ENUM_PARAMS` mechanism instead of raising the budget — which is the right instinct and not the obvious one. The `.mdx` exclusion and the "no import edges from doc text" guard both show you thought about where this leaks.

**Your drive-by fix is a real bug and we're taking it.** Confirmed here independently:

- `VALID_KINDS` in `parser/symbols.py` holds seven kinds and `field` is not among them, while this repo's own index carries **399 `field` symbols**.
- `server.py:2114` hardcodes the `kind` enum to the same seven, so an MCP client passing `kind="field"` is rejected at the wire schema while the runtime serves it happily. A schema-derived enum is the right fix, exactly as you had it.

We'll ship that separately and credit you in the CHANGELOG and release notes.

**On the gap you found:** you're right that `.md` being dropped at discovery as `wrong_extension` means a jcodemunch user cannot see their own README. The answer we want is jdocmunch pointed at the same repo, not a second parser here. If that handoff is unobvious in the docs, that's a real problem and worth an issue — it would be a better use of your time than this diff, and I'd rather have your read on it.

jgravelle:
Your `field` finding is fixed and pushed: ef259ce.

`KIND_ORDER` is the one authority now — `VALID_KINDS` derives from it and the schema publishes it, so the wire enum can't drift from what the parser emits again. Kept as an ordered tuple rather than publishing the frozenset directly: `str` hashing is randomised per process, so set iteration order would have served a different schema on every server start for the same build.

Two things your PR made me check that I'm glad you did:

- `test_search_symbols_tool_schema` asserted the literal seven-kind roster, so it could only pass while the bug existed. Rewritten to the property, per the practice that says to check whether a red test was the defect's witness.
- Live `core_compact` measures **3,998 of the hard 4,000**. Our own CLAUDE.md said 3,990 with "roughly ten tokens of slack" — you were right that there is effectively none, and that figure is now corrected.

Credited in the CHANGELOG and it'll be in the next release notes.

PR #578, opened 2026-09-03
ci probe: broken entry point (must block at stage 3)

jgravelle:
pyproject scripts target renamed to a missing symbol. Expect: stages 1-2 PASS (nothing imports the console script), package: install and handshake FAIL on both OS.


jgravelle:
Probe complete; its block is recorded in docs/cicd/VERIFICATION.md §2 with this PR's run links. Branch kept until the series is reported.

PR #581, opened 2026-09-03
ci probe: tool surface changed, no docs (must block at stage 5)

jgravelle:
One tool's registered name changed; README.md, CLAUDE.md and CHANGELOG.md untouched. Expect: done: tool surface documented FAIL naming the added/removed tool. The fast tier will also fail (surface pins), which is why the done-* jobs no longer depend on it.


jgravelle:
Probe complete; its block is recorded in docs/cicd/VERIFICATION.md §2 with this PR's run links. Branch kept until the series is reported.

PR #584, opened 2026-09-04
ci: release workflow (dry-run by default; trusted publishing; no real publish this session)

jgravelle:
Phase 3 step 5 of docs/cicd/DESIGN.md, stacked on #583. Dispatch with version + dry_run=true builds, twine-checks, smokes the built wheel on both OSes, renders the notes, and stops before any upload, tag, release or registry write. A hand-pushed v* tag runs the pre-flight and fails it. Trusted publishers must be registered on PyPI and Test PyPI before dry_run=false (docs/cicd/RUNBOOK.md).


jgravelle:
CodeQL's two findings on release.yml: the unquoted colon (invalid YAML) is fixed and every workflow is now parsed by tests/test_workflows_pinned.py; the untrusted-checkout alert is dismissed with a reason (dispatch-only, repo refs; the tag path fails before checkout). docs/cicd/FINDINGS.md C-10.

PR #614, opened 2026-09-05
competitive: the fourth competitor adapter, an LSP-backed tool over MCP stdio (Phase 3 item 2, fourth of eight)

jgravelle:
Phase 3, item 2, fourth adapter: the LSP-backed tool over MCP stdio, at the version pinned in `sandbox/serena.requirements.txt`, with pyright 1.1.403 as its language server. Stacked on `competitive/adapter-code-review-graph`; merge in order, or retarget after.

## What changed

- `sandbox/serena.Dockerfile` + `sandbox/serena.requirements.txt` (+ `.in`): the wheel (sha256 `6dbf1459…`), `pyright==1.1.403` (the version the tool pins) and every dependency by hash, `--require-hashes`. pyright's Python wrapper fetches Node at first use, so the image fetches it at build into `/opt`, world-readable. The wrapper's pinned Node-version variable is unset for the run: set, it re-runs the installer on every start, and that download killed the language server under `--network none` in probe 1.
- `sandbox/serena_config.yml`: the template's values with `web_dashboard: False`, `gui_log_window: False`, `ls_specific_settings.python.ls_path` pointing at the image's `pyright-langserver` (the documented bypass of `uvx`), and `project_serena_folder_location` under `/out` because the default is inside the read-only corpus.
- `adapters/serena.py`: two sessions per run. Session 1: initialize, `tools/list`, one uncharged `get_current_config` (its wall time plus the initialize round trip is the index time: the language server's start and first analysis), then `find_symbol(name)` for P1/P2 and `search_for_pattern(words joined by |, code files only)` for T, all charged. Session 2: `find_symbol(include_body=true)` for each of the first three P1 matches and `find_referencing_symbols(name_path, relative_path)` for each exact-name P2 match, charged, after that session's own uncharged `get_current_config` (the server and the language server start again in session 2; review round 1 found that start charged to the first body read). Citations from the tool's JSON (`relative_path`, `body_location`, LSP 0-based lines cited 1-based); a P2 answer cites the references only. P4 is not answered: the README's retrieval table has no importers row, so the row is NOT COMPARABLE, never a zero.
- `sandbox/mcp_driver.py`: both pipes are read on threads. stderr is drained continuously into a bounded tail (probe 2 hung on every call after the first: the tool logs each result to stderr at INFO and the driver read stderr only at exit); stdout goes through a queue so a deadline read works where `select()` has no pipes, which is what lets `tests/test_competitive_mcp_driver.py` run on Windows. That test's stub server writes 4 MB to stderr before answering; it fails against the pre-change driver (`evidence/red.txt`) and its `drain_stderr=False` arm must time out. A harness fix that can only unblock a row, never improve one; the two earlier MCP rows are re-measured with it in this PR's run and their pinned image digests move with it.
- `docs/competitive/fairness/serena.md` was written before the first number; its last section records the three probes.
- `tests/test_competitive_serena.py` (8) over a fixture captured from v1.7.0 in the sandbox.

## Result

The recorded rows are `results/2026-09-05-76e75398.json` (three runs, seven adapters, container), rendered in `results/latest.md`; what each is evidence of, and its caveats, are FINDINGS CF-27 to CF-30. This body does not restate them. The two earlier MCP rows were re-measured in the same run with the changed driver; CF-27 names their medians and the one row whose runs did not all agree.

## Review

Independent reviewer (`.claude/agents/reviewer.md`, fresh context), three rounds, APPROVE at 89be2fdc. Round 1 REQUEST CHANGES: no test failed against the pre-change driver; a result described as "repeated exactly" when one run differed; a KB range and a probe figure the result file did not support; the fairness note's launch line disagreed with the adapter; the second session's server restart was charged to the first body read. Round 2 REQUEST CHANGES: the regenerated CF-28 published a JSON string's character count as a tool count; the warm-call claim needed to say what the files show; the non-vacuity pass was a sentence, not an evidence file. Every item resolved to a file or a line in round 3 (`evidence/review.md`). Two notes stand open and non-blocking: the tool's config template cannot be verified from the tree, and the generated summary cannot say what an index second is for a language-server tool.

## Evidence

- fast and full tier on the committed tree: fast PASS, full PASS at 89be2fdc (9771 passed, 19 skipped; `evidence/full.md`)
- bench tier `--offline`: PASS at 89be2fdc, every block (`evidence/bench.md`)
- checklist against `origin/main`: 12 of 12 met or n.a. at 89be2fdc (`evidence/checklist.md`); red.txt runs the new driver test against the pre-change driver at 4087ee31, green.txt at HEAD (the drain_stderr=False arm times out, which is the pre-change driver's behaviour reproduced on any box)


jgravelle:

PR #615, opened 2026-09-06
competitive: corpus and task fairness checks, the pinned corpus set, tasks for nine corpora (Phase 3 item 3)

jgravelle:
Phase 3, item 3: the corpus and task fairness checks as scripts that run before scoring, the pinned corpus set they needed to pass, and line-level tasks for nine corpora. Stacked on `competitive/adapter-cocoindex`; merge in order, or retarget after.

## What changed

- `benchmarks/competitive/corpus_check.py` + `corpus_policy.json`: the set of corpora is judged before anything is scored, on five criteria (language count, a file over a line floor, a corpus over a file floor, more than one domain, no language over a share cap), every threshold read from the policy file and the verdict written into every result header. A failing set stops the run (exit 5). The three corpora pinned by `benchmarks/tasks.json` plus this tree's own `src/` fail two of the five, not the four `DESIGN.md` s3.4 predicted (FINDINGS CF-46).
- `corpora.json` + `corpora.py`: five additional corpora pinned by full SHA (a utility library, an HTTP client library, a TypeScript monorepo, and two repositories over 10,000 files in different languages; one alone puts its language over the cap, which is why there are two), fetched by SHA into a cache OUTSIDE the tree because `benchmarks/` ships in the sdist.
- `task_check.py`: the answerability check moved out of `run.py`, the symmetric capability-only split, and a post-run `tools_not_called` list (a tool that cited nothing on every P task of a corpus).
- `tasks/from_sverklo.py`: a third party's task rules (sverklo-bench, CC-BY-4.0, pinned by commit) reproduced in Python over our checkouts for three corpora; every hand-verified definition line is re-verified at our SHA (a moved line refuses the generator), and the three tasks that do not exist at our SHAs are dropped with the reason (CF-48). `tasks/authored.py`: for the other three corpora, symbols chosen by hand, expected sets computed by the same rules, never typed. Nine task files; `run.py` reads a `--tasks` directory, `--set`, `--only`, `--corpus ID=PATH|DOMAIN`.
- One author wrote every adapter and every task; DESIGN s4.2's independence rule is not met by one agent and is recorded, not softened (CF-47). The reviewer was told.
- Found by the first full-set run and fixed in the harness: a sandbox timeout killed the docker client and not the container (CF-49; containers are named and killed, with a test that leaks one against the pre-fix code), and a checkpoint of finished runs is written per run. `--set none` (a smoke run) records the corpus check without enforcing it and `--record` refuses it.
- `tests/test_competitive_checks.py` (19, two of them the CF-49 kill); ARCHAEOLOGY row; CHANGELOG "Added" entry (no competitor named, no figure).

## Result

The recorded rows are `results/2026-09-06-0e3a1706.json` (self corpus + the two third-party corpora, eleven adapters, three runs, container), rendered in `results/latest.md`; what they are evidence of, and the two things the run found that matter more than its rows (our own adapter's P2 mapping, CF-51; a competitor image short of its documented configuration, CF-52), are FINDINGS CF-50 to CF-53. The whole set does not fit DESIGN s9.2's budget on a workstation (CF-53); the schedule item is sized from that. This body does not restate the rows.

## Review

Independent reviewer subagent, two rounds. Round 1 REQUEST CHANGES: a FINDINGS row counted tools the same result file flags `tools_not_called` as wins; another carried a figure from no recorded file; the red evidence was a collection error and the container-leak test had skipped without a daemon; typed literals in a generator. All four fixed at 534f557c: rows regenerated from two recorded files (one new, `docs/competitive/evidence/corpus_check_old_set.json`), the leak reproduced against the pre-fix sandbox with a daemon (`red.txt`), the outcome test run green. Round 2 APPROVE: "no Floor moved; no test retired or skip-marked; no competitor named in CHANGELOG.md"; every number it spot-checked matched the files. Open by design and disclosed: CF-47, CF-51, CF-52, CF-53. Verbatim in `.claude/state/evidence/review.md`.

## Evidence

- fast tier PASS on every commit (hook); full tier PASS at 534f557c (HEAD), `9818 passed, 20 skipped` (`.claude/state/evidence/full.md`); red.txt is the container leak against the pre-fix sandbox with a daemon (EXIT=1) plus the new test file's collection error against the base tree; green.txt is the three competitive test files, 36 passed, 0 skipped
- bench tier `--offline`: PASS (`bench.md`)
- checklist against `origin/main`: 12/12 met or n.a., unmet none (`checklist.md`)


jgravelle:

PR #621, opened 2026-09-06
competitive: Phase 5, the instructions (CLAUDE.md section, RUNBOOK section 10, ENFORCEMENT-PLAN)

jgravelle:
Phase 5 of the competitive layer: the instructions. Stacked on `competitive/phase4-verify` (Phase 4); merge in order, or retarget after. Docs only.

## What changed

- `CLAUDE.md`: a "Competitive" section beside the Inbound and CI/CD ones, in their shape: where the loop, the set, the verification and the human's part live; that every number is script-written from a result file and a typed number is a review finding; that a competitor's README figure is never a measurement and competitor code runs only in the sandbox; that losses are recorded unsoftened, naming our own open P2 defect; that the three jobs are OFF until a human acts; that nothing touches marketing. Measured against the file's budget by the harness's own check (the fast tier's `claude_md.max_chars` row).
- `docs/cicd/RUNBOOK.md` section 10, the human's part of the loop: the second variable and its exact-string rule; the four labels to create, with descriptions; running the monthly job by hand with a reason and an optional tool, and why the first dispatched run is the measurement two findings wait for; approving a draft on the ledger branch; reading a gap draft, including the two caveats that must travel with a token row and that a `tool_not_called` hypothesis is the adapter's fault first; the weekly feed; what to do when something looks wrong.
- `docs/standard/ENFORCEMENT-PLAN.md`: a dated status paragraph for the layer in the shape of the workflows and inbound ones (no item moves by it; no Floor added, no threshold edited; a `standard-proposal` can only ask a human to change a Target), and the out-of-scope line that parked head-to-head benchmarking on a VM now says it was built, in a container.
- `docs/inbound/POLICY.md` needed no edit: 4.4's label exception for the post job was amended in #618 and the prompts re-rendered there.

## Result

Docs only. The instructions point at things that exist: the variable name the post job reads, the four label names `post.py` accepts, the workflow file names and their inputs, the ledger paths, the artifact names, the container name prefix.

## Review

[...]

PR #622, opened 2026-09-07
workflows: a commit that stages a file a Floor reads runs the fast tier (W-39)

jgravelle:
Workflows follow-up W-39, found by the reviewer of #621: a docs-only commit that edits `CLAUDE.md` moves a Floor and the pre-commit hook was not measuring it. Stacked on `competitive/phase5-instructions` (#621); merge in order, or retarget after.

## What changed

- `.claude/hooks/pre_commit.py`: a commit whose staged content reaches a code root OR a file a Floor's Method reads runs the fast tier; every other docs-only commit stays free. The trigger list (`FLOOR_INPUTS`) is `CLAUDE.md` plus the frozen benchmark artifacts the fidelity, schema and goldset Floors read; the rule is a named predicate (`tier_needed`), with the reason in the header. Before this, "docs commits are free" let a PR that edited only `CLAUDE.md` reach the reviewer with a `fast.md` from a tree two commits older, while the one Floor the diff moved (`claude_md.max_chars`) went unmeasured on HEAD; it passed when run by hand, which is luck, not process.
- `tests/test_workflow_hooks.py`: the predicate tested directly (driving the hook would run the whole tier), and a ratchet over the property: it parses the path literals `harness/__main__.py` reads from the tree and fails when the hook does not trigger on one, so the next Floor input added to the harness cannot become a free commit. The first draft named `CLAUDE.md` alone; the reviewer's round 1 found the benchmark artifacts to be the same class one directory over, and red.txt is the ratchet against that draft (fails, naming them); green.txt is the two hook test files.
- `docs/workflows/FINDINGS.md` W-39, which also records the second half the reviewer found: a by-hand `harness fast --summary` APPENDS to `fast.md` (the hook unlinks first), so two PASS blocks with two different sizes sat in one evidence file. That half is a habit until the summary flag gains a truncate mode: unlink first, or stage `CLAUDE.md` and let the hook run.

## Review

Independent reviewer subagent, two rounds. Round 1 REQUEST CHANGES: the trigger named one spelling of the property (`CLAUDE.md` alone) while the frozen benchmark artifacts the fidelity, schema and goldset Floors read were the same class one directory over, and DESIGN §4's H1 row still stated the pre-fix rule. Both taken at ece3ae8a: the trigger is every file the harness reads for a Floor, a ratchet over the harness's path literals fails on an uncovered one (run red against the first draft), and the DESIGN row describes the rule as committed. Round 2 APPROVE: "both round-1 change requests are closed; no Floor FAIL, no threshold change, no retired test, no CI or release-path change". Two residues it placed in the next touch, not this PR: the DESIGN row's path list has no binding test; and red.txt, which was of a sibling draft, has been regenerated from the committed test after the verdict (evidence only). Verdicts verbatim in `.claude/state/evidence/review.md`.

## Evidence

- fast tier PASS on each commit (the hook, since `tests/` was staged); full tier PASS at ece3ae8a (`.claude/state/evidence/full.md`); red.txt EXIT=1 (the ratchet against the round-0 hook); green.txt 81 passed over `tests/test_workflow_hooks.py` and `tests/test_workflows_registered.py`; checklist 12/12 met or n.a. against `competitive/phase5-instructions`


jgravelle:

PR #633, opened 2026-09-07
harness: weekly bench result (2026-09-07)

jgravelle:
Weekly harness results from run 34158092318 on 2937e8df. Opened by hand from the branch main.yml pushed; the job could not open it (cicd FINDINGS C-17: the repository forbids GITHUB_TOKEN from creating pull requests).


jgravelle:
Closing: the job committed this as harness-bot@users.noreply.github.com, which GitHub resolves to a real account that is not ours, so the commit is attributed to a stranger and CLA Assistant reports it unsigned (cicd FINDINGS C-17, third cause). The same two result files are re-authored as github-actions[bot] on a fresh branch and opened as a new PR; the workflow fix is in #634.

PR #669, opened 2026-09-12
docs: classify_intent claims a fallback it never ran, and a transcribed headroom went stale

jgravelle:
## What was wrong

Three claims in `classify_intent`'s docstring, none of them true. It said the
function "combines the curated intent rules (high precision) with a
catalog-search fallback (high recall), de-duplicated, primary first." The body
walks `_INTENT_RULES` and returns; there is no fallback, so nothing is
combined, nothing is de-duplicated against a second source, and "primary first"
names an ordering between sources that never both exist. The lexical fallback
lives in `_handle_route` in `server.py` and fires only when this function
returns nothing.

Separately, CLAUDE.md's `core_compact` entry transcribed a headroom figure that
went stale. Both were found while assessing an unrelated design proposal; no
user reported either.

## Why

The docstring is a second copy of a fact whose owner is elsewhere, which is the
shape the Standing lessons keep naming -- a second generator, a second call
site, a second derivation. Here the consequence is specific: a reader who
believes the docstring thinks an action ranked by neither channel was scored
and lost, when in fact an action a rule preempted was **never scored at all**.
That is the `rule_preempted` miss class, which the module comment above
`_INTENT_RULES` already documents correctly -- so the file disagreed with
itself.

The CLAUDE.md half is the project's own rule firing on the project: never
transcribe a fact with an expiry date, name where to read it. The entry was
stale in the alarming direction, claiming the ceiling had two tokens of room
when it has thirty-three, which is the sort of wrong that gets a real warning
ignored.

Not a regression of an ARCHAEOLOGY row; no test covered either.

## What is now impossible

A reader cannot take `classify_intent` for a two-channel ranker. The docstring
names the single channel, states the empty-list case, and points at the
function that owns the fallback -- by name, not by line number, so the pointer
does not rot.

The CLAUDE.md entry can no longer carry a stale number, because it carries no
number: the headroom is named as a location. Checked the other spelling of the
same mistake while in there -- `claude_md.max_chars` has the identical shape
(a `target` beside a `floor`, only the floor gated) and is also breached, so
F-27 records the class rather than the instance, and warns against the obvious
wrong fix of promoting a target to a floor.

No behaviour change. The published tool surface is byte-identical.

## Evidence

Full tier, this tree:

    10256 passed, 24 skipped, 21 warnings in 326.43s (0:05:26)
    coverage.min            crit N2  floor >= 74      observed 82        PASS
    ci.skips_windows        crit N7  floor <= 25      observed 24        PASS
    suite.full_seconds      crit N1  floor <= 360     observed 327.89    PASS
    HARNESS PASS
    full-tier stamp: ok=True tree=9df4f8b00e61

Fast tier:

    counter.saving_min      crit 4   floor >= 0.95    observed 0.9604    PASS
    route.control_at1       crit 4   floor >= 40.0    observed 40.0      PASS
    claude_md.max_chars     crit N6  floor <= 140000  observed 139978    PASS
    suite.fast_seconds      crit N1  floor <= 90      observed 65.54     PASS
    HARNESS PASS

`ruff check src/` clean. The three Floors above are the ones this diff could
have moved: the counter saving and route recall prove the surface and router
are untouched, and `claude_md.max_chars` is the one the CLAUDE.md edit moves
directly. The live artifact reads `core_compact: 3967`; the removed literal
said 3,998.

## Definition of Done

| DoD | item | verdict | evidence |
|---|---|---|---|
| 1 | A test that fails on the pre-change tree and passes on the post-change tree | met | evidence/red.txt fails=True; evidence/green.txt passes=True |
| 2 | `ruff check src/` clean and the touched test files green locally BEFORE the full suite | met | fast tier pass=True; full tier pass=True |
| 3 | For a change under `src/`: a `CHANGELOG.md` entry under `[Unreleased]` | n.a. | label no-changelog |
| 4 | If a tool was added or its description changed: README, CLAUDE.md Key Files, ... | n.a. | surface unchanged (scripts/surface_diff.py) |
| 5 | If a benchmark artifact moved: every mirror re-synced | n.a. | benchmarks/ untouched |
| 6 | If a config key, env var or CLI subcommand was added: a row in CONFIGURATION.md or CLI-AND-ENV.md | n.a. | config.py and cli/ untouched |
| 7 | If a new background, persistent or network behaviour was added: a README disclosure | n.a. | no thread/socket/http/scheduled addition in the diff |
| 8 | If a number is published in a response: a `*_basis` field | n.a. | no new rate/share/confidence key in the diff |
| 9 | For a contributor PR: trial-merged onto `main` | n.a. | our own PR |
| 10 | `python -m harness fast` passes | met | fast pass=True; bench not required |
| 11 | If a test was retired: a `harness/retired.json` entry | n.a. | no test file deleted |
| 12 | If a threshold moved: tightened with history, or loosened with a block | n.a. | harness/thresholds.json untouched |

Checklist: 12/12 met or n.a.; unmet: none.

⚠ **Read row 1 with a caveat, and it is a finding rather than an excuse.** This
diff changes no behaviour, so there is no red/green pair to produce, and the
row reads `met` because `dod_checklist.py` resolves `evidence/red.txt` and
`evidence/green.txt` with **no tree binding** -- so it inherited the artifacts
of the previous change on this box (07:26 and 07:44 today, from #666). The
stamp mechanism immediately next to it, `.claude/state/full-tier.json`, is
tree-keyed and refused this PR until the tier was re-run on tree
`9df4f8b00e61`, which is exactly the right behaviour and exactly what row 1
lacks. Filing separately; not fixed here, because a checklist change belongs in
its own diff and this one is meant to be boring.


jgravelle:
Second commit adds `F-28` to `docs/harness/FINDINGS.md`, recording a CI flake that happened **on this PR**: `test_registration_race_updates_persisted_symbols[live_tree-polling]` failed on `full: test (windows-latest, 3.12)` (run 34717936085) while the other seven matrix jobs passed and the full tier passed twice locally on this tree.

Recorded rather than re-run silently, because this is the **second** occurrence of that test — the first was windows 3.11 on 2026-09-11 (run 34644461374), and that entry set the rule that a second occurrence earns a FINDINGS row.

The diff here is a docstring and two docs files, nothing within reach of the watcher, so this is the flake and not the change. The assertion survived this time because CI keeps what `run_full.py` drops, and it names a real interleaving rather than a slow runner: the polling watcher saw the arm-time write and both pre-existing files and never saw the post-recovery one, so re-registration and the write raced. The row says pin the interleaving and explicitly says not to raise the timeout, which would hide the race the test exists to find.

Also removed the `agent-authored` label I put on this PR by mistake. It is the inbound layer's marker for bot PRs, so it invoked `inbound-selfcheck`, which correctly failed all six of its clauses — including `author 'jgravelle' is not the App 'jcodemunch-inbound[bot]'`. That check is not a required context and never blocked the merge. Unrelated to the code here.

PR #674, opened 2026-09-13
docs: point at jFlowMunch, and pay for it by de-transcribing the tool count

jgravelle:
## What was missing

CLAUDE.md said nothing about jFlowMunch. It now carries one instruction in the
work-practices section: a proposal touching more than one control-flow step
writes `.jFlowMunch/current.json` before approval is asked, with the schema left
where it lives (`C:\MCPs\jflowmunch\agent\jflowmunch-instructions.md`) rather
than copied in.

## Why it is a trade and not an addition

**CLAUDE.md had six characters of headroom** against its 140,000 floor —
`claude_md.max_chars`, the one threshold whose `loosened` block reads *"the 10k
buffer is the last one. Never raise again."* So the pointer had to be paid for.

Three things it is explicitly **not** paid for with, because CLAUDE.md forbids
each: raising `BUDGET`, deleting a ⚠⚠ rule, or performing the Key-Files split
the file says is jjg's call.

It is paid for by the `Tool count` line, which read:

> 91 visible in `full` / 94 in catalog (front door hidden; counts verified
> 2026-07-30 ... +1 v1.108.111 `get_parity_map`, +1 v1.108.112
> `get_decorator_census`, +1 v1.108.113 `get_architecture_metrics`)

That is a number with an expiry date plus a per-version log of which release
added which tool — a literal that had been stale for six weeks, and history git
already holds. It is the same shape as F-27, one section over, and F-27 was
filed earlier today about exactly this.

⚠ **The rule the line exists to state survives verbatim**: read it from
`jcodemunch-mcp surface`, never type it here. Only the transcribed numbers and
the changelog left.

## Evidence

    main    139,994 chars   headroom 6
    here    139,974 chars   headroom 26

The file is **smaller than it started** with the new instruction in it.

    tests/test_claude_md_size.py
    tests/test_claude_md_rotation.py
    tests/test_key_files_split.py
    tests/test_cli_env_split.py
    32 passed in 12.00s

Full tier, this tree:

    10256 passed, 24 skipped, 75 warnings in 232.96s (0:03:52)
    suite.full_seconds   crit N1  floor <= 360  observed 234.22  PASS
    HARNESS PASS
    full-tier stamp: ok=True tree=69594fb9a0ea

⚠ The four CLAUDE.md gates are run and quoted separately above because the
full-tier stamp's `TIER_PATHS` deliberately exclude docs — `tree_id`'s own
docstring names this residual ("a doc edit CAN flip a doc-reading test
(CLAUDE.md size); the PR gate is the authority"). The stamp being valid is
therefore not evidence about this diff, and the four tests are.

## Note for review

This is the second CLAUDE.md de-transcription today; #669 does the same to the
`core_compact` headroom. Both are the same finding: a literal with an expiry
date, sitting in the one file that cannot afford to carry it. Nothing about this
PR depends on #669, and the two touch different sections.


jgravelle:
Second commit ignores `.jFlowMunch/`.

Writing the first flow document under the instruction this PR adds produced `?? .jFlowMunch/` at the repo root — an untracked scratch file, in the repository whose v0.2.6 release shipped credentials out of exactly that position, and whose `relnotes.md` incident weeks later was the same shape again. `git add -A` does not ask.

The file is current state by its own definition: overwritten whenever a plan changes, one document at a time, never appended. So it is nobody's deliverable.

Ignored in the same PR that mandates it rather than after someone commits one, so the rule and the artifact it creates arrive together. `tests/test_sdist_exclusions.py` + `tests/test_build.py`: 17 passed.

jgravelle:
Closing unmerged.

Both changes here exist only because the tool reached this repository through the project tree. It no longer does, so the instruction describes a mechanism that is not there, and the ignore rule covers a directory that is never created here. Neither is wrong; both are moot.

Nothing from this branch goes to main.

Issue #745, opened 2026-09-17
The spec-map guard checks a name_fields entry EXISTS, never that it resolves


jgravelle:
## What is wrong

`tests/test_language_spec_maps_agree.py` checks that a declared node type has a
`name_fields` entry. It does not check that the entry names a field the grammar
actually sets, so a spec can declare a symbol, map it to a kind, give it a name
field, produce nothing, and stay green.

PHP is a confirmed instance (filed separately):

```
php symbol_node_types['property_declaration'] = 'property'
php name_fields['property_declaration']       = 'name'
child_by_field_name('name')                   -> None
actual named children: ['visibility_modifier', 'property_element']
```

Every assertion in the guard passes. No PHP class property has ever been
indexed.

## Why this is its own issue

Fixing PHP fixes PHP. Fixing the guard finds every other instance — and the
standing lesson is explicit that the ratchet is the half worth having:

> **Write the ratchet before concluding the reported list is the list.** #489
> reported three sites; a test over the PROPERTY found five.

The guard was written for #712, where a node type was listed in one map and
absent from another. It answers "is this node type nameable in principle?" The
question it cannot answer is "does the name actually come out?", and those differ
by exactly one indirection — which is where #698, #712, #722 and #732 all lived.

## Why a static scan cannot close it

Measured while writing this: the compiled grammar's symbol table
(`Language.node_kind_count` / `node_kind_for_id`) enumerates node KINDS, so it
can prove a declared node type is never emitted — that is #724's property A, and
it reports clean across all 118 declared `(node_type, field)` pairs. It cannot
enumerate which FIELDS a grammar sets on a node, so the PHP case is invisible to
it. **The number of other instances is therefore unmeasured**, and that is the
reason to build the behavioural check rather than an estimate of how bad it is.

## Shape of the fix

Per language, per declared node type, parse a sample and assert a symbol comes
out with the expected kind — the same behavioural form as
`test_every_declared_field_pattern_actually_yields_a_field` (#735) and
`_CLASS_SCOPED_SAMPLES` (#732), both of which replaced a structural check that
could pass over a hole. A declared node type with no sample should fail by name,
so the samples cannot silently fall behind the specs.

⚠ Expect this to fail on arrival for more than PHP. That is the point of
building it, and each failure is a triage decision (declare the sample, or fix
the spec), not a reason to weaken the check.

Related: #712, #724, #732, #735.

Issue #744, opened 2026-09-17
PHP class constants are not indexed; file-scope constants are


jgravelle:
## What is wrong

A PHP class constant yields no symbol. A file-scope `const` in the same file
does.

```
parse_file('<?php\nclass C {\n  const K = 3;\n}\nconst TOP = 4;\n', 'a.php', 'php')
-> [('C', 'class'), ('TOP', 'constant')]
```

`K` is absent. A class constant is the idiomatic way to write a constant in PHP,
so the shape that is missing is the common one.

## Why

The node type is right and the SCOPE GATE is what refuses it.
`PHP_SPEC.constant_patterns` names `const_declaration`, and the grammar spells a
class constant with that same node type inside `declaration_list`. But
`_walk_tree` gates the constant channel on `parent_symbol is None` unless the
language is in `_CLASS_SCOPED_CONSTANT_LANGUAGES` — and that set is
`{"java", "kotlin"}`. Inside a class `parent_symbol` is `C`, so the channel never
runs.

This is exactly the hole #428 opened that set for in Java, and that #732 closed
for Kotlin:

> The constant channel is gated on `parent_symbol is None` unless the language is
> in that set, so at class or object scope it never ran for Kotlin.

PHP is the third language with the same shape and was not considered either
time.

## The fix is one set entry plus its sample, and the set's own rule says so

`_CLASS_SCOPED_CONSTANT_LANGUAGES` is deliberately a named set rather than a
general `parent_is_container` relaxation, because widening it for every language
would move symbol counts in every index. Adding `php` requires a sample proving
what membership buys, which
`test_only_named_languages_reach_constants_through_a_container` now enforces
behaviourally (#732) — a language added without one fails by name.

⚠ Worth checking in the same fix: whether an `enum` case and an `interface`
constant are the same node type or need their own entry, and that a class
constant comes out qualified (`C.K`) rather than bare — `_constant_symbol`
hardcodes `qualified_name = name` and only Rust qualifies at the call site.

## How it was found

Answering the "what other spellings of the same input exist" question while
fixing #735. Verified against `origin/main` in a clean worktree.

Related: #428, #732, #735.

Issue #743, opened 2026-09-17
PHP class properties are not indexed: name_fields names a field the grammar never sets


jgravelle:
## What is wrong

No PHP class property yields a symbol, in any visibility.

```
parse_file('<?php\nclass C {\n  public $prop = 1;\n  private $hidden;\n  public static $shared = 2;\n  public function m() {}\n}\n', 'a.php', 'php')
-> [('C', 'class'), ('m', 'method')]
```

`prop`, `hidden` and `shared` are all absent. A PHP class indexes with its
methods and none of its state — #735's symptom in a second language.

## Why — this is #712's shape, not #698's, and that is the interesting part

`PHP_SPEC.symbol_node_types` **does** name the node type:

```
php symbol_node_types['property_declaration'] = 'property'
php name_fields['property_declaration']       = 'name'
```

So the form is declared, mapped to a kind, and given a name field — and yields
nothing, because **the grammar never sets a `name` field on that node**:

```
child_by_field_name('name') -> None
actual named children: ['visibility_modifier', 'property_element']
```

The name is two levels down, at `property_element > variable_name`, the same
shape as Kotlin's `variable_declaration > simple_identifier` in #732. A
`name_fields` entry pointing at a field the grammar does not produce resolves to
nothing, `_extract_name` returns None, and the symbol is dropped in silence.

Note also that one `property_declaration` can bind several names
(`public $a = 1, $b = 2;`), which `_extract_name` cannot express — see #735 for
the channel that handles N-name declarations.

## Why nothing caught it

`tests/test_language_spec_maps_agree.py` requires every node type in
`symbol_node_types` to be *nameable*: a `name_fields` entry, or an
`_extract_name` branch. PHP has the entry, so the guard passes. It checks that
an entry **exists**, never that it **resolves**. That gap is filed separately.

This is also why `property` sat in `KIND_ORDER` as a declared-and-dead kind for
years, which #732 recorded when Kotlin became its first live emitter:

> `PHP_SPEC` had mapped the kind for years while nothing emitted one, so it sat
> declared-and-dead and no test could see it.

That entry named the symptom. This is the cause.

## How it was found

Answering the "what other spellings of the same input exist" question while
fixing #735. Verified against `origin/main` in a clean worktree.

Related: #712, #732, #735, #571.

Issue #742, opened 2026-09-17
JS/TS `var` declarations are not indexed at all


jgravelle:
## What is wrong

A `var` declaration yields no symbol at all.

```
parse_file('const MAX = 1;\nlet counter = 0;\nvar legacy = 2;\n', 'a.js', 'javascript')
-> [('MAX', 'constant'), ('counter', 'constant')]
```

`legacy` is absent. Same in `typescript` and `tsx`.

## Why

`var` is a different node type from `const`/`let`. The grammar spells it
`variable_declaration`; `const` and `let` are `lexical_declaration`.
`JAVASCRIPT_SPEC` names `lexical_declaration` in `constant_patterns` and names
`variable_declaration` in nothing — not `symbol_node_types`, not
`constant_patterns`. `TYPESCRIPT_SPEC` and `TSX_SPEC` are the same.

This is #698's shape: the grammar spells one concept as two node types, the spec
names one, and the spec looks complete.

## Why it matters

`var` is still what a large amount of existing JavaScript is written in — every
pre-ES6 file, every transpiler output that targets ES5, and most snippets in
older documentation. A module written that way indexes with no top-level
bindings at all.

## How it was found

Answering the "what other spellings of the same input exist" question while
fixing #735. Verified against `origin/main` in a clean worktree, not on the #735
branch.

## Worth deciding as part of the fix

What kind a `var` should carry. It is mutable, so `constant` is wrong for the
reason the sibling issue gives. `property` is the kind #732 added for exactly
this "mutable named binding" case in Kotlin, and `field` (#571) exists too —
whichever is chosen, a kind absent from `KIND_ORDER` is refused by
`search_symbols` and omitted from the wire enum, so the choice is a wire
decision as well as a naming one.

Related: #698 (the same class in TypeScript), #732, #571, #735.

PR #746, opened 2026-09-17
fix: every Java field is a symbol, not only the `static final` ones (#735)

### LINKED ISSUE #735: A plain Java instance field is not indexed
## What is wrong

An instance field yields no symbol. A `static final` field in the same class
does.

```
parse_file('public class A {\n  private int field;\n  public static final int CONST = 1;\n  void m() {}\n}\n', 'A.java', 'java')
-> [('A', 'class'), ('CONST', 'constant'), ('m', 'method')]
```

`field` is absent.

## Why

`JAVA_SPEC.symbol_node_types` does not name `field_declaration`. `CONST`
extracts through `constant_patterns`, a different channel, which is why the gap
looks like it is about `final` rather than about the node type.

This is the widest of the nine gaps #724 found by volume: every Java class with
state indexes without it.

## How it was found

`tests/test_grammar_spelled_forms.py` (#724) compares what each compiled grammar
spells against what the language's node-type map names. `field_declaration` is in
`tests/fixtures/grammar_declaration_inventory.json` and in that file's
`_CONFIRMED_GAPS`, which records the gaps verified by running the product rather
than by reading the scan.

The `_CONFIRMED_GAPS` entry asserts this node type stays in the inventory. When
this issue is fixed the entry leaves and `test_a_confirmed_gap_is_in_the_inventory`
names the line to delete, so the record cannot outlive the defect.

Related: #698, #712, #713, #722 - the same class in four other languages.

jgravelle:
Closes #735.

A Java class indexed with its methods and none of its state. `private int
balance`, a package-private `String owner`, a `static int instances` and a
`final Logger log` all yielded no symbol, while `public static final int
MAX_RETRIES` in the same class did. The widest of the nine gaps #724's grammar
inventory found, by volume — every Java class with state had it, for the whole
life of the spec.

## The gap reads as being about `final` and is about the node type

`MAX_RETRIES` reaches the index through `constant_patterns`, which matches
`field_declaration` and requires **both** `static` and `final` (#428).
`JAVA_SPEC.symbol_node_types` never named that node type at all, so everything
the constant channel declined — which is most fields — had no channel to fall to.
A reader who finds the `field_declaration` branch in the extractor sees coverage
and stops, exactly as in #732.

## Why this is a channel and not a map entry

`symbol_node_types` **structurally cannot express this form**: `_extract_symbol`
returns one `Optional[Symbol]` per node, while `int a, b, c;` is one node and
three declarations. The constant channel has bound every declarator since #428,
so a field channel naming only the first would have made the discriminator
between *indexed* and *silently dropped* the presence of `static final` — #732's
capitalisation incoherence in a new costume.

So: `LanguageSpec.field_patterns` + `_extract_fields`, mirroring
`constant_patterns` / `_extract_constants`, rather than a `language == "java"`
branch in `_walk_tree`. **#731 (Go's package-level `var`) is the same shape and
inherits the channel** instead of growing a second copy of the rule.

⚠ A third node-type list beside two write-only ones is a risk, not a neutral
addition. `type_patterns` and `return_type_fields` are declared by 19 and 14 of
the 79 specs and read by **nothing** (#725), and a list no channel consults is
indistinguishable from the defect it was added to fix.
`test_every_declared_field_pattern_actually_yields_a_field` asserts the
readership through the product, keyed on the spec so the second member is checked
on arrival rather than joining unwatched.

## One predicate, both channels ask it

`field_declaration` is now in `constant_patterns` **and** `field_patterns`, and
`_walk_tree` runs the two independently on the same node — not as an `elif` — so
`static final int MAX` emits twice unless something decides.
`java_field_is_constant` is **extracted from** `_extract_java_constants` and
moved, not transcribed. #732's lesson applied on arrival rather than after three
review rounds.

## No scope gate, and that is a fact about the grammar

Java spells a local `local_variable_declaration`, a different node type from
`field_declaration`, so #732's Kotlin problem — one node type serving both a
member and a local — cannot arise here. Asserted anyway across eight scopes
(method body, `for` body and header, `if` body, lambda body, constructor,
instance initialiser, static initialiser), because a fix routed through
`variable_declarator` — which locals *do* use — fails those.

## Two review rounds

**Round 1** found a false claim I had written: *"Java is the first live
emitter"*. The Python parser has emitted `field` for dataclass attributes since
#355, which is how @devtomnl found 399 of them rejected by both gates in #571.
The sentence was lifted from the `property` entry one CHANGELOG block down, where
it is true for Kotlin. Corrected in three places, and
`test_the_field_kind_was_already_live_before_java` now pins the prior emitter.

It also found an **ownership split I created and did not record**: a field comes
out `Account.balance` and the `static final` constant beside it comes out bare
`MAX`, so whether a Java declaration knows its owner is now decided by `static
final` — the kind of discriminator this very entry refuses. That half is #428's
behaviour and fixing it threads a parent through the Bash, Go, PHP and Java
binders at once, so it is **named and left**, not silently inherited.

**Round 2** found that round 1's own correction had reached one copy of three.
"written for all 79 specs" survived in `languages.py` and in
`test_language_spec_maps_agree.py`, both authored by this PR — the reported call
site fixed, the mechanism left, inside a review round about that. Grepping for
the *claim* rather than the spelling found five copies, two of which are a
different and **true** statement ("the test asserts the property over all 79
specs" — it does iterate the whole registry).

Round 2 also caught the new ARCHAEOLOGY row claiming test coverage the file did
not have, naming the two scopes #732's round 3 caught its Kotlin gate failing in.
Closed by **adding** the shapes rather than softening the sentence, with each
name verified present in the fixture so the absence assertions cannot pass
vacuously.

## Found by asking, filed separately

The "what other spellings exist" question was asked of every spec, and produced
five issues (policy 1 — one issue, one verdict), each re-verified against
`origin/main` in a clean worktree:

- **#741** JS/TS index a mutable `let` as `kind="constant"` — the error #732
  explicitly refused for Kotlin's `var`.
- **#742** JS/TS `var` declarations are not indexed at all (`variable_declaration`,
  a node type neither map names — #698's shape).
- **#743** PHP class properties are not indexed: `name_fields` names a `name`
  field the grammar never sets. This is the **cause** behind the declared-and-dead
  `property` kind #732 recorded as a symptom.
- **#744** PHP class constants are not indexed while file-scope ones are — the
  `_CLASS_SCOPED_CONSTANT_LANGUAGES` gate, third language.
- **#745** the spec-map guard checks that a `name_fields` entry **exists**, never
  that it **resolves**. Fixing PHP fixes PHP; fixing the guard finds the rest.

## Two records corrected that outlive this PR

⚠ **`PARSER_GENERATION` is not bumped, deliberately.** #732 took it 7 → 8 one PR
ago and both entries are under `[Unreleased]`, so every index a *release* reaches
re-parses already. The uncovered population is a tree indexed from source
*between* the two commits — a maintainer's own box (Practice 11), whose remedy is
the re-index Practice 11 already requires.

⚠⚠ **#724's `_CONFIRMED_GAPS` removal here was voluntary, and #724 claimed it
could not be.** The gate fires only when a recorded gap leaves the inventory, and
the inventory derives from `symbol_node_types` alone — so a gap closed through
`field_patterns` (or `constant_patterns`) leaves the row and the record stays
green. "The record cannot outlive the defect" holds only for fixes taking that
one route, and neither #732 nor #735 took it.

## Out of scope, recorded rather than left silent

A Java interface's `int X = 1;` is `constant_declaration`, a different node type
that yields nothing today — and so is the identical declaration in an annotation
type (`@interface Ann { int LIMIT = 3; }`). Both stay in #724's inventory.

## Blast radius

A new symbol class for every Java file changes symbol counts, and
`find_dead_code` applies no `kind` filter, so an unreferenced private field now
enters the dead-code corpus and moves `dead_code_pct` and the health-radar grade
for every Java repo. That is what the tool is for, and it is a grade movement
users will see on their next re-index. `get_dead_code_v2`, `get_hotspots`,
`get_untested_symbols`, `get_repo_health`'s complexity and test axes,
`get_parity_map` and `get_architecture_metrics` were each checked and are
unaffected.

## Evidence

- Red on tree `5aab4773802c` (12 failed), green on tree `1b1c31452ffe` — different
  trees.
- Full tier **PASS**: `10802 passed, 24 skipped` in 311.26s; `coverage.min` 82
  (floor 74), `ci.skips_windows` 24 (ceiling 25), `suite.full_seconds` 311.26
  (ceiling 360).
- DoD checklist 12/12 met or n.a., unmet: none.
- Verified by execution, not reading: no symbol lost (A/B with `field_patterns`
  emptied over a 24-shape corpus plus `tests/fixtures/java/Sample.java`), no
  declaration emitted twice across seven modifier permutations and their
  multi-declarator forms, and the `static final` rule surviving in exactly one
  place in `src/`.

https://claude.ai/code/session_01LnhGVkUAcCNkNyspGHxicW


jgravelle:
Conflicted by #747's merge and resolved locally.

Both conflicts were pure insert-at-the-same-anchor — this PR and #747 each added
a `[Unreleased]` entry at the top of `CHANGELOG.md` and a row before
`test_kind_enum_is_derived.py` in `docs/harness/ARCHAEOLOGY.md`. That is the
#693–#696 shape, and the resolution is keep-both: #747's landed entry keeps its
position relative to what follows it, this PR's goes above, and both ARCHAEOLOGY
rows are present. Verified afterwards that no content was dropped from either
side.

⚠ The merged tree was re-run rather than assumed green — branch-green is not
merged-green. Full tier **PASS**: `10809 passed, 24 skipped` in **193.95 s**
(faster than this branch's own earlier runs, because the merge brings in #747's
`sys.monitoring` coverage core).

⚠ The DoD checklist went to **11/12** on the merge and is back to 12/12. Two
causes, both worth naming:

1. The merge moved the tree, so the green stamp no longer bound it — the stamp
   binds a tree id by design.
2. More interestingly, `.claude/state/evidence/` is a **single shared
   directory**, so working a second branch in the same session had overwritten
   this PR's red evidence with the other branch's. The checklist caught it by
   name ("stamped on branch 'fix/736-738-inline-extractors'"), which is the
   check doing exactly its job.

Red was therefore re-derived properly rather than re-stamped over: the `src/`
half of this change was reverted to `origin/main`, the tests run (13 failed),
stamped, then restored and verified byte-identical to HEAD before the green run.
Red tree `eb5d3802e026`, green tree `06434d5c4717`.

PR #765, opened 2026-09-18
fix: resolve #763 - Go's blank identifier `_` is indexed as a constant

### LINKED ISSUE #763: Go's blank identifier `_` is indexed as a constant
## What is wrong

Go's blank identifier is indexed as a constant. `const ( _ = iota; KB; MB )` --
the idiomatic way to write a unit ladder -- yields a symbol named `_`:

```
parse_file('package m\n\nconst (\n\t_ = iota\n\tKB\n\tMB\n)\n', 'a.go', 'go')
-> [('KB', 'constant'), ('MB', 'constant'), ('_', 'constant')]
```

`_` is Go's DISCARD. It cannot be referenced, several can appear in one file,
and each becomes a symbol competing in every ranking under the same name. It is
noise shaped like data: a `search_symbols` result, a row in an outline, and a
name that no reference can ever resolve to.

## Scope

Found while fixing #731, which added the `var` channel; that channel skips the
blank identifier and `tests/test_go_variables.py::test_the_blank_identifier_is_not_a_symbol`
asserts it. The CONSTANT channel was deliberately left alone there rather than
changed in passing, because it is a different channel with its own history
(#428) and folding it in would have widened a change about `var` into one about
every language that binds N constants.

The same shape exists at least in Java, where `static final int _ = 1;` yields a
constant named `_`. Whether that matters is worth deciding per language: in Go
the blank identifier is a language keyword with defined semantics, in Java `_`
is a legal (if unusual) identifier and became a reserved word only in 21. So
this is not automatically a one-line shared fix, and the Go half is the one with
no defensible reading.

## Suggested shape

`_extract_go_constants` skips a spec name equal to `_`, mirroring
`_extract_go_variables`. A test over the `iota` ladder, since that is the form
real code uses. The Java question gets its own decision rather than riding
along.

fathirramadhan-web:
Closes #763

### Overview
Addresses the problem reported in #763 by resolving the underlying issue and ensuring test coverage passes.

### Verification
- Tested locally
- Linters/tests pass


jgravelle:
Thanks for this — the fix is right, and it is the fix I would have written. Skipping `_` inside the per-spec loop (rather than at the spec level) keeps `const _, B = 1, 2` yielding `B`, and the iota-ladder test is the case that matters.

Four things are holding the merge. Two are yours, one is mine, one is a line that should come out.

**1. The CLA is not signed.** I redelivered the `pull_request` webhook just now, because CLA Assistant had not fired on open at all and an absent status cannot be told apart from an unsigned one. It has reached its own verdict since, and the comment above carries the link. Signing is a form and takes about thirty seconds.

**2. `done: changelog` fails.** Anything under `src/` needs an entry in `CHANGELOG.md` under `[Unreleased]` (Definition of Done 3). It should cite #763 and credit you: what was wrong, why it was wrong, and what the index cannot do any more. `_` cannot be referenced, several can sit in one file, and each became a symbol under the same name competing in every ranking — that is the entry.

**3. Please drop the `.gitignore` hunk.** The nine lines added under `# Auto-added by github-engineer` are unrelated to the blank identifier, and one of them (`nem_stats.c`) names a file that has never existed in this repository. `.gitignore` here is a security surface — an sdist credential leak in v0.2.6 is the reason its contents are asserted by `tests/test_build.py` and `tests/test_sdist_exclusions.py` — so it does not take drive-by additions. The other two files are fine as they are.

**4. `full: test (windows-latest, 3.13)` is mine, not yours.** It failed on `test_removing_watched_root_fails[polling]` with `WinError 5`, a known Windows flake in the watcher's polling arm (#676) that has nothing to do with your change. I have re-run that job.

One optional note: your test docstring says the constant channel mirrors `_extract_go_variables`. That function is not on `main` yet — it arrives with PR #764, which is open. True shortly, not true today.

**Timebox: 2026-09-19 12:37 UTC.** If the CLA, the changelog entry and the `.gitignore` revert are done by then, this merges as your commit. If they are not, I implement the same fix myself and credit you in the CHANGELOG, in the release notes and in the comment that closes #763 — so the window decides whose commit it is, never whether you are credited and never whether the fix ships. Ask if you want longer and you can have it.

fathirramadhan-web:
Updated as requested. Thanks for the feedback @CLAassistant!

fathirramadhan-web:
Updated as requested. Thanks for the feedback @jgravelle!

fathirramadhan-web:
and

fathirramadhan-web:
and

jgravelle:
Thanks, @fathirramadhan-web. You put the skip inside the per-spec loop of `_extract_go_constants`, which is the right place, and your iota ladder test pins the case that matters. You also added the CHANGELOG entry and dropped the `.gitignore` hunk well before the deadline.

I'm closing this without merging. `license/cla` still reads "Contributor License Agreement is not signed yet" on the current head, so the default from my earlier comment applied at 2026-09-19 12:37 UTC. Without a signed CLA I can't use any of this PR's code, so neither the two lines nor the test are in the fix.

The fix is #793, written independently, and it closes #763. It drops the discard once in `parse_file`, so it covers every Go symbol kind (including `func _() {}`) and the other languages where `_` is a discard. JavaScript and TypeScript keep `_`, where it's an ordinary identifier.

The CHANGELOG entry is headed "(#763, fix by @fathirramadhan-web)". The release notes and the comment closing #763 credit you too.

The CLA signature is per account and doesn't expire, so once it's signed your future PRs merge normally. If jCodeMunch is useful to you, tell someone about it.

 -jjg

